PR #8, opened 2026-05-23
fix(implement): resolve gate script via vendored path + harden auth gate

The01Geek:
## Summary

Fixes three code-review findings on the `/devflow:implement` authorization gate (introduced earlier on this branch).

### 1. Critical: gate script unreachable in consumer installs
The `gate` job invoked `bash scripts/resolve-implement-trigger.sh` (repo root), but `install.sh` vendors the plugin into `.claude/plugins/devflow/`. In any consumer repo the script does not exist at the root, so the step fails under `set -euo pipefail`, `should_run` never becomes `true`, and **`/devflow:implement` never starts** — the headline feature of the branch was dead everywhere except the source repo (whose self-tests kept `scripts/` at root and masked it). Now uses the vendored path, matching every other workflow (e.g. `devflow-review.yml`).

### 2. Auth gate denied legitimate users on transient API errors
`resolve-implement-trigger.sh` collapsed **all** `gh api .../permission` failures to `"none"` (`2>/dev/null || echo none`), so a rate-limit / 5xx / network blip silently denied a genuine write/admin collaborator and reported it as a permission problem. Now distinguishes HTTP 404 (definitively not a collaborator) from transient failures, retries once before failing closed, and emits an accurate diagnostic. Two new tests cover both paths.

### 3. Misleading `@claude`-redundant nudge
The nudge claimed prefixing `@claude` "wakes the light `@claude` listener for nothing", but `claude.yml` carries `!contains(..., '/devflow:implement')` on every trigger clause, so it never fires on an implement command. Corrected the explanatory comment and the user-facing message.

## Testing
`bash lib/test/run.sh` → **231 passed, 0 failed** (was 227; +4 for the new transient/404 resolver cases). Workflow YAML validated.


The01Geek (commented):
# Review Report

## Verdict: APPROVE with notes (all three target fixes are correct; remaining items are advisory robustness/diagnostics/test-coverage improvements — none blocking)

## Issue Compliance
No related issue found — requirement compliance not checked. (PR resolves three findings from a prior code review of the `/devflow:implement` gate.)

## Diff classification
`engine_self_modifying` (touches `lib/test/run.sh`) → full engine ran; Phase 3 agent gates bypassed. `type-design-analyzer` was not dispatched (diff is shell-only — zero new types to analyze); single-agent omission, below the 2-agent "partial coverage" threshold.

## Verification Checklist Results
- (14 checked, 14 passed, 0 failed, 0 inconclusive)
- All cross-boundary contracts hold: vendored script path matches `install.sh` vendoring + the `devflow-review.yml` convention (VC-3); `should_run`/`number` GHA output keys match end to end (VC-7); the resolver's `grep 'HTTP 404'` matches the gh CLI stderr format and the test stub reproduces it faithfully (VC-1/VC-4); event-context coalescing, `[bot]`-suffix handling, and the trigger-text/regex contract are consistent (VC-6/VC-9/VC-10/VC-13).
- Note (VC-2): the authorized set `admin|write|maintain` is correct, but GitHub's legacy `.permission` field returns `write` for a maintain-role user, so the `maintain` case arm is effectively unreachable dead-code. Harmless — authorization still succeeds for maintainers.

## Code Review Findings

### Important / Major
- [silent-failure-hunter] `bt="$(echo "$b" | xargs)"` trims `allowed_bots` entries via `xargs`, which does shell word-splitting/quote processing rather than plain trimming (raised by 1/6 agents). Today's logins (`[a-z0-9-]`) are safe, but a malformed config value containing a quote/backslash could alter the parsed token or make `xargs` exit non-zero — and since it's a bare assignment under `set -e`, that aborts the resolver mid-loop and fails the gate job hard rather than failing closed. Hardening fix: use parameter-expansion trimming. `scripts/resolve-implement-trigger.sh:49-53`, kind=security.
- [silent-failure-hunter] If the vendored script is absent (repo not yet `install.sh`-vendored / mid-migration), the gate step exits non-zero under `set -e` and the whole workflow goes red with a raw `No such file or directory` and no issue comment — a misleading signal to the triggerer (raised by 1/6 agents). Suggested: preflight-check the path and fail *closed* (`should_run=false`) with an `::error::` naming the missing-vendoring root cause. `.github/workflows/claude-implement.yml:119-126`, kind=logic_error. *(Note: arguably acceptable — a genuinely broken install failing loudly is defensible; offered as a robustness improvement.)*

### Suggestion / Minor
- [comment-analyzer + code-reviewer] The `deny_reason` message and the inline comment say "write/admin collaborator" but the code authorizes `admin|write|maintain` — a denied maintainer gets a message that misstates which roles qualify (**corroborated, 2/6 agents**). Test #7 bakes the wording into an assertion, locking the drift in. `scripts/resolve-implement-trigger.sh:55,59`, kind=comment_drift.
- [silent-failure-hunter] The resolver buckets *all* non-404 `gh` failures into one `__lookup_failed__`/"transient" diagnostic, including non-transient causes (gh missing, `GH_TOKEN` unset, 401/403 scope) — operators see "transient... failing closed" and retry a permanent misconfiguration forever (1/6). Suggested: surface the first line of captured stderr in the warning. `scripts/resolve-implement-trigger.sh:64-84`, kind=logic_error.
- [silent-failure-hunter] `grep -q 'HTTP 404'` is load-bearing string-matching against gh's human-readable stderr; a future gh rephrase would misclassify a real non-collaborator as transient (still fails closed, but adds a needless retry-sleep on every unauthorized comment) (1/6). `scripts/resolve-implement-trigger.sh:69-72`, kind=api_misuse.
- [silent-failure-hunter] The nudge step's `|| echo "::warning::Could not post..."` discards gh's actual error; capture `2>&1` so the cause is visible (1/6). `.github/workflows/claude-implement.yml:158-168`, kind=logic_error.
- [silent-failure-hunter] The nudge's `NUMBER` guard silently no-ops on the "shouldn't happen" empty/`null` path with no breadcrumb (1/6). `.github/workflows/claude-implement.yml:160-166`, kind=logic_error.
- [comment-analyzer] `docs/cloud-setup.md:82-88` "Who can trigger it" callout still frames repo-permission gating as label-path-only; after this PR the comment/review/body path is *also* collaborator-gated by the resolver (1/6). kind=comment_drift.
- [pr-test-analyzer] Test gaps (ratings 4-6, 1/6): the regex's `#?` and case-insensitive (`-i`) arms are untested (only `/devflow:implement 7` is exercised); the `allowed_bots` whitespace-trim and non-first-element match ride on a single happy-path assertion; the retry loop is tested for *outcome* but no stateful stub proves it *recovers* a collaborator on attempt 2 (a regression removing the loop would still pass). `lib/test/run.sh`, kind=test_gap.

### Positive (verified, not assumed)
- All 231 shell + 58 python assertions pass; the new transient-vs-404 split is independently confirmed by exercising every branch.
- Injection-safe: `TRIGGER_TEXT`/`ACTOR`/`BODY` pass via `env:`, never interpolated into `run:` bash.
- The double-fire claim is correct: `claude.yml` carries `!contains(..., '/devflow:implement')` on every clause and does not subscribe to `labeled`, so neither the bare nor the `@claude`-prefixed form double-fires (the fix #3 rationale is now accurate).
- Fail-closed posture holds for empty `ACTOR`/`REPO`/`ALLOWED_BOTS` and unresolvable numbers.

## Verdict Criteria
- No checklist FAIL or INCONCLUSIVE → not REJECT on Phase 1+2.
- No Critical findings from review agents → not REJECT on Phase 3.
- Only Important/Suggestion findings remain → **APPROVE with notes.**
- The three target fixes (vendored path, transient-vs-404 auth hardening, nudge rationale) are all confirmed correct with matching test coverage.

PR #172, opened 2026-06-29
fix: teach /devflow:implement Phase 1.4 to use the harness worktree branch instead of creating a second branch

### LINKED ISSUE #168: fix: teach /devflow:implement Phase 1.4 to use the harness worktree branch instead of creating a second branch
## Problem Statement
When `/devflow:implement` is triggered from inside the local harness's pre-created git worktree (e.g. the `worktree-issue-165` worktree, on a branch like `worktree-issue-165`), Phase 1.4 fails to recognize that it is already on the branch the harness set up for it. The detection only accepts branch names matching `claude/issue-*` or `issue-*`, so the worktree branch falls through to the create path and Phase 1.4 creates a **second** branch inside the worktree. The implementation then lands on a branch the harness did not provision, leaving duplicate/confusing branch state and a PR built from the wrong branch. This hits any DevFlow operator who runs implement inside a harness-provided worktree.

## Current Behavior
`skills/implement/SKILL.md` Phase 1.4 ("Create or Detect Feature Branch") instructs: *"If the current branch matches `claude/issue-*` or `issue-*`, use it — skip branch creation. Otherwise, create a new branch."* A harness worktree branch matches neither pattern, so a redundant branch is created on top of it.

## Desired Behavior
Phase 1.4 detects that it is running inside a **linked git worktree** and uses that worktree's current branch as the feature branch, creating no second branch — regardless of the worktree branch's name. The existing `claude/issue-*` / `issue-*` name detection (the cloud-tier GitHub Action path) continues to work unchanged. When the current branch is the configured base branch, Phase 1.4 still creates a new feature branch (it never builds directly on trunk).

## User Impact
DevFlow operators running implement inside a harness worktree get the implementation, commits, and PR on the single branch the harness provisioned — no orphaned second branch, no confusion about which branch carries the work. The detection is naming-independent, so it survives any future change to the harness's worktree naming.

## Technical Context

- **Relevant Classes/Files**
  - `skills/implement/SKILL.md` — Phase 1.4 ("Create or Detect Feature Branch"). This is the **single site** carrying the detection predicate; `git grep` confirms no mirror in any `.sh`/`.py`/`.jq`/`.yml`/`.md` file.
  - `lib/test/run.sh` — already has a Phase 1.4 *base_branch* drift guard but no pin on the branch-detection predicate; this is where the new test guard lands.
  - `scripts/branch-for-issue.py` — computes the canonical `issue-{N}-{slug}` name; only invoked on the create path that this change skips for worktrees.
- **Architecture Alignment**
  - Detection uses `git rev-parse --git-common-dir` vs `git rev-parse --git-dir`: in a linked worktree they differ (the common dir points at the main repo's `.git`); in the main working tree they are equal. This is the deterministic, naming-independent signal.
  - The base branch is read the same way Phase 1.4 already reads it (`config-get.sh .base_branch main` with the existing empty-read fallback) — no new config key.
- **Dependencies** — `git` only. No new dependency.
- **Cross-layer Impact** — confined to the `/devflow:implement` Phase 1.4 setup step. Provenance/retrospective detection is label-first (the hardcoded `DevFlow` label, branch-name independent) and the workpad lives on the issue, so a PR built from a `worktree-issue-*` branch is detected and processed normally downstream — using the worktree branch as-is is safe.

## Acceptance Criteria
- [ ] When `/devflow:implement` starts on a branch the harness pre-created in a linked worktree (e.g. current branch `worktree-issue-165`), Phase 1.4 uses that branch and creates **no** second branch.
- [ ] Detection is by linked-worktree state (`git rev-parse --git-common-dir` ≠ `--git-dir`), independent of the worktree branch's name.
- [ ] Existing `claude/issue-*` and `issue-*` name detection still skips branch creation (cloud-tier GitHub Action path unbroken).
- [ ] When the current branch is the configured base branch (`main`/`base_branch`), Phase 1.4 creates a new feature branch and does **not** build on the base — even inside a worktree.
- [ ] When not in a worktree and not on a recognized feature branch, Phase 1.4 creates a new branch via `branch-for-issue.py` (unchanged).
- [ ] `lib/test/run.sh` gains a guard pinning the new worktree-detection behavior, asserting it fails closed; the full suite (`lib + python tests`) and lint stay green.

## Implementation Notes
- **Approach** — In `skills/implement/SKILL.md` Phase 1.4, before the existing name-pattern check, add a deterministic linked-worktree branch: resolve `BASE` (existing `config-get.sh .base_branch main` read), capture `CUR=$(git branch --show-current)`, and if `git rev-parse --git-common-dir` ≠ `git rev-parse --git-dir` **and** `CUR` is non-empty and not equal to `BASE`, use `CUR` and skip creation. Keep the `claude/issue-*` / `issue-*` name match as a second skip condition for the cloud tier. Fall through to the existing `branch-for-issue.py` create path otherwise. Then fill the workpad `Branch` line exactly as today.
- **Code Patterns** — Mirror the existing Phase 1.4 `BASE` read and its empty-read fallback guard; reuse `git branch --show-current` already used elsewhere in the skill for context recovery.
- **Testing Strategy**
  - **Boundary & level:** The change is shell logic embedded in a skill body, exercised via `lib/test/run.sh` (the suite's pattern for SKILL.md invariants). Test level: a shell-level guard in `run.sh` that pins the documented predicate.
  - **Happy path:** a guard asserting the Phase 1.4 text/logic treats "linked worktree + current branch ≠ base" as "use current branch, skip creation."
  - **Skipped-step / guarantee path:** because this is the deterministic detection mechanism, the guard must assert it fires on the worktree path where the old name-pattern check would have *failed* (branch named `worktree-issue-<N>`, which matches neither `claude/issue-*` nor `issue-*`) — proving the new behavior covers exactly the case that previously created a second branch.
  - **Base-branch guard:** assert that "inside a worktree but on the base branch" does **not** skip creation.
  - **Regression framing:** the new guard must fail against the current Phase 1.4 text (which has no worktree handling) and pass after the edit.
  - Maps every AC above to at least one assertion in `run.sh`; no assertion stands without an AC.
- **Documentation Needed** — Update the Phase 1.4 one-liner in `docs/DEVFLOW_SYSTEM_OVERVIEW.md` ("Create or detect the branch …") to mention worktree detection. No external/customer docs affected.
- **Potential Gotchas**
  - This is a `SKILL.md` edit — it **must** be done via the `superpowers:writing-skills` RED/GREEN discipline, not hand-edited.
  - Coupled-invariant discipline: land the `run.sh` guard in the **same** change as the SKILL.md edit, not a follow-up commit.
  - Keep the cloud-tier name-pattern detection intact; the GitHub Action checks out `claude/issue-*` and is not a worktree.

The01Geek:
## Summary
- Teaches \`/devflow:implement\` Phase 1.4 to **reuse the branch a local harness pre-created in a linked git worktree** instead of creating a redundant second branch — so the implementation, commits, and PR land on the single branch the harness provisioned.
- Detection is deterministic and **naming-independent** (a linked worktree's \`git rev-parse --path-format=absolute --git-common-dir\` differs from its \`--git-dir\`, compared byte-for-byte in absolute form), so it survives any future change to the harness's worktree-branch naming; the existing \`claude/issue-*\`/\`issue-*\` cloud-tier name match is preserved as a second skip condition.
- Hardened across four review iterations + two shadow passes: the decision is fully fail-closed (never builds on trunk, never reuses a detached HEAD, never false-positives a worktree on a \`git rev-parse\` failure, empty result, or asymmetric result), and every destructive create-path step is guarded with an attributable breadcrumb.

## Changes
**\`skills/implement/\` (Phase 1.4)**: Replaced the name-pattern-only branch detection with a single consolidated decision block — resolve \`BASE\`, capture \`CUR\`/the git-dir layout defensively, then reuse the current branch via either the linked-worktree signal (\`--git-common-dir\` ≠ \`--git-dir\`, both captures in absolute form via \`--path-format=absolute\` so the comparison is byte-consistent regardless of cwd spelling, compared only when both are non-empty) or the cloud-tier name match — with the non-empty + not-base guards wrapping **both** signals. The create path guards the fetch (with breadcrumb), branch-name derivation from \`branch-for-issue.py\` (exit-status + empty-name guards with breadcrumbs), and \`git checkout -b\` itself; the reuse path skips fetching.

**\`lib/test/\`**: Coupled guard for the new mechanism — token pins on the SKILL surface (CUR capture, both rev-parse flags, base guard, fetch breadcrumb, \`branch-for-issue.py\` exit and empty-name guards), a \`decide_branch\` behavioral matrix (reuse / create / fail-closed rows incl. base-named-\`issue-*\`, detached-HEAD, symmetric rev-parse failure, and asymmetric rev-parse results), and the \`base_branch\` resolver/guard behavioral coverage block. A real-git integration test exercising \`--path-format=absolute\` in an actual worktree is tracked as a deferred test gap (F-9).

**Docs / versioning**: \`docs/DEVFLOW_SYSTEM_OVERVIEW.md\` Phase 1.4 one-liner updated to describe worktree detection with the config-driven base; version bumped to \`2.8.20\` with a matching CHANGELOG entry.

## Resolves
Resolves #168

## Test Plan
- [ ] \`lib + python tests\` CI job is green on this PR (observed locally: \`lib/test/run.sh\` → 1665 passed, 0 failed; \`shellcheck\` + \`ruff\` clean).
- [ ] Run \`/devflow:implement <N>\` from inside a harness-provided linked worktree (branch \`worktree-issue-<N>\`) and confirm it reuses that branch and creates no second branch (this PR was itself produced that way on branch \`worktree-issue-168\`).
- [ ] Confirm the cloud-tier path is unchanged: a \`claude/issue-*\` branch in the main working tree is still reused.

## Visual Changes
N/A

## Breaking Changes
None


The01Geek:
## Review findings addressed

All findings from the REJECT run have been resolved. Summary:

**Blocker (VC-1 FAIL — docs/DEVFLOW_SYSTEM_OVERVIEW.md Phase 1.4 not updated):** Fixed in commit `310baca`. Line 224 now reads: *"Branch creation is skipped when the run is already in a linked git worktree (naming-independent harness signal) or when the current branch matches a recognized feature-branch pattern — the existing branch is reused as-is."* The hard-coded `origin/main` reference was also replaced with `config-driven base_branch`.

**Important #1 (Signal 1 decision not gated on both rev-parse operands non-empty):** Fixed in `1bd82ca`. The worktree-reuse `if` at `skills/implement/SKILL.md:298` was hardened to `if [ -n "$COMMON_DIR" ] && [ -n "$GIT_DIR_PATH" ] && [ "$COMMON_DIR" != "$GIT_DIR_PATH" ]`, gating the decision on both operands being present before the inequality is evaluated. An asymmetric result (one value empty) now falls through to the create path.

**Important #2 (CUR capture discards exit status with no breadcrumb):** Fixed in `310baca`. `CUR=$(git branch --show-current)` → `CUR=$(git branch --show-current 2>/dev/null) || CUR=""`, mirroring the rev-parse treatment.

**Important #3 (rev-parse-failure breadcrumb could self-contradict Signal 2 reuse):** Fixed in `310baca`. The breadcrumb was rephrased to "linked-worktree detection (Signal 1) disabled" — it no longer claims the run will create a branch, so a Signal 2 name-match that follows cannot contradict it.

**Rebase + re-version:** The branch was rebased onto current main (2.8.19) and re-versioned to **2.8.20**. The `lib/test/run.sh` pin for the CUR-capture token was updated to match the hardened `2>/dev/null) || CUR=""` form (the old bare-form pin caused a suite FAIL post-rebase; fixed in `770f1c5`).

**Suite:** 1663 passed, 0 failed (run locally after rebase + pin fix).

The01Geek:
## Review findings addressed (third run — APPROVE with notes)

### Important #1 — PR description self-record drift
Fixed by updating the PR body. The description previously claimed \`cd … && pwd -P\` normalization, a real-git worktree integration test, and version 2.8.18 — none of which matched what was shipped. The body now accurately describes \`--path-format=absolute\`, a literal-string \`decide_branch\` matrix, and version 2.8.20.

### Important #2 — Create-path BRANCH unguarded
Fixed in \`skills/implement/SKILL.md\`. \`branch-for-issue.py\` now has an exit-status check and its result has an empty-name guard, each with an attributable DevFlow breadcrumb — consistent with the fetch/rev-parse hardening on the same block. The PR body's prior claim that this was already guarded was incorrect; the self-record is now accurate.

### Suggestion #2 — "--path-format=absolute" overstated as "directory identity"
Fixed in both the SKILL comment and the CHANGELOG entry. Phrasing changed to "byte-consistent regardless of how the caller's cwd was spelled" with an explicit note that symlinks and \`...\` are not canonicalized.

### Suggestion #4 — Breadcrumb wording implied git-fetch fatal-exit semantics
Fixed. Breadcrumb rephrased to name the specific failure (\`git rev-parse --path-format=absolute failed\`) and list plausible causes (repo corrupt, git < 2.31, broken env) — no longer implies the same fatal-exit semantics as the adjacent fetch guard.

### Suggestion #1 — F-8 deferred pointer missing
Added "Source of truth: SKILL Phase 1.4" comment above the \`decide_branch\` mirror, plus an F-9 deferred note that \`--path-format=absolute\` normalization is not behaviorally exercised (mirror receives pre-resolved strings; a real-git integration test would be needed to cover the rev-parse call itself).

### Important #3 / Suggestion #5 — Normalization not behaviorally tested
Noted as F-9 deferred in the test comment block. The asymmetric rev-parse rows already cover the fail-closed decision path; a full real-git test exercising \`--path-format=absolute\` in an actual worktree is the remaining gap.

### Not addressed (pre-existing, out of scope)
- Suggestion #3: \`workpad.py --branch\` swallows its exit status — pre-existing, not introduced or relocated by this diff; tracked as a separate concern.

**Suite:** 1665 passed, 0 failed (run locally post-commit).

The01Geek:
## Review findings addressed (fourth run — APPROVE with notes)

### Important #1 — Breadcrumb collapses asymmetric and full-failure cases into same message

The breadcrumb fires when `[ -n "$COMMON_DIR" ] && [ -n "$GIT_DIR_PATH" ]` is false — i.e. when *either or both* values are empty. The previous wording "git rev-parse --path-format=absolute failed" implied both calls exited non-zero, which mischaracterises the asymmetric case (one value empty via `|| VAR=""` — e.g. an injected `GIT_DIR` or `GIT_COMMON_DIR` env override that made one call return nothing).

Fixed: rephrased to **"one or both git-dir path values are empty"** with explicit causes covering all three trigger cases: git < 2.31 lacking `--path-format`, repo corrupt, and injected env override. Coupled pins updated: old `git rev-parse --path-format=absolute failed` pin replaced with pins on the emptiness wording and the env-override cause.

### Important #2 — `USE_CURRENT` skip enforced only by SKILL prose

The create block (fetch → branch-for-issue.py → git checkout -b) was guarded only by the prose "If `USE_CURRENT` is set, skip branch creation entirely." Fixed: wrapped the entire create block in `if [ -z "$USE_CURRENT" ]; then ... fi`, making the reuse/create decision a structural shell guard. New `assert_pin_unique` pin on `[ -z "$USE_CURRENT" ]`.

**Verification (per updated step 7 discipline):**
- Important #1 fix verified for all three trigger conditions: both-empty (symmetric failure), COMMON_DIR-empty-only (asymmetric), GIT_DIR_PATH-empty-only (asymmetric) — wording "one or both…empty" covers all three; named causes cover each scenario.
- Important #2 fix verified for all USE_CURRENT states: `USE_CURRENT=1` (either signal) → `[ -z "1" ]` false → create block skipped ✓; `USE_CURRENT=` (neither signal) → `[ -z "" ]` true → create block runs ✓.

**Suite:** 1672 passed, 0 failed (post-rebase, post-commit).

**Re-version:** main took 2.8.20 for #173 while this branch was in review. Rebased onto current main; re-versioned to **2.8.21**. CHANGELOG consolidated to a single entry (prior iterative re-version entries removed).

PR #177, opened 2026-06-29
Mechanize the review-and-fix iter-N.json persist: validate fields + derive loop_role as a best-effort backstop in efficiency-trace

### LINKED ISSUE #170: Mechanize the review-and-fix iter-N.json persist: validate fields + derive loop_role as a best-effort backstop in efficiency-trace
## Problem Statement

PR #166 (issue #165) adds a `loop_role` field to the `/devflow:review-and-fix` per-iteration workpad (`iter-<N>.json`). The schema defines it and Step 3 item 7 of `skills/review-and-fix/SKILL.md` tells the orchestrator to hand-write it on every iteration — but **nothing enforces or realizes the persist.** The whole `iter-<N>.json` is composed inline by the orchestrator (an LLM), and `loop_role` is the *weakest* field on it: the SKILL.md schema note states verbatim that it is **"legibility-only: no consumer reads it."** Confirmed in code — `loop_role` appears nowhere in `lib/efficiency-trace.jq`, `lib/efficiency-trace.sh`, or any other consumer.

So the field rides the most fragile, most-droppable step in the loop (the inline per-iteration persist, which the skill itself treats as droppable — that is exactly why Step 3 item 7 exists as the incremental backstop for a dropped Loop Exit) **and** has zero functional pressure to be written, because no consumer errors when it is absent. The legibility win it promises — "see fix-vs-promoted iteration roles at a glance from the persisted workpads" — is real but, in practice, will be silently omitted and lost with nothing detecting the loss.

The maintainer auditing a multi-iteration run hits this pain: the persisted record they rely on for legibility is exactly the record that can quietly come back incomplete.

## Current Behavior

- `iter-<N>.json` is hand-composed inline by the review-and-fix orchestrator (Step 3 item 7), one model-authored JSON object per iteration with ~12 fields (`fix_commit_sha`, `fix_files`, `loop_role`, `checklist`, `phase3_dispatched`, `diff_profile`, findings, `fix_decisions`, `convergence_inputs`, `cap_drops`, `telemetry`, …).
- No helper validates that an emitted `iter-<N>.json` actually carries the fields the schema requires. A dropped field is invisible until a downstream derivation silently produces a poorer record.
- `loop_role` specifically has no consumer and no derivation: if the orchestrator forgets it, it is simply gone.
- The existing backstop layer already derives a per-run record from these iter files and degrades gracefully: `lib/efficiency-trace.sh --self-check` (Layer 2: warn-only, never writes, never fails), `--persist` (Layer 3: derive + commit the per-run record), and `lib/efficiency-trace.jq` (the shared derivation). `loop_role` is not part of any of it.

## Desired Behavior

The `loop_role` legibility win is made real by a **best-effort validating backstop** in the existing `efficiency-trace` machinery, and the persist of the whole `iter-<N>.json` gains a safety net:

- `lib/efficiency-trace.jq` **derives** `loop_role` for every iteration of the per-run record from the shadow-block rule, so the value is present in the derived record whether or not the orchestrator wrote it into the iter file.
- The derived per-run record **surfaces** `loop_role` per iteration, giving the field a real consumer — it is no longer "no consumer reads it."
- `lib/efficiency-trace.sh --self-check` additionally **warns** (best-effort, never writes, never fails) when any `iter-<N>.json` in the run is missing a field the schema expects, naming the field and the iteration — turning a silently-dropped persist field into a visible warning.
- All of this degrades gracefully (exit 0, sensible defaults) on missing/malformed input, matching the existing `efficiency-trace` contract that every mode never aborts.

## User Impact

The maintainer auditing a `/devflow:review-and-fix` run gets a per-run record whose `loop_role` is reliably present (derived, not dependent on the orchestrator remembering), and a warning when any other iter-field was dropped during the inline persist. The downstream efficiency-trace / retrospective consumers gain `loop_role` as real signal. The fix reuses the existing backstop layer rather than adding a new enforcement surface.

## Technical Context

- **Relevant Classes/Files**
  - `lib/efficiency-trace.sh` — owns `--mode` / `--self-check` / `--persist` over the run's iter files; the validation + warn pass is added here (it already holds the best-effort/never-abort backstop style).
  - `lib/efficiency-trace.jq` — the shared derivation that reads `iter-<N>.json`; the `loop_role` derivation + surfacing in the per-run record goes here.
  - `skills/review-and-fix/SKILL.md` — the `iter-<N>.json` schema block (`"loop_role": "fix | promoted"`), the Step 3 item 7 persist rule, and the schema note that currently calls `loop_role` "legibility-only: no consumer reads it" (which this change makes false and must correct). The Decide-outcome-2 promotion semantics (Step 2.6 "Decide" outcome 2 / Step 4.5) define the derivation rule.
  - `lib/test/run.sh` — carries the `loop_role` source-of-truth pins (Issue #165 Part A assertions) and is where the new coverage lands.
- **Architecture Alignment** — Extends the existing three-layer efficiency-trace backstop (derive / self-check / persist) rather than introducing a parallel mechanism. The derivation mirrors the schema note's own statement that `loop_role` would otherwise be "reconstructed by inference from the prior iter's shadow block." Best-effort/never-abort posture matches `--self-check` and `--persist`.
- **Dependencies** — **Depends on PR #166 (issue #165) merging to `main` first.** `loop_role`, its schema block, and the Step 3 item 7 persist rule do not exist on `main` yet (they live on the open `issue-165-…` branch). This issue cannot be implemented until #166 lands; the SKILL.md "no consumer reads it" note this change corrects only exists after #166. No code change here predates that merge.
- **Data/Schema Considerations** — The `loop_role` derivation rule: iteration 1 is `fix`; iteration N is `promoted` when iteration N−1's `shadow` block records a Decide-outcome-2 promotion, otherwise `fix`. A persisted `loop_role` value, when present, is preserved; an absent/empty one is derived. The "expected iter-field set" the `--self-check` validation checks is defined once and kept in sync with the SKILL.md schema (a `run.sh` pin asserts they agree).
- **Cross-layer Impact** — Backend/tooling only: shell + jq helpers, one SKILL.md note correction, and tests. No frontend, API, or DB surface.

## Acceptance Criteria

- [ ] `lib/efficiency-trace.jq` derives `loop_role` for every iteration of the per-run record: iteration 1 → `fix`; iteration N → `promoted` when iteration N−1's `shadow` block records a Decide-outcome-2 promotion, else `fix`.
- [ ] When an `iter-<N>.json` already carries a non-empty `loop_role`, the derivation preserves that persisted value; when it is absent or empty, the derived value is used (idempotent on consistent inputs).
- [ ] The per-run record emitted by `lib/efficiency-trace.jq` includes `loop_role` per iteration, so the field has a real consumer.
- [ ] `skills/review-and-fix/SKILL.md` is corrected in the same change so the note no longer claims `loop_role` is "no consumer reads it"; it states that `efficiency-trace.jq` derives and surfaces it. The schema/persist-rule wording stays internally consistent.
- [ ] `lib/efficiency-trace.sh --self-check` emits a `::warning::` to stderr — and exits 0, writing no files — for each `iter-<N>.json` in the run that is missing any field in the expected iter-field set, naming the missing field and the iteration number.
- [ ] The expected iter-field set used by `--self-check` is defined in exactly one place and kept in sync with the SKILL.md schema; a `lib/test/run.sh` assertion fails if the two diverge.
- [ ] On missing run directory, an unparseable `iter-<N>.json`, or an absent `shadow` block, every new code path exits 0 and yields the `fix` default (and, for `--self-check`, a warning) — never aborts, never mutates an iter file, consistent with the existing `efficiency-trace` "every mode never aborts" contract.
- [ ] `bash lib/test/run.sh`, `shellcheck --severity=warning`, and `ruff check` all pass (the `lib + python tests` CI job is green).

## Implementation Notes

- **Approach** — Extend `lib/efficiency-trace.jq` to compute `loop_role` per iteration from the shadow-block rule (preserving any persisted value), and to surface it in the per-run record. Extend `lib/efficiency-trace.sh --self-check` with a warn-only validation pass over each `iter-<N>.json` against a single-source expected-field set. Correct the `loop_role` note in `skills/review-and-fix/SKILL.md` and reconcile its `run.sh` pins in the same commit. No change to the orchestrator's inline-composition step is required — the backstop sits behind it. Because `skills/review-and-fix/SKILL.md` is a skill file, the edit is made through the `superpowers:writing-skills` RED/GREEN discipline, not by hand.
- **Code Patterns** — Mirror the existing `--self-check` warn-only path in `lib/efficiency-trace.sh` (`::warning::` to stderr, exit 0, no writes) and the existing `efficiency-trace.jq` derivation helpers. Keep the jq `#` comments apostrophe-free ASCII and type-guard any field access (`(.x | strings)` / type-check before indexing) per the repo's jq gotchas. Portability: no GNU-only flags.
- **Coupled-invariant discipline** — The SKILL.md schema literal `"loop_role": "fix | promoted"`, the Step 3 item 7 persist rule, the "no consumer reads it" note, the expected-field-set in `efficiency-trace.sh`, and the `run.sh` pins are coupled sites. `git grep` `loop_role` across `.sh`/`.jq`/`.md`/`.py` first, change every site together, and reconcile the asserting tests in the **same** commit — not a follow-up.
- **Testing Strategy** —
  - **Move 1 — Boundary & level.** Automated tests apply. Two boundaries: (1) the jq derivation's *output value* (`loop_role` per iteration in the per-run record), and (2) the `--self-check` shell *contract* (stderr `::warning::` text + exit code + no-file-write). Both are exercised by `lib/test/run.sh` (gh-stubbed, no network) at the helper level — a jq-output assertion and a shell-invocation assertion.
  - **Move 2 — Coverage dimensions.**
    - *Happy path:* a fixture run where iter-1's `shadow` block records a Decide-outcome-2 promotion → derivation yields iter-2 `loop_role: "promoted"`, iter-1 `"fix"`, both surfaced in the per-run record.
    - *Boundary/degenerate:* iteration 1 always derives `fix`; an iter file with `loop_role` absent **and** no `shadow` block derives `fix`; an iter file with a persisted `loop_role` has that value preserved.
    - *Error/malformed:* a missing run directory and an unparseable `iter-<N>.json` each yield exit 0 with the `fix` default and no file mutation.
    - *Skipped-step path (the reason the backstop exists):* a fixture iter file that **omits** `loop_role` entirely (simulating the orchestrator dropping it) still yields `loop_role` in the derived record — the guarantee must hold on the path where the cooperative persist was skipped.
    - *Validation warning:* a fixture iter file missing a non-derivable expected field (e.g. `telemetry`) → `--self-check` emits a `::warning::` naming that field and the iteration and exits 0; the iter file is byte-identical afterward.
    - State/concurrency, scale, security: do not apply — read-only derivation over run-scoped scratch, no shared state, no access boundary.
  - **Move 3 — Named assertions.** Each AC maps to ≥1 `run.sh` assertion: derivation-rule ACs → the promoted/fix fixture assertions; preserve-persisted AC → the preserved-value assertion; consumer AC → the per-run-record-contains-`loop_role` assertion; SKILL.md-correction AC → a pin asserting the note no longer contains "no consumer reads it" and the schema literal is intact; self-check-warn AC → the missing-field warning + exit-0 + byte-identical-file assertion; single-source-field-set AC → the schema-vs-validation pin; graceful-degradation AC → the missing-dir / unparseable-file exit-0 assertions. No orphan assertions. Fixtures: hand-authored `iter-<N>.json` files under a temp run dir; nothing is mocked beyond the existing `gh` stub. The unit under test (the jq filter and the shell pass) is never mocked.
- **Documentation Needed** — The `loop_role` note correction in `skills/review-and-fix/SKILL.md` is the doc change; if `docs/DEVFLOW_SYSTEM_OVERVIEW.md` or `docs/shadow-review.md` describe the workpad/backstop layers, align them with the new derivation/consumer in the same pass.
- **Potential Gotchas** — `loop_role` is currently introduced on the unmerged #166 branch; do not start until it is on `main` (see Dependencies). The SKILL.md edit must go through `superpowers:writing-skills`. The "no consumer reads it" note becomes false the moment the jq surfaces the field — leaving it stale is the dominant `convention-violation` desync this repo guards against. Keep the change versioning-correct (a `plugin.json` version bump needs a matching `CHANGELOG.md` entry; default `patch`).

The01Geek:
## Summary
- Makes the `/devflow:review-and-fix` per-iteration `loop_role` field real: `lib/efficiency-trace.jq` now **derives and surfaces** it per iteration of the per-run record, giving the field a real consumer (it shipped in 2.8.17 as hand-written, legibility-only, no-consumer state).
- Adds a best-effort **field-validation pass** to `lib/efficiency-trace.sh --self-check` so a silently-dropped `iter-<N>.json` persist field becomes a visible `::warning::`.
- Corrects the now-false `skills/review-and-fix/SKILL.md` note that called `loop_role` "legibility-only / no consumer reads it".

## Changes
**`lib/efficiency-trace.jq`**: Derives `loop_role` per iteration (iteration 1 → `fix`; iteration N → `promoted` when iteration N−1's `shadow.promoted_to_iter_next` is set, else `fix`), preserving any persisted non-empty value, and surfaces it in the record's `per_iteration`. `shadow_promoted` is coerced to a strict boolean so a malformed producer value can never over-classify.
**`lib/efficiency-trace.sh`**: Adds the single-source `ITER_EXPECTED_FIELDS` constant and a warn-only validation pass in `do_self_check` — for each `iter-<N>.json` missing an expected field it emits a `::warning::` naming the field and the iter file, exit 0, no writes. The jq call is `if !`-guarded so a malformed/unparseable iter never aborts the pass under `set -e`.
**`skills/review-and-fix/SKILL.md`**: Corrects the schema note + Step 3 item 7 to name `efficiency-trace.jq` as the deriving/surfacing consumer (drops "legibility-only").
**`lib/test/run.sh`**: Adds the `loop_role #170` coverage — derivation, persisted-value preservation, graceful degradation (unparseable / missing-dir / unparseable-`--self-check` → exit 0), the missing-field warning + byte-identical no-mutation, 3-iter non-propagation, empty/non-string fallback, strict-boolean coercion, and a single-source `ITER_EXPECTED_FIELDS` ↔ SKILL.md schema divergence guard; all mutation-checked.
**`docs/efficiency-trace.md`**: Documents the `loop_role` derivation in the per-run record and the new `--self-check` field validation.
**Versioning**: Bumps `2.8.18` → `2.8.19` with the matching `CHANGELOG.md` entry (re-versioned from 2.8.18 after `main` advanced to 2.8.18 mid-run).

## Resolves
Resolves #170

## Test Plan
- [ ] `lib/test/run.sh` passes (the `lib + python tests` CI job is green) — locally observed 1642 passed, 0 failed.
- [ ] `shellcheck --severity=warning` and `ruff check` pass — locally observed clean.
- [ ] Confirm the `loop_role` derivation rule against a real multi-iteration `iter-<N>.json` run (iter 1 `fix`; a post-promotion iter `promoted`).

## Visual Changes
N/A

## Breaking Changes
None — additive backstop on the existing `efficiency-trace` machinery; the orchestrator's inline persist is unchanged.


The01Geek:
## Review response

**Both Important findings were already resolved** in a prior commit (`b446d0a`) added during earlier review iterations — the `if !` arm emits a `::warning::` for corrupt/unreadable files, and the `__non_object__` sentinel warns for parsed-but-non-object shapes. The review ran against the pre-fix state.

**Suggestion 1 (stale `path:line`):** Fixed — replaced the "line 647" reference with a textual description, per the project convention.

**Suggestions 3, 6, 2 (test gaps):** Added three new tests:
- **(11)** `--self-check` is silent on a fully-complete iter (all 13 `ITER_EXPECTED_FIELDS`, no `shadow` key). Guards an inverted set-difference from passing all missing-field assertions while failing the clean case. Record file pre-created so only field-validation warnings can surface.
- **(12)** Mid-chain double-promotion: `false/true/true` shadow chain → `fix/promoted/promoted`. Locks the positional-prior indexing against an off-by-one (existing test (7) covers `fix/promoted/fix`; this covers consecutive promotions).
- **(13)** Persisted `"fix"` suppresses a derived promotion — mirror of test (8)'s persisted-`"promoted"` case.

**Suggestions 4 and 5:** Folded into test (11) (shadow-exempt iter at runtime) and left as advisory respectively — `.iter` ordering is already documented as best-effort in the comment.

**Suggestion 7 (advisory):** No action — doc note deferred per the advisory designation.

Suite: **1702 passed, 0 failed** (up from 1699).

Re-versioned to `2.8.22` (main advanced to `2.8.21` during rebase).

PR #234, opened 2026-07-01
fix: don't clobber a hand-set devflow_version on install.sh re-run

The01Geek:
## Summary
- `install.sh`'s `set_config_version` unconditionally re-stamped `devflow_version` in `.devflow/config.json` to the newly-installed commit SHA on every run, even when a user had deliberately set it to something else (e.g. `"main"` to track the branch, or a tag) — silently converting a "track main" pin into a frozen SHA on the next update run.
- This contradicted `docs/cloud-setup.md`, which documents re-runs as preserving existing values and only backfilling newly-added keys.

## Changes

**install.sh**: `set_config_version` now only re-stamps `devflow_version` when the existing value is empty or already looks like a commit SHA (7-40 lowercase hex chars — i.e. it was itself written by a previous installer run). A non-SHA value (`"main"`, a tag) is treated as a deliberate pin and left untouched. Both the `jq` and `python3` backends were updated identically.

**docs/cloud-setup.md**: documented this as the one narrow, intentional exception to "existing values are preserved" on re-run.

**lib/test/run.sh**: added coverage for both branches — a hand-set non-SHA `devflow_version` survives re-run, and a previously auto-stamped SHA-like value still gets updated.

## Test Plan
- [x] `lib/test/run.sh` — full suite passes (2237 passed, 0 failed), including new `scv:` assertions for both preserve and re-stamp cases
- [x] `shellcheck --severity=warning -e SC1091 install.sh` — clean
- [x] Manually exercised `set_config_version` against `main`, a SHA-like value, and an absent key on both the `jq` and `python3` backends

## Visual Changes
N/A

## Breaking Changes
None. This only changes behavior for users who had hand-set `devflow_version` to a non-SHA value (e.g. `"main"`); previously that value was silently clobbered on the next `install.sh` re-run, now it is preserved as intended.


The01Geek:
Addressed the two Important findings from the Devflow Review (commit bb2e50b):

1. **jq/python3 message parity on idempotent re-run** — the jq backend classified pinned-vs-preserved by diffing before/after values; on a same-SHA re-run (existing `devflow_version` already equals the SHA being installed) that produced "kept as a deliberate pin", contradicting the python3 backend's "pinned" message for the identical input. Now classifies on the existing value directly, same as python3. Added regression coverage for the idempotent case on both backends.
2. **python3 backend test coverage gap** — the preserve (`main`) / re-stamp (`abc1234`) assertions only ran in the jq test block. Mirrored both into the python3 test block (which shadows jq off PATH).

Deferring the three Suggestion-level notes (not implementing, per the already-approved verdict):
- Boundary/edge-case matrix (uppercase hex, length boundaries, tag pins) — real coverage idea, but not a correctness defect in this diff; a good candidate for a follow-up hardening pass if this parser sees more edge cases in practice.
- Hand-pinned SHA silently advanced on re-run — acknowledged as a narrow, intentional consequence of "empty-or-SHA re-stamps"; the PR's own comment already documents this trade-off. Not fixing since a SHA-valued `devflow_version` is definitionally indistinguishable from an auto-stamp.
- Docs wording nit (lowercase-only regex, "bump to a newer SHA" guidance interacting with preserve framing) — cosmetic; deferring rather than risk widening this diff. Low value fix if anyone touches `docs/cloud-setup.md` next.

Local suite (`lib/test/run.sh`): 2243 passed, 0 failed. Shellcheck: clean on both edited files.

The01Geek:
Addressed the REJECT review's two blocking findings (commit df4db7d):

1. **Critical — unguarded `mv` reports false success** — both backends `;`-chained `mv "$tmp" "$cfg"; log "pinned..."; return 0` unconditionally, so a failed move (read-only `.devflow/`, ENOSPC, cross-device rename) still logged success while leaving the config unpinned. `mv` is now the condition of its own `if`; a failed move falls through to the existing generic warning.
2. **Important — jq recheck conflated a real error with a legitimate false result** — `jq -e`'s eligibility recheck treated any non-zero exit (a genuine false/null, exit 1, OR an actual jq parse/runtime error, observed exit 5) as "deliberate pin". It now distinguishes them the same way the python3 backend already does via its explicit `sys.exit(3)` gate.

Both had to stay inside `if`-condition guards rather than a bare statement + `local rc=$?` — install.sh runs under `set -euo pipefail`, and a bare failing command outside an if/while/&&/|| context aborts immediately under `set -e`. My first pass at the jq fix used a bare `jq -e` statement and hit exactly this: the function returned the recheck's raw exit code instead of 0, and the "kept existing... deliberate pin" message never printed. Caught it by running the new tests directly rather than trusting the manual smoke check, reworked to mirror the already-safe python3 pattern.

Added regression coverage in `lib/test/run.sh`: a failed `mv` for both backends (asserts return 0, falls through to the warning, never logs "pinned", on-disk value untouched), and a genuine jq recheck error via a fault-injecting `jq` stub (asserts return 0, falls through to the warning, is NOT misreported as "kept... deliberate pin").

Local suite (`lib/test/run.sh`): 2255 passed, 0 failed. Shellcheck: clean on `install.sh`.

The three Suggestion-level notes from this round remain deferred per the prior comment's rationale (boundary/edge-case matrix, hand-pinned-SHA-advances-on-re-run, docs wording) — none are correctness defects in this diff.

The01Geek:
Also closed the deferred **Suggestion-2** note (commit b49d04c): the "hand-set non-SHA (main) is preserved" tests, both backends, asserted only the on-disk value, never the "kept existing...deliberate pin" message or return code.

Elevating this from deferred to fixed because it's no longer theoretical: my first draft of the jq Important-finding fix (never pushed) had a `set -e` abort bug — the function returned a leaked non-zero exit code and never printed the "kept" message — and this exact preserve-path test still reported PASS, because the on-disk value happened to stay unchanged either way. Only a differently-scoped new test caught it. Added RC=0 + log-message assertions to both backends' preserve-path tests so a similar regression can't hide behind "the value didn't change" again.

Local suite: 2259 passed, 0 failed.

Remaining deferred: the boundary/edge-case matrix, hand-pinned-SHA-advances-on-re-run, docs wording, and the python3 `re.match`→`re.fullmatch` trailing-newline parity gap — none are correctness defects in this diff.

Issue #284, opened 2026-07-03
Portability wave 3: migrate the multi-statement rc-capture guard recipes off the inline-bash variable-stripping hazard


The01Geek:
## Problem Statement

Follow-up to #275. Wave 2 (PR #279) made every local-tier skill's *bundled-helper anchor* a single-statement portable form, so the skills now start and resolve helpers on GitHub Copilot CLI / Windows. But the skill bodies still carry multi-statement guard recipes — `VAR=$(…); VAR_RC=$?` capture-then-discriminate blocks — whose later-statement variable reads the same inline-bash marshaling strips. On such a runner those guards read empty, their rc-discriminating branches (and `::warning::` breadcrumbs) silently never fire, and value variables can be left empty rather than at their documented defaults. PR #279 documented this as a known residual (`docs/install.md`, CHANGELOG 2.8.60); this issue is the tracked fix.

## Current Behavior

- `skills/review-and-fix/SKILL.md` (`MAX_ITERS`/`FIX_THRESHOLD`/`ENABLED` reads), `skills/review/SKILL.md` (`VERDICT_THRESHOLD`, `WP=…; rc=$?`), `skills/receiving-code-review/SKILL.md` (`REOPEN_THRESHOLD_RC` — whose `*` breadcrumb arm also misreports a stripped-empty rc as a resolver failure), and `skills/implement/phases/phase-4-documentation.md` (`GH_RC`/`HELPER_RC`, `DEFERRED_LABELS_RC` — the highest-blast-radius instance: the Phase 4.1 documentation gate's fail-closed empty-stdout check is inert on a stripping runner) all read an rc/value variable in a later statement of the same inline command.
- `skills/retrospective-weekly/SKILL.md`'s Stage-B validity check now uses a single-statement `[ ! -e ]` wrapper precheck (fixed in PR #279), but that precheck misses a present-but-non-executable wrapper (rc 126) — the #247 existence-vs-executability class; tighten to `[ ! -x ]`.
- `scripts/workpad.py`'s in-process config read is cwd-relative (deliberate parity with `config-get.sh`); a run invoked from a repo subdirectory silently misses a configured custom marker.

## Desired Behavior

[...]

Issue #275, opened 2026-07-02
Make the local-tier skills run first-class on GitHub Copilot CLI / Windows (portability wave 2)


The01Geek:
## Problem Statement

DevFlow is intended to run first-class on every major agentic CLI, not only Claude Code. A live `/devflow:implement` run on **GitHub Copilot CLI 1.0.68** (Windows 11, WSL bash, real `gh.exe`, Windows Python 3.13) — captured in `.devflow/tmp/devflow-copilot-cli-feedback.md` — **halted before Phase 1** and then broke several documented DevFlow mechanisms. The first wave of portability work (#241, #245, #247, #248) shipped the resolver family (`DEVFLOW_GH`, `DEVFLOW_JQ`, `DEVFLOW_BASH`, `normalize-path.sh`) and a portable skill-dir anchor recipe, but wired the anchor recipe into **only** `skills/create-issue/SKILL.md` and left the generalization to the other skills deferred. This wave closes the gaps that block a real run on a non-Claude-Code runner.

The pain lands on any user driving the local/interactive tier on Copilot CLI (and, by class, Cursor / Codex CLI / Gemini CLI), especially on Windows: today `/devflow:implement` is effectively unusable there without expert manual path substitution.

## Current Behavior

1. **Skill-dir anchor is not generalized (🔴 blocks Phase 1).** 21 skill/phase files — including `skills/implement/SKILL.md` and all four `skills/implement/phases/phase-*.md` — still use the bare `${CLAUDE_SKILL_DIR}/../../scripts/…` form. Under Copilot CLI, `${CLAUDE_SKILL_DIR}` expands **empty** (Copilot injects the location as a context header line `Base directory for this skill: …` instead). The implement entry-gate treats an empty anchor as fail-closed and **stops**; even ignoring the stop, every phase-file read and helper path collapses to a broken `/../../scripts/…`.

[...]

PR #299, opened 2026-07-04
Resolve .devflow/ from the repo root so config & prompt extensions load from a subdirectory

### LINKED ISSUE #295: Resolve `.devflow/` from the repo root so config & prompt extensions load when a skill runs from a subdirectory
## Problem Statement
A DevFlow consumer running any `/devflow:*` skill with a **current working directory below the git repo root** silently loses every committed customization. A ShipRush maintainer hit this: repo root is `C:\Src\shiprush-web\`, but they invoke skills from the nested project directory `C:\Src\shiprush-web\shiprush\Current\WebApplications\ShipRushWeb\`. From there, their `.devflow/prompt-extensions/create-issue.md` is never applied and their `.devflow/config.json` is never read — with no error and no warning. This layout (a nested "solution/project root" far below the git root) is the norm in large enterprise repos, which is exactly where DevFlow's wedge lives. The failure is silent, so the consumer only discovered it by noticing their `create-issue` title customization stopped taking effect.

## Current Behavior
Four readers resolve the `.devflow/` directory **relative to the current working directory** rather than to the repo root:

- `scripts/load-prompt-extension.sh` — `ext_file=".devflow/prompt-extensions/${skill}.md"`. When the file is not found at the cwd-relative path it prints nothing and exits 0 — the no-op path the calling skill reads as "no extension, proceed unchanged." The consumer's committed extension is silently discarded.
- `scripts/config-get.sh` — the default config path is the bare `.devflow/config.json`. When absent at cwd it returns the caller-supplied default, so `base_branch`, `effort`, `allowed_tools`, `workpad_marker`, doc paths, and every other key silently revert to fallbacks.
- `scripts/workpad.py` — reads `Path('.devflow/config.json')` in-process (the marker read), cwd-relative for the same reason.
- `scripts/match-deferrals.py` — `DEFAULT_CONFIG = ".devflow/config.json"`, passed as the **explicit** config-path argument to `config-get.sh` when reading `.devflow.allowed_bots`. This one is subtle: because it passes the cwd-relative path *explicitly*, root-anchoring `config-get.sh`'s **default** does not reach it — it needs its own fix, or it stays broken behind the "explicit path honored verbatim" contract.

The loader's own guards (broken symlink / non-regular / unreadable) can't catch this — they only fire once a file is *found* at the cwd-relative path; a wrong base directory slips past all of them. CI never sees it because the checkout cwd equals the repo root, so the bug hides from the test suite (`lib/test/run.sh` runs from repo root) and only bites real local/subdirectory use.

Reproduction (real repo, `.devflow/` at git root, launched from a nested project dir):

| Invocation cwd | `load-prompt-extension.sh create-issue` |
| --- | --- |
| git repo root | extension loaded (non-empty output) |
| nested project dir | 0 bytes, exit 0 (silently skipped) |

## Desired Behavior
All four readers resolve `.devflow/` **anchored to the git repo root** — `git rev-parse --show-toplevel`, falling back to `pwd` (bash) / `Path.cwd()` (Python) when the tree is not a git repo. A skill invoked from any subdirectory loads the consumer's `.devflow/prompt-extensions/<skill>.md` and `.devflow/config.json` exactly as if it were invoked from the repo root. When cwd already **is** the repo root, resolution is byte-for-byte unchanged. `config-get.sh` (bash), `workpad.py`, and `match-deferrals.py` (Python subprocesses) perform the *same* discovery so they always resolve to the same file, preserving lockstep. When no git root **and** no `.devflow/` can be located, each reader emits a one-line stderr breadcrumb before falling through to defaults, so the silent-drop class cannot recur unnoticed.

## User Impact
Every consumer whose repo layout puts the working directory below the git root — the common enterprise/monorepo case — gets their committed DevFlow customizations honored instead of silently ignored. It also aligns the four readers with `lib/config-source.sh`, which **already** resolves the repo root and passes it as an absolute path to `config-get.sh`, removing an internal split-brain where two config resolvers disagree about where `.devflow/config.json` lives.

## Technical Context

- **Relevant Classes/Files**
  - `scripts/load-prompt-extension.sh` — anchor `ext_file` to the repo root.
  - `scripts/config-get.sh` — anchor the default `config_file` to the repo root (still honoring an explicit `CONFIG_FILE` 3rd argument verbatim).
  - `scripts/workpad.py` — anchor the in-process `.devflow/config.json` marker read to the repo root via a `git rev-parse --show-toplevel` subprocess (`Path.cwd()` fallback), mirroring the bash discovery. It already imports `subprocess` and has a `_run` helper, so this is a same-shape addition.
  - `scripts/match-deferrals.py` — build its `DEFAULT_CONFIG` path from the git root (via the same guarded `git rev-parse` subprocess) instead of the bare cwd-relative literal, so its explicit-arg call to `config-get.sh` resolves the root config. It already locates `config-get.sh` via `Path(__file__)` and uses `_run`.
  - `lib/config-source.sh` — the **existing reference implementation**: `_DEVFLOW_REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"`, passed to `config-get.sh` as an absolute 3rd arg. Callers going through `config-source.sh` are already correct; the bug is the **direct** (2-arg) `config-get.sh` calls skills make, and the explicit-cwd-path call `match-deferrals.py` makes. Make the four readers consistent with this reference; do not diverge the discovery expression.
- **Architecture Alignment** — `lib/efficiency-trace.sh` and `lib/fetch-pr-context.sh` already resolve paths through `devflow_repo_root()`; this change extends the same repo-root anchoring to the four cwd-relative readers.
- **Dependencies** — `git` (already a hard DevFlow preflight prerequisite via `lib/preflight.sh`); `python3` (already required) for the Python readers' subprocess.
- **Data/Schema Considerations** — none; `.devflow/config.json` shape and keys are unchanged.
- **Cross-layer Impact** — shell helpers + two Python helpers; no cloud-workflow change (workflows use the vendored `.devflow/vendor/devflow/scripts/…` path and run at the checkout root).

**Windows note (supporting, not a risk).** The `#275` rationale for keeping `workpad.py` off `config-get.sh` was that Windows cannot *exec* a `.sh` ([WinError 193]). Resolving the root via a **`git.exe` subprocess** sidesteps that entirely — it is a native-binary call, exactly like the existing `gh` subprocess `workpad.py` already makes — so the Python readers stay Windows-safe. `git rev-parse --show-toplevel` returns a forward-slash path on Git Bash/WSL that both bash and Python `Path` consume directly, and `config-source.sh` already relies on this on all platforms. Confirm the bash and Python resolutions produce byte-identical paths on a Windows runner before closing — assumption, confirm before implementing.

**Overrides the deliberate `#275` cwd-relative contract — update all mirror sites in the same change.** The current cwd-relative behavior is documented as intentional (`config-get.sh` header "SHARED CWD-RELATIVE CONFIG CONTRACT", the `workpad.py` header, and `CLAUDE.md` line 33, all tied to issue #275). This is a coupled invariant: the behavior change and every doc surface that describes the old behavior must move together, and any `lib/test/run.sh` pin that asserts the old contract must be reconciled in the **same** commit — never a follow-up.

**Known limitation of the git-root approach.** Anchoring to `git rev-parse --show-toplevel` fixes the reported case (a plain subdirectory of a repo whose `.devflow/` sits at the git root). It does **not** cover: (a) a nested project dir that is itself a git **submodule** or nested repo — `--show-toplevel` returns that inner root, not the superproject where `.devflow/` lives; (b) a monorepo where `.devflow/` is deliberately **not** at the git root. A walk-up-to-nearest-`.devflow/` resolver would cover both but was declined for this fix in favor of consistency with `config-source.sh`. State this limitation in the docs so the submodule/monorepo case is not assumed fixed.

**Related, but out of scope (tracked separately).** The `.devflow/tmp/` **write** paths are also split-brained: `efficiency-trace.sh`, `fetch-pr-context.sh`, and `phase-3-review.md` anchor to `$ROOT`, but `phase-4-documentation.md` (`SLUG_DIR=".devflow/tmp/review/pr-${PR_NUMBER}"`), `pr-description/SKILL.md`, and create-issue's own `.devflow/tmp/` scratch writes are cwd-relative — so from a subdirectory a writer and a reader of the same review-artifact tree can diverge. This issue is scoped to the **config/extension readers** (the reported symptom). The write-path divergence is a larger, mostly-cloud surface that touches multiple `SKILL.md` bodies (requiring the `writing-skills` RED/GREEN discipline) and warrants its own follow-up issue; it is called out here so it is not silently dropped, not folded into this fix.

## Acceptance Criteria
- [ ] `load-prompt-extension.sh <skill>` invoked with cwd set to a nested subdirectory of a repo whose `.devflow/prompt-extensions/<skill>.md` exists prints that file's contents verbatim (identical to invoking it from the repo root).
- [ ] `config-get.sh <key> <default>` (no explicit config-file argument) invoked from a nested subdirectory reads the repo-root `.devflow/config.json` and returns the configured value, not the default.
- [ ] `workpad.py` resolves its `.devflow.workpad_marker` from the repo-root `.devflow/config.json` when invoked from a nested subdirectory, resolving to the **same** file `config-get.sh` resolves for the same cwd (lockstep preserved).
- [ ] `match-deferrals.py` resolves `.devflow.allowed_bots` from the repo-root `.devflow/config.json` when invoked from a nested subdirectory (its explicit config-path argument no longer defeats the root anchoring).
- [ ] When cwd **is** the repo root, all four readers behave byte-for-byte as before this change (no regression on the common path).
- [ ] In a non-git tree, all four readers fall back to `pwd` / `Path.cwd()` and still resolve a `.devflow/` present at that location.
- [ ] When neither a git root nor a `.devflow/` directory can be located, each reader writes a single non-empty stderr breadcrumb before using defaults (the fallback is no longer silent).
- [ ] An explicit `CONFIG_FILE` third argument to `config-get.sh` (e.g. the absolute path `config-source.sh` passes) is still honored verbatim (repo-root anchoring applies only to the default).
- [ ] The `#275` cwd-relative doc contract is updated to describe the repo-root behavior in `config-get.sh`'s header, `workpad.py`'s header, and `CLAUDE.md`, and documents the submodule/non-root-`.devflow/` limitation; any `lib/test/run.sh` pin asserting the old contract is reconciled in the same change.
- [ ] `bash lib/test/run.sh`, `shellcheck`, and `ruff` all pass (the `lib + python tests` CI job is green).

## Implementation Notes
- **Approach** — In `load-prompt-extension.sh` and `config-get.sh`, compute the root inline exactly as `config-source.sh` does — `root="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"` — and prefix the `.devflow/…` path with it (for `config-get.sh`, only when no explicit 3rd-arg `CONFIG_FILE` was passed). In `workpad.py` and `match-deferrals.py`, resolve the root with a guarded `subprocess`/`_run` call to `git rev-parse --show-toplevel` (`Path.cwd()` on failure) and build the config path from it, keeping the discovery semantically identical to the bash side. Do **not** make `config-get.sh` source `config-source.sh` — it is invoked directly by skills as the command's leading token under the read-only cloud allowlist, and must stay standalone. Add the stderr breadcrumb on the final no-`.devflow/` fallback in each reader.
- **Code Patterns** — mirror `lib/config-source.sh:14` verbatim for the bash discovery expression (note: `|| pwd`, not `$PWD`); reuse `workpad.py`'s existing `_run` helper for the Python `git` subprocess; mirror the best-effort stderr-breadcrumb style already used across the `resolve-*.sh` family for the fallback warning.
- **Testing Strategy**
  - **Move 1 — boundary & levels.** Every reader has an automated boundary: `load-prompt-extension.sh` and `config-get.sh` are exercised by shell assertions in `lib/test/run.sh` (stdout + exit code); `workpad.py`'s and `match-deferrals.py`'s config resolution are Python-visible return values. Level: shell integration tests for the two `.sh` readers, plus direct-invocation tests for the two Python readers' resolution. These run against the **real** scripts (no mocking of the unit under test).
  - **Move 2 — coverage dimensions.**
    - *Happy path:* create a temp git repo with `.devflow/config.json` + `.devflow/prompt-extensions/<skill>.md` at the root, `cd` into a nested subdir, and assert each of the four readers resolves the root file.
    - *Boundary / degenerate:* cwd **is** the root (must be unchanged); nested subdir exactly one level down and several levels down.
    - *Error / fallback:* non-git tree with a `.devflow/` at cwd (falls back to `pwd`/`cwd` and resolves); neither git root nor `.devflow/` present (asserts the stderr breadcrumb fires and defaults are used).
    - *Adversarial:* an explicit `CONFIG_FILE` 3rd arg to `config-get.sh` from a subdir is still honored verbatim (root anchoring must not override it); and confirm `match-deferrals.py`'s explicit-path call now resolves the root config rather than being defeated by that same contract.
    - *State / lockstep:* from the same subdir, assert `config-get.sh`, `workpad.py`, and `match-deferrals.py` resolve the identical absolute config path.
  - **Move 3 — named assertions (test-first; each fails on today's code for the right reason — the defect reproduces).**
    - `assert "loader from subdir emits root extension bytes"` — fails today because the loader prints 0 bytes from a subdir.
    - `assert "config-get from subdir returns configured value not default"` — fails today because it returns the default.
    - `assert "workpad marker from subdir == root marker"`, `assert "match-deferrals allowed_bots from subdir == root value"`, and `assert "all readers resolve identical config path from subdir"` — the lockstep set; fails today because each silently misses the root config.
    - `assert "cwd==root behavior byte-identical"` — guards the no-regression path.
    - `assert "no-.devflow fallback writes a non-empty stderr breadcrumb"` — the guarantee-class assertion: it proves the diagnostic fires on the path where resolution failed, which is the whole reason the breadcrumb exists.
    - `assert "explicit CONFIG_FILE arg still honored from subdir"`.
    - Fixtures: temp git repos built with `git init` in the existing `lib/test/run.sh` sandbox style (the `#161 git_sandbox` helper); nothing about git or the filesystem is mocked, since the resolution *is* the unit under test.
  - Every AC above maps to at least one of these assertions; no assertion is orphaned.
- **Documentation Needed** — update the `#275` contract prose in `config-get.sh`'s header, `workpad.py`'s header, and `CLAUDE.md` (via `/claude-md-management:revise-claude-md`) to describe repo-root anchoring **and the submodule/non-root-`.devflow/` limitation**; add a line to `docs/install.md` noting subdirectory invocation now works. Confirm `docs/DEVFLOW_SYSTEM_OVERVIEW.md` has no cwd-relative claim to correct. The standard patch version bump + `CHANGELOG.md` entry is handled automatically by `/devflow:implement` (via `.devflow/prompt-extensions/implement.md`) — no manual step here.
- **Potential Gotchas**
  - `config-get.sh` must stay standalone (do not source `config-source.sh`) — the read-only cloud allowlist grants it only as the command's leading token.
  - `config-get.sh` is a fresh subprocess invoked many times per skill run, so it forks `git rev-parse` on every call (fast, and `git` is a hard preflight prereq, so acceptable). Callers going through `config-source.sh` pay the discovery only once (it is cached in a variable) — do not "optimize" `config-get.sh` by caching across its own separate invocations; it cannot.
  - Keep the bash and Python discovery semantically identical, or the readers desync — the exact failure `#275`'s lockstep clause guards against.
  - `match-deferrals.py` passes the config path *explicitly* to `config-get.sh`, so it is **not** fixed by the `config-get.sh` default change alone — it must root-anchor its own `DEFAULT_CONFIG`.
  - Do not touch the cloud-tier workflows' vendored `.devflow/vendor/devflow/scripts/…` path — they run at the checkout root and are unaffected.

The01Geek:
## Summary
- DevFlow resolved `.devflow/` relative to the current working directory, so a `/devflow:*` skill invoked from a subdirectory below the repo root silently lost the consumer's committed config and prompt extensions (no error, invisible to CI whose cwd is the repo root).
- The four `.devflow/` readers now anchor the **default** `.devflow/` path to the git repo root (`git rev-parse --show-toplevel`, falling back to `pwd`/`Path.cwd()`), mirroring `lib/config-source.sh`, so a subdirectory invocation loads the root config/extension exactly as if run from the root.
- Explicit config-path args stay honored verbatim; cwd-at-root is byte-for-byte unchanged; a no-git-root-and-no-`.devflow/` tree now emits a stderr breadcrumb so the silent-drop class cannot recur unnoticed.

## Changes
**`scripts/config-get.sh`**: Anchor the default config path to the git repo root; honor an explicit (non-empty) `CONFIG_FILE` 3rd arg verbatim (empty 3rd arg still means "use the default", now root-anchored); breadcrumb only when neither a git root nor a `.devflow/` is found.
**`scripts/load-prompt-extension.sh`**: Anchor `ext_file` to the repo root with the same set -e-safe discovery + breadcrumb.
**`scripts/workpad.py`**: Add `_repo_root()` (native `git` subprocess, Windows-safe; catches `CalledProcessError`/`OSError` → cwd fallback) and anchor the in-process `_workpad_marker` config read to it.
**`scripts/match-deferrals.py`**: Replace the module-level `DEFAULT_CONFIG` literal with a lazy `_default_config_path()`; `--config` defaults to `None` and resolves the root-anchored default in `main`, so its explicit-arg passing no longer defeats root anchoring.
**Docs & contract**: Rewrite the `#275` cwd-relative contract in the `config-get.sh` / `workpad.py` headers and `CLAUDE.md` to the repo-root behavior (with the submodule/non-root-`.devflow/` limitation); note subdirectory invocation in `docs/install.md` + a release note; reconcile the `#284`→`#295` `lib/test/run.sh` pins in the same change.
**Version**: Patch bump `2.8.65` → `2.8.66` with a matching `CHANGELOG.md` entry.

## Resolves
Resolves #295

## Test Plan
- [ ] `bash lib/test/run.sh` passes (the new `#295` block: subdir loader/config-get/workpad-marker/match-deferrals resolution incl. end-to-end value + lockstep, cwd==root byte-identical, non-git pwd fallback, all-four-readers breadcrumbs, explicit `CONFIG_FILE` honored; plus the reconciled `#295 AC9` presence/absence pins).
- [ ] `shellcheck --severity=warning -e SC1091` on `scripts/config-get.sh` + `scripts/load-prompt-extension.sh` is clean.
- [ ] `ruff check` on the Python readers is clean.
- [ ] From a nested subdirectory of a repo with a root `.devflow/config.json`, `config-get.sh .base_branch main` returns the configured value (not the default).

## Visual Changes
N/A

## Breaking Changes
None. Root-anchoring changes only the default resolution; explicit config-path args are honored verbatim and cwd-at-root behavior is byte-for-byte unchanged.


The01Geek (commented):
## Verdict: APPROVE with notes (repo-root `.devflow/` anchoring — correct, well-tested; only advisory wording/coverage refinements)

## Issue Compliance
Reviewed against issue #295: *Resolve `.devflow/` from the repo root so config & prompt extensions load when a skill runs from a subdirectory*. The four `.devflow/` readers now anchor the **default** path to the git repo root; delivered as specified.

## Verification
- **Full test suite green on this branch:** `lib/test/run.sh` → **3292 passed, 0 failed** (incl. the new `#295` subdir-resolution block, the `#295 AC9` reconciled pins, and the `test_python_scripts.py` `#295` marker test). The pre-existing `main` CI RED (app-token pin desync, unrelated to this diff) does **not** affect this branch.
- **Version gate:** `.claude-plugin/plugin.json` `2.8.67` matches the new `CHANGELOG.md [2.8.67]` entry ✓.
- **Diff classification:** `engine_self_modifying` (touches `lib/test/**`) → full checklist + always-on agents; `has_new_types=false` (type-design skipped); test files changed → pr-test-analyzer dispatched.

## Code Review Findings
Five Phase-3 agents returned; all findings are **single-source** (corroboration 1/5) and **below the `critical` verdict threshold** — none drives a REJECT. No self-contradicting-diff finding (the wording items below are precision refinements, not falsehoods contradicting the code's intent).

### 🟠 Important / Major
1. `config-get.sh` header/contract (and the mirrored CHANGELOG / CLAUDE.md / `docs/install.md`) say an explicit `CONFIG_FILE` "is honored **verbatim**" without qualification, but the gate is `[ -n "${3:-}" ]` — an explicitly-passed **empty** 3rd arg falls through to the root-anchored default. The inline code comment documents this correctly, so the header wording is in mild tension with it. Suggest: "a **non-empty** explicit `CONFIG_FILE` is honored verbatim (an explicit empty string still selects the root-anchored default)." (comment-analyzer, 1/5)
2. `match-deferrals.py`'s replacement comment claims `--config` "mirrors config-get.sh's explicit-CONFIG_FILE contract"; on the empty-string edge it doesn't mirror — an empty `--config` is re-defaulted by config-get.sh's own fallback rather than honored verbatim. Same precision nit as (1). (comment-analyzer, 1/5)
3. **Test gap — no assertion that a git-root-with-no-`.devflow/` tree stays SILENT.** Every reader deliberately emits its breadcrumb *only* when NEITHER a git root NOR a `.devflow/` is found; the tests pin the positive (bare-tree → breadcrumb, AC7) but not the negative branch. A regression moving the `.devflow`-existence check outside the `if [ -z "$_devflow_root" ]` guard would spam stderr on **every normal unconfigured in-git run** with no test going RED. Worth adding a silence assertion (empty stderr + default value) for the four readers. (pr-test-analyzer, 1/5)

### 🟡 Suggestion / Minor
4. Non-128 `git rev-parse` failures (e.g. `safe.directory`/dubious-ownership refusals) are reclassified as "not in a git repo" and fall back to cwd — the same silent-config-drop class #295 fixes, narrowed to a git-config edge, with a breadcrumb that misstates the cause. Not a regression (pre-#295 degraded similarly). Consider softening the breadcrumb to "could not resolve a git repo root" or surfacing git's stderr on a non-128 failure. (silent-failure-hunter, 1/5)
5. `match-deferrals.py` breadcrumb text says "cwd-**relative** default config path" but returns an absolute `Path.cwd()/…` path; sibling breadcrumbs are consistent with their code — this one's wording is the outlier. (comment-analyzer, 1/5)
6. Optional test-completeness: `--config` explicit-honored unpinned for match-deferrals; AC5 cwd==root identity covers only 2 of 4 readers; AC6 non-git pwd-fallback covers only config-get. All low-value (adjacent branches cover the same code paths). (pr-test-analyzer, 1/5)

over-grade annotation: no finding flagged.

## Verdict Criteria
- No checklist FAIL/INCONCLUSIVE (suite green).
- No self-contradicting-diff finding.
- No finding at or above the `critical` threshold → not a REJECT.
- Findings below threshold present → **APPROVE with notes.**

**Strengths (from all five agents):** symmetric, correct fix across all four readers; `set -e`/`set -u`-safe rc capture (`… || _devflow_root=""`, `[ -n "${3:-}" ]`); Windows-safe native `git` subprocess (not a `.sh` exec, per #275); breadcrumbs fail loud only in the genuine no-root-no-`.devflow` case; coupled-invariant mirror sites (headers, CLAUDE.md, docs, `run.sh` pins) updated in lockstep with the old cwd-relative literals asserted GONE; tests assert VALUES (symlink-robust) and include an end-to-end `_config_get(None)` read.

None of the notes block merge. Recommend addressing (1)/(2) (duplicated across four surfaces) and (3) (guards a user-visible stderr-spam regression) as a small follow-up if desired.

PR #367, opened 2026-07-09
Eliminate no-verdict cloud review runs by grounding the review engine in its allowed tools and CI results

### LINKED ISSUE #363: Eliminate no-verdict cloud review runs by grounding the review engine in its allowed tools and CI results
## Problem Statement

The cloud review engine can consume an entire run — 60 turns, 8.9 minutes, $11.64 — and terminate having produced **no verdict at all**. The maintainer is left with a red, required `Devflow Review` check that carries no findings, no verdict, and no indication of what went wrong. The only remedy offered is a Re-run, which hits the same wall deterministically.

This happened on PR #340 (run 29027928470). The `review / run` job reported `success`; the engine's own result event reported `subtype: "success"`, `is_error: false`, `num_turns: 60`, `permission_denials_count: 14`. Seven of those 14 denials were the engine trying, seven different ways, to execute `lib/test/run.sh` — a command the auto-review tool profile can never permit. The engine stopped after Phase 2 with its progress comment frozen at `**Status:** 🚀 Reviewing`.

The root cause is that **the review engine has no ground truth about its own capabilities, and none about the CI results that already exist for the commit it is reviewing.** It guesses at both, and it guesses wrong in a way that silently consumes the run.

## Current Behavior

**The engine does not know which commands it may run.** `devflow-runner.yml` resolves an explicit `--allowed-tools` string (the `review` profile, `devflow-runner.yml:298`) and passes it to `claude_args` (line 585). `devflow.yml` builds a separate allowlist for the light command path and passes it to its own `--allowed-tools` (line 564). Neither string is ever shown to the agent. The agent discovers its boundary by attempting commands and collecting denials. Claude Code's permission documentation is explicit that this is the expected outcome: *"Permission rules are enforced by Claude Code, not by the model. Instructions in your prompt or `CLAUDE.md` shape what Claude tries to do, but they don't change what Claude Code allows."* Nothing currently shapes what the agent tries.

**The engine does not know CI already ran.** `devflow-review.yml` triggers the review on `workflow_run` for `workflows: [CI], types: [completed]`, and `scripts/derive-review-preconditions.sh` enforces `require_ci_green` (default `true`, `.devflow/config.example.json:38`), which defers the review until every other CI signal on the head has completed without failing. On the auto path, CI is therefore guaranteed complete and non-failing before the agent takes its first turn. The agent is never told this and spends turns trying to re-derive it by running the suite itself.

**The same skill runs under two profiles that disagree.** `skills/review/SKILL.md` is executed by the auto-review path (`devflow-runner.yml`'s `review` profile) and by the manual `/devflow:review` comment path (`devflow.yml`'s `claude` job, line 513). `devflow.yml`'s allowlist grants `Bash(mkdir:*)` and `Bash(tee:*)`; `devflow-runner.yml`'s does not. Claude Code splits compound commands on `&&`, `||`, `;`, `|`, `|&`, `&`, and newlines, and requires every subcommand to match a rule independently. So Phase 0.2's `mkdir -p …` followed by `gh pr diff … | tee …/diff.patch` (`skills/review/SKILL.md:231-232`) succeeds on the manual path and is denied on the auto path. The cached `diff.patch` is therefore never created during an auto review, and Phase 3 substitutes `{DIFF_PATH}` into its agent prompts pointing at a file that does not exist. The denial appears in run 29027928470 and in run 29025850980, which nonetheless reached APPROVE.

**Two review phases have never executed in cloud.** `skills/review/SKILL.md` invokes `git cat-file -e` in Phase 0.3.6's blocker-recheck precondition and `git checkout` in Phase 3.1's dirty-tree restore. The `review` profile grants `git status`, `git diff`, `git log`, `git show`, `git ls-files`, `git rev-parse`, `git merge-base`, `git blame`, and `git branch` — neither `git cat-file` nor `git checkout`. Phase 0.3.6 therefore falls through to the full pipeline on every cloud run, and Phase 3.1's backstop cannot restore a path an agent dirtied.

**A bundled-helper invocation is denied by path shape.** `skills/review/SKILL.md` is read as a file from `.devflow/vendor/devflow/skills/review/`, so its `${CLAUDE_SKILL_DIR}` anchor resolves to the allow-listed `.devflow/vendor/devflow/scripts/…` prefix. The final-pass reviewer `skills/requesting-code-review/SKILL.md:13` is dispatched as an installed **Skill**, so its anchor resolves to the repository checkout and its `load-prompt-extension.sh` invocation lands outside every allow-listed prefix. It was denied in both observed runs, meaning the consumer prompt extension silently never loads for the final-pass reviewer.

**A false comment records the opposite of the truth.** `devflow-runner.yml:579` asserts that `--permission-mode acceptEdits` "auto-approves Edit/Write plus the filesystem Bash commands mkdir/touch/rm/rmdir/mv/cp/sed" and that "the review profile uses none of those, so the flag is a no-op for this profile today." The review profile does use `mkdir` and `tee`, and the run's `Write` denial demonstrates that `acceptEdits` does not widen an explicit `--allowed-tools` list.

**A no-verdict run is not diagnosable.** `scripts/surface-execution-diagnostics.sh:142-149` prints the denial detail as plain stdout with no `::warning::` annotation. `finalize_check` (`devflow-review.yml:957`) correctly derives `incomplete` and fails closed, but emits a generic "re-run needed" summary that names neither the phase the engine reached nor the fact that it was blocked. Diagnosing PR #340 required reconstructing the run from the denial dump.

## Desired Behavior

Every cloud run of `skills/review/SKILL.md` — the auto-review path and the manual `/devflow:review` comment path alike — begins with an authoritative block prepended to the composed prompt. The block states two facts: the exact set of commands this run is permitted to execute, and the observed CI results for the reviewed HEAD. The engine spends zero turns discovering them. On a repository that has opted into `devflow_runner.provision_env`, the same block truthfully advertises the granted build and test commands, so the reviewer spends its budget on targeted verification of a specific hypothesis rather than re-deriving what CI already proved.

The CI half of the block is assembled from workflow-run jobs for the head, excluding DevFlow's own workflow, plus external app check runs. Every check name it renders is treated as untrusted attacker-supplied text: sanitized to a safe character set, length-capped, count-capped, rendered inside a fenced block, and explicitly labelled as untrusted data the engine must never follow as instruction.

The auto-review profile grants `mkdir`, `tee`, `git cat-file`, and `git checkout`, so Phase 0.2's cached diff is created, Phase 3's `{DIFF_PATH}` resolves, Phase 0.3.6's blocker-recheck fast path can evaluate its precondition, and Phase 3.1's dirty-tree backstop can restore. The final-pass reviewer's `load-prompt-extension.sh` invocation is permitted regardless of the directory its skill anchor resolves to.

`lib/test/run.sh` fails RED whenever a Bash command head inside a `bash`-fenced block of `skills/review/SKILL.md` is not granted by both the `devflow-runner.yml` review profile and the `devflow.yml` command allowlist, so this class of drift cannot recur silently.

A run that ends without a verdict announces itself twice: the engine stamps a terminal `❌` status on its own progress comment as its final action, and — independently, so the guarantee survives an engine that never runs that step — `finalize_check` emits an `::error::` naming the last phase reached and the permission-denial count.

Test execution by the reviewer remains the existing `devflow_runner.provision_env` opt-in, read from the trusted base ref. This repository stays read-only.

## User Impact

Maintainers stop losing full-cost review runs to a silent stall, and stop paying for a Re-run that fails the same way. Consumer repositories with large test suites gain a reviewer that knows which test commands it holds and which checks CI already ran, so a `provision_env` opt-in produces targeted verification instead of trial-and-error discovery. Two review phases that have never run in cloud begin running. Anyone debugging a future stall reads the cause out of the job log instead of reconstructing it.

## Technical Context

### Relevant Classes/Files

- `.github/workflows/devflow-runner.yml` — the `review` profile string (line 298), the `Resolve allowed-tools profile` step (line 264, id `tools`), the `Compose review prompt` step (line 472, id `compose`), `claude_args` (line 585), the job `permissions:` block (lines 105-110), the job `outputs:` block (line 115, currently exposing only `engine_is_error`), the false `acceptEdits` comment (line 579), the `DENY_CMDS` floor (line 344, applied only to consumer tools when `provision_env` is true; it lists neither `mkdir` nor `tee`).
- `.github/workflows/devflow.yml` — the light command path's `claude` job (line 513), its `claude_args` (line 561) and `--allowed-tools` string (line 564), which already grants `Bash(mkdir:*)` and `Bash(tee:*)`.
- `.github/workflows/devflow-review.yml` — `workflow_run` trigger on `workflows: [CI]`, the `precheck` job's `permissions:` block (lines 171-182, already declaring `checks: read`, `actions: read`, `statuses: read`), the runner caller job's permission grant (lines 902-907), the `startup_failure` note (lines 896-901), `finalize_check` (line 957, `name: Devflow Review` at line 980, declaring `contents: read` + `checks: write` + `pull-requests: write`).
- `scripts/derive-review-preconditions.sh` — the prior art for every CI query in this change: `SELF_WORKFLOW_NAME` exclusion (line 123), the `--paginate` + `jq -rs 'map(.check_runs // []) | add'` concatenated-array flatten, and the `app.slug != "github-actions"` / `name != "Devflow Review"` self-exclusion filter (lines 306-320).
- `scripts/derive-review-verdict.sh` — HEAD-scoped verdict lookup; fails closed to `incomplete`.
- `scripts/surface-execution-diagnostics.sh` — renders the run summary and denial detail (lines 142-149).
- `scripts/parse-engine-error.sh` — the existing precedent for extracting workflow-inline logic into a unit-tested helper.
- `lib/resolve-gh.sh`, `lib/resolve-jq.sh` — the execution-verified binary resolvers every new shell helper sources.
- `skills/review/SKILL.md` — Phase 0.2 scratch and diff caching (lines 231-232), Phase 0.3.6's `git cat-file -e` precondition, Phase 3.1's `git checkout` restore, the progress comment's `**Last updated:**` field (line 113), Phase 3's `{DIFF_PATH}` substitution, the Phase 4 telemetry step's guidance that the profile lacks `Bash(tee:*)`.
- `skills/requesting-code-review/SKILL.md` — line 13, the denied `load-prompt-extension.sh` invocation.
- `.devflow/config.example.json` — `devflow_review.require_ci_green` (line 38), `devflow_runner.provision_env` (line 25).

### Architecture Alignment

The injection reuses the existing precedent in `Compose review prompt`, which already prepends a `> [!IMPORTANT]` block to the base prompt when `steps.provision.outputs.health_summary` is non-empty. The CI summary is extracted into a new `scripts/summarize-ci-checks.sh`, matching the established pattern of `parse-engine-error.sh` / `derive-review-verdict.sh` / `derive-review-preconditions.sh`: a small always-exit-0 helper with stderr breadcrumbs, driven directly by `lib/test/run.sh` with a stubbed `gh` over an input-shape matrix. It routes `gh` through `lib/resolve-gh.sh` and `jq` through `lib/resolve-jq.sh`, per the execution-verified resolver convention.

The summary is computed generically, with no hardcoded job names, mirroring `derive-review-preconditions.sh`'s deliberately generic design and reusing its self-exclusion filters verbatim.

### Dependencies

`gh` (via `lib/resolve-gh.sh`), `jq` (via `lib/resolve-jq.sh`). No new external dependency.

### Data/Schema Considerations

No configuration schema change.

The capability half of the block is derived from the already-resolved `steps.tools.outputs.tools` output on the auto path and from the `--allowed-tools` string on the command path.

The CI half is assembled from two sources, mirroring `derive-review-preconditions.sh`:

1. `GET /repos/{owner}/{repo}/actions/runs?head_sha={sha}`, excluding runs whose workflow name equals `SELF_WORKFLOW_NAME`, then `GET /repos/{owner}/{repo}/actions/runs/{run_id}/jobs` for each surviving run, yielding per-job names and conclusions. This is what produces `lib + python tests`. Requires `actions: read`, which `devflow-runner.yml` already declares.
2. `GET /repos/{owner}/{repo}/commits/{ref}/check-runs`, filtered to `app.slug != "github-actions"`, yielding external app checks (Codecov, Vercel). Its `filter` parameter defaults to `latest` — verified against the GitHub REST documentation (https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28) — so one paginated call returns the latest run per check name. Requires `checks: read`.

Both queries use `--paginate`, whose output is concatenated JSON arrays, so results are flattened with `jq -s` before selection, exactly as `derive-review-preconditions.sh` does.

`devflow-runner.yml`'s job currently declares `contents: read`, `pull-requests: write`, `id-token: write`, `actions: read`, `issues: read`. GitHub's reusable-workflow documentation states that "permissions can only be maintained or reduced—not elevated—throughout the chain" (https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows), so `checks: read` is added to **both** `devflow-runner.yml`'s `permissions:` block and the calling job's grant in `devflow-review.yml`. `devflow-review.yml:896-901` records that a called workflow exceeding the caller's grant aborts the run as `startup_failure`.

**Check-run and job names are attacker-controlled.** A pull request can add a workflow file whose job `name:` is arbitrary text, and that string becomes a check-run name on the head. The auto-review job runs under `pull_request_target` for bot actors, holds `CLAUDE_CODE_OAUTH_TOKEN` and the DevFlow-Reviewer App token, and posts approvals and rejections. Names are therefore untrusted input on the same footing as the diff, and are sanitized before they reach the prompt.

The engine's `permission_denials_count` lives in the execution file inside the `review / run` job. That job's `outputs:` block (`devflow-runner.yml:115`) currently exposes only `engine_is_error`, and `finalize_check` runs in a separate job, so the count is plumbed through as a new job output rather than re-read from the runner's filesystem. `finalize_check` declares `contents: read`, `checks: write`, `pull-requests: write`, and job-level permissions replace workflow-level ones; reading the progress comment through `GET /repos/{owner}/{repo}/issues/{n}/comments` may additionally require `issues: read` — assumption, confirm before implementing. When the comment cannot be read, the annotation still names the HEAD SHA and the denial count.

### Cross-layer Impact

Workflow layer (`devflow-runner.yml`, `devflow-review.yml`, `devflow.yml`), helper-script layer (`scripts/`), skill layer (`skills/review/SKILL.md`), test layer (`lib/test/run.sh`), documentation layer (`docs/`).

### The CI summarizer

- [ ] `scripts/summarize-ci-checks.sh` exists, reads `REPO`, `HEAD_SHA`, and `SELF_WORKFLOW_NAME` from the environment, and prints one line per CI signal giving its name and its conclusion.
- [ ] It sources `lib/resolve-gh.sh` and `lib/resolve-jq.sh`, and `lib/test/run.sh` asserts it contains no bare `gh` and no bare `jq` invocation.
- [ ] It excludes every workflow run whose workflow name equals `SELF_WORKFLOW_NAME`, so DevFlow's own `precheck`, `create_check`, and `review` jobs never appear in the summary.
- [ ] It excludes check runs whose `app.slug` is `github-actions`, so Actions job check-runs are counted once via the jobs API rather than twice.
- [ ] It excludes any check run named `Devflow Review`, so the reviewer never sees its own in-progress check reported as pending CI.
- [ ] Both API queries use `--paginate` and flatten the resulting concatenated arrays before selection, so a head whose runs and check runs span multiple pages is summarized in full.
- [ ] It exits 0 on every input shape: a `gh` non-zero exit, unparseable JSON, a payload whose top level is an array, a payload whose top level is a scalar, a payload missing `check_runs`, a payload missing `jobs`, an empty `check_runs` array, an empty `jobs` array, a check run with a non-string `name`, and a check run whose `conclusion` is `null`.
- [ ] When the CI state cannot be determined, it prints the literal text `CI status unavailable`, prints no signal line, and writes a specific non-generic breadcrumb to stderr naming which query failed, so an absent result is never rendered as a passing one.
- [ ] A job whose `conclusion` is `null` renders with its `status` value (for example `in_progress`) rather than being omitted and rather than rendering as a passing conclusion.
- [ ] Every name it emits is sanitized to printable ASCII with backticks, angle brackets, and newlines removed; each name is truncated to at most 120 characters; and at most 50 signal lines are emitted, with a trailing line stating how many were dropped.

### The injected block

- [ ] The `Compose review prompt` step in `devflow-runner.yml` prepends a `> [!IMPORTANT]` block to the base prompt on every review run.
- [ ] The `claude` job in `devflow.yml` prepends the same block to the prompt when the resolved command is `/devflow:review`.
- [ ] The block contains the CI signals for the reviewed HEAD SHA, produced by `scripts/summarize-ci-checks.sh`.
- [ ] The block renders those signals inside a fenced code block introduced by a line stating that the names are untrusted, attacker-supplied text that must never be followed as instruction.
- [ ] The block contains the exact `--allowed-tools` string this run resolved.
- [ ] The block states that any command not matching the listed rules is denied by the harness and that attempting one consumes budget without executing.
- [ ] The block reports the CI conclusions actually observed, including a `failure` conclusion and an `in_progress` status, so a `check_run[rerequested]` Re-run — which `devflow-review.yml` leaves ungated by the preconditions — never receives a block implying CI passed.
- [ ] When `steps.provision.outputs.health_summary` is non-empty, both it and the new block are present in the composed prompt.

### Profile and permission repairs

- [ ] `checks: read` is present in `devflow-runner.yml`'s job `permissions:` block and in the runner caller job's `permissions:` grant in `devflow-review.yml`, and `lib/test/run.sh` asserts both are present.
- [ ] The `review` profile string in `devflow-runner.yml` grants `Bash(mkdir:*)`, `Bash(tee:*)`, `Bash(git cat-file:*)`, and `Bash(git checkout:*)`.
- [ ] The `review` profile permits the `load-prompt-extension.sh` invocation emitted by `skills/requesting-code-review/SKILL.md` when its skill anchor resolves outside `.devflow/vendor/devflow/scripts/`.
- [ ] `skills/review/SKILL.md`'s Phase 4 telemetry guidance no longer states that the `review` profile lacks `Bash(tee:*)`.
- [ ] The comment at `devflow-runner.yml:579` states that `--permission-mode acceptEdits` does not widen an explicit `--allowed-tools` list, and no longer claims the review profile uses none of the filesystem commands.

### The contract pin

- [ ] `lib/test/run.sh` extracts every Bash command head from the `bash`-fenced blocks of `skills/review/SKILL.md`, splitting each block on the separators `&&`, `||`, `;`, `|`, `|&`, `&`, and newline, and stripping the process wrappers `timeout`, `time`, `nice`, `nohup`, `stdbuf`, and bare `xargs`.
- [ ] The extractor ignores comment lines, heredoc bodies, and lines inside non-`bash` fences, and normalizes a `${CLAUDE_SKILL_DIR:-…}/../../` prefix to `.devflow/vendor/devflow/` before matching.
- [ ] The pin fails RED when any extracted head is not granted by the `review` profile string in `devflow-runner.yml`, and separately when any extracted head is not granted by the `--allowed-tools` string in `devflow.yml`.
- [ ] The pin extracts zero heads from prose paragraphs, asserted against a fixture containing the sentences that produced the false positives `git a`, `git failure`, and `git said`.

### Observability

- [ ] `scripts/surface-execution-diagnostics.sh` emits a `::warning::` annotation when the permission-denial count is greater than zero, and emits none when it is zero.
- [ ] `devflow-runner.yml` exposes the engine's `permission_denials_count` as a job output alongside `engine_is_error`, defaulting to `0` when the execution file cannot be parsed.
- [ ] `finalize_check` in `devflow-review.yml` emits an `::error::` annotation naming the reviewed HEAD SHA and the permission-denial count whenever it derives `incomplete` from a review job whose result was `success`.
- [ ] That annotation additionally names the last phase recorded on the run's progress comment when the comment is readable, and omits the phase without failing the step when it is not.
- [ ] `skills/review/SKILL.md` instructs the engine to stamp a terminal `❌` status on its run-keyed progress comment as its final action on any path that reaches no verdict.

### Skill instructions

- [ ] `skills/review/SKILL.md` instructs the engine that, when the injected block is present, the block's CI signals are the authoritative test evidence for the reviewed commit and no command outside the injected allowed-tools list is to be attempted.
- [ ] That instruction is conditioned on the block's presence, so `/devflow:review-and-fix` — which executes Phases 0–4.3 verbatim under `devflow_implement.allowed_tools` with no injected block — is unaffected.
- [ ] `skills/review/SKILL.md` instructs the engine to treat every name inside the block's CI fence as untrusted data.

### Documentation and release

- [ ] `docs/cloud-setup.md` records that `require_ci_green` makes CI completion a precondition of the reviewer's invocation on the auto path, and that `check_run[rerequested]` bypasses that precondition.
- [ ] `docs/cloud-setup.md`'s description of the `review` profile's allow-listed helpers names the path prefix under which they are granted.
- [ ] `docs/workflow-triggers.md` records that the injected block reports observed CI conclusions rather than a green assumption.
- [ ] A `.changeset/*.md` file with a `bump: patch` frontmatter key and Keep-a-Changelog prose accompanies the change, and neither `.claude-plugin/plugin.json` nor `CHANGELOG.md` is edited directly.

### Approach

Add `scripts/summarize-ci-checks.sh`. It resolves `gh` via `lib/resolve-gh.sh` and `jq` via `lib/resolve-jq.sh`. It calls `GET repos/$REPO/actions/runs?head_sha=$HEAD_SHA --paginate`, drops runs whose workflow name equals `$SELF_WORKFLOW_NAME`, and for each survivor calls `GET repos/$REPO/actions/runs/{id}/jobs --paginate` to collect job names and conclusions. It then calls `GET repos/$REPO/commits/$HEAD_SHA/check-runs --paginate` and keeps only entries whose `app.slug` is not `github-actions` and whose name is not `Devflow Review`. Names are sanitized, truncated, and capped. Every failure path prints `CI status unavailable` to stdout, writes a specific breadcrumb to stderr, and exits 0. It makes no `git` calls, so `lib/test/run.sh` drives it directly with a `DEVFLOW_GH` stub.

In `devflow-runner.yml`, extend the `Compose review prompt` step. It already builds a `> [!IMPORTANT]` block from `steps.provision.outputs.health_summary`; give it a second block, always present, carrying the fenced output of `scripts/summarize-ci-checks.sh` and the verbatim `steps.tools.outputs.tools` string, followed by the statement that any command outside those rules is denied. The `Resolve allowed-tools profile` step already runs before `Compose review prompt`, so `steps.tools.outputs.tools` is available. Add `checks: read` to the job's `permissions:` block and to the caller's grant in `devflow-review.yml`; these two are a coupled pair and are edited together.

In `devflow.yml`, add the equivalent compose step ahead of the `claude` job's `Run Claude Code`, gated on the resolved command being `/devflow:review`, reading the same `--allowed-tools` string it already assembles.

Add `Bash(mkdir:*)`, `Bash(tee:*)`, `Bash(git cat-file:*)`, and `Bash(git checkout:*)` to the `review` profile string. Phase 0.2 of `skills/review/SKILL.md` then works unmodified, and Phases 0.3.6 and 3.1 can execute. Grant the final-pass reviewer's helper by adding a wildcard-specifier rule for `load-prompt-extension.sh`; Claude Code's documented Bash rule syntax permits a `*` at any position in the specifier.

In `skills/review/SKILL.md`, invoke the `writing-skills` skill first and follow its RED/GREEN discipline. Add the block-conditioned instructions, the untrusted-data instruction, and the terminal-`❌` obligation. Delete the Phase 4 telemetry sentence asserting the profile lacks `Bash(tee:*)`.

In `finalize_check`, on the `success` review-result path where the derived verdict is `incomplete`, emit an `::error::` naming the HEAD SHA and the `permission_denials_count` job output, and — when the progress comment is readable — the `**Last updated:**` field, before the existing exit 1.

### Code Patterns

- `scripts/derive-review-preconditions.sh` — the authority for this change's CI queries: `SELF_WORKFLOW_NAME` exclusion, the `app.slug` / `Devflow Review` filters, `--paginate` with `jq -rs` array flattening, and the fail-closed `unverifiable` arm.
- `scripts/parse-engine-error.sh` and `scripts/derive-review-verdict.sh` — always-exit-0 helpers, stderr breadcrumbs, workflow reads stdout, unit-tested in `lib/test/run.sh` with `gh` stubbed.
- The `health_summary` conditional in `Compose review prompt` — the existing prompt-prepend shape.
- The `checkout-token:` pins in `lib/test/run.sh` (#357) — the precedent for asserting a coupled pair of workflow values from the test suite.

### Testing Strategy

**Boundary.** Every piece of this change has an automated boundary. `summarize-ci-checks.sh` is a CLI with stdout, stderr, and an exit code. The profile strings, the permission blocks, and the skill's command vocabulary are static text a test greps. The `finalize_check` annotation is shell logic. Test levels: **unit** tests for the helper and for the command-head extractor against fixtures, and **static-contract** pins in `lib/test/run.sh` for the coupled workflow, skill, and profile values. Both levels are exercised; neither substitutes for the other.

**Test-first, and what must fail first.** The profile pin is a **bug-fix regression test**: written against today's `devflow-runner.yml` and `skills/review/SKILL.md`, the command-head extractor must fail by reporting `mkdir`, `tee`, `git cat-file`, and `git checkout` as ungranted — reproducing the exact defects that silenced Phase 0.2, Phase 0.3.6, and Phase 3.1 — and pass only after all four rules are added. Run the same extractor against `devflow.yml`'s allowlist: it must report `git cat-file` and `git checkout` ungranted there too, and must report neither `mkdir` nor `tee`, which proves the extractor discriminates between the two profiles rather than always failing. The `checks: read` pin must fail today by finding the key absent from both blocks. The `::warning::`-on-denials pin must fail today because `surface-execution-diagnostics.sh` prints denials as plain stdout.

**Guarantee-class assertion.** The command-head pin and the `finalize_check` annotation are deterministic backstops, so each is tested on the path where the cooperative actor skipped its step. The pin is fed a fixture `SKILL.md` containing a command head absent from a fixture profile string and must fail RED; it is fed a compound command joined by a newline, by `|`, by `;`, and by `&&` and must evaluate **each** subcommand, matching Claude Code's documented separator set; it is fed `timeout 300 bash x.sh` and must resolve the head to `bash`. The `finalize_check` annotation is asserted on a simulated run whose review job returned `success`, whose derived verdict is `incomplete`, and whose engine never updated its progress comment — the precise PR #340 shape — and it must still emit the `::error::` with the denial count.

**Extractor false-positive assertion.** The extractor is fed a fixture containing the prose sentences that yielded `git a`, `git failure`, and `git said` during investigation, plus a `jq` program inside a `bash` fence, a heredoc body containing `rm -rf /`, a commented-out `# git diff …` line, and a fence tagged `json`. It must extract zero heads from all of them. A pin that fires on prose is worse than no pin, because it trains the maintainer to weaken it.

**Prompt-injection assertions.** `summarize-ci-checks.sh` is fed a check-run name containing a fenced-block terminator, a `> [!IMPORTANT]` line, an `Ignore previous instructions` sentence, a null byte, a newline, and 5000 characters. The rendered output must contain no unescaped fence terminator, must be truncated at 120 characters, and must not permit the injected text to escape the fence. It is fed 200 check runs and must emit at most 50 lines plus the dropped-count line.

**Self-exclusion assertions.** The helper is fed a payload containing a workflow run whose name equals `SELF_WORKFLOW_NAME`, an Actions check run whose `app.slug` is `github-actions`, and a check run named `Devflow Review` with `status: in_progress`. None appears in the output. A consumer job legitimately named `review` inside a non-self workflow **does** appear — this asserts the exclusion is keyed on workflow identity, not on the string `review`.

**Adversarial input matrix.** Per the best-effort-parser discipline, `lib/test/run.sh` asserts exit 0 plus a **specific**, non-generic breadcrumb for each of: `gh` exits non-zero on the runs query; `gh` exits non-zero on the jobs query; `gh` exits non-zero on the check-runs query; stdout is not JSON; the top level is an array; the top level is a scalar; `check_runs` is absent; `jobs` is absent; `check_runs` is an empty array; `check_runs` is an object rather than an array; a name is a number; a `conclusion` is `null` with `status: in_progress`; a `conclusion` is `failure`. The `null`-conclusion and empty-array rows are load-bearing: each is a valid-falsy shape that a `// "success"`-style extraction would silently coerce into a passing result, which is the documented-off-switch bug class (#312).

**Pagination assertion.** The stubbed `gh --paginate` returns two concatenated JSON arrays. The helper must emit signals from both pages, proving the `jq -s` flatten is present.

**Capability-renderer cases.** An empty allowed-tools string renders a block that grants nothing and still states the denial rule. A profile string whose entries contain commas inside `Bash(...)` specifiers is rendered without being split mid-rule.

**Security assertion.** `lib/test/run.sh` asserts the base `review` profile string grants neither `Bash(bash:*)` nor a rule matching `lib/test/run.sh`, so this change does not silently widen the read-only posture into arbitrary shell execution. It asserts `devflow_runner.provision_env` remains `false` in `.devflow/config.example.json`.

**Not mocked.** The `gh` binary is stubbed; the helper's own parsing, the profile strings, and the extraction logic under test are never stubbed. No test asserts on `gh`'s internal call shape.

**AC mapping.** Each acceptance criterion above maps to at least one named assertion in `lib/test/run.sh`; the injected-block criteria map to pins over the rendered compose-step output in both workflows, and the skill criteria map to greps over `skills/review/SKILL.md`.

### Documentation Needed

- `docs/cloud-setup.md` — record the `require_ci_green` precondition relationship, the `check_run[rerequested]` bypass, the path prefix under which the `review` profile grants its helpers, and the new injected block.
- `docs/workflow-triggers.md` — record that the injected block reports observed conclusions.
- `docs/DEVFLOW_SYSTEM_OVERVIEW.md` — record the injected block in the review-engine section.
- `CLAUDE.md` — via `/claude-md-management:revise-claude-md`, record the skill↔profile coupled invariant (one skill, two allowlists) and its new pin.

### Potential Gotchas

- `skills/review/SKILL.md` runs under **two** allowlists: `devflow-runner.yml`'s `review` profile and `devflow.yml`'s command allowlist. They currently disagree on `mkdir` and `tee`. Any command head added to the skill must be granted by both, which is what the pin enforces.
- A reusable workflow requesting a permission its caller did not grant aborts the run at graph-build time. `checks: read` lands in `devflow-runner.yml` and `devflow-review.yml` in the same commit; landing one alone breaks every review.
- Claude Code splits compound commands on `&&`, `||`, `;`, `|`, `|&`, `&`, and **newlines**, and every subcommand must match a rule. It also strips the process wrappers `timeout`, `time`, `nice`, `nohup`, `stdbuf`, and bare `xargs` before matching — `timeout 300 bash lib/test/run.sh` matched as `bash …` in the observed run. The extractor models both behaviors.
- Phase 0.3.6 and Phase 3.1 have never executed in a cloud review. Granting `git cat-file` and `git checkout` makes them run for the first time, so both are unexercised paths that may surface further defects once unblocked.
- `git checkout` mutates the working tree. It is granted because Phase 3.1's dirty-tree backstop exists precisely to restore agent-dirtied paths; the reviewer's `contents: read` token still makes a push impossible.
- `skills/review/SKILL.md` is the shared review engine. `/devflow:review-and-fix` executes Phases 0–4.3 verbatim and holds the `Write` tool, so the new instructions are conditioned on the injected block's presence rather than stated unconditionally.
- `skills/requesting-code-review/SKILL.md` is vendored into third-party consumer repos and its body must stay repo-agnostic — no `lib/test/run.sh`, no DevFlow-internal step numbers.
- `require_ci_green` treats `skipped` and `neutral` as green. Rendering per-signal conclusions rather than a summary boolean is what keeps a skipped test job from reading as a passing one.
- Check-run and job names originate from workflow files a pull request can add, so they are attacker-controlled strings entering a `pull_request_target` prompt. Sanitization is a security control, not cosmetics.
- This change touches the engine surface, so it requires a `.changeset/*.md` and must not edit `.claude-plugin/plugin.json` nor `CHANGELOG.md` directly.

The01Geek:
## Summary
- A cloud review run could burn its entire budget and end with **no verdict** — a red, required `Devflow Review` check carrying no findings and no cause. The engine had no ground truth about which commands it may run, nor about the CI that already ran on the commit it reviews, so it discovered both by trial and denial (on PR #340, seven of fourteen denials were the engine trying, seven ways, to run the test suite).
- Every cloud review prompt now begins with an engine-ground-truth block stating the CI conclusions observed for the reviewed HEAD and the exact `--allowed-tools` string the run resolved. The `review` profile grants the commands the engine actually invokes, so two review phases that had never executed in cloud now run.
- A run that reaches no verdict announces itself twice, independently: the engine stamps a terminal ❌ on its progress comment, and `finalize_check` emits an `::error::` naming the head and the denial count. The second survives an engine that never runs the first.

## Changes

**Injected ground-truth block**: `scripts/render-grounding-block.sh` renders a `> [!IMPORTANT]` block prepended to the prompt on every cloud review — the automated path (`devflow-runner.yml`'s `Compose review prompt`) and the manual `/devflow:review` comment path (`devflow.yml`'s new `Compose review grounding block`, gated so `/devflow:review-and-fix` is unaffected). It lives in **one** file because it carries prompt-injection-defense prose; both workflows are pinned to invoke it and to carry no second copy. `devflow.yml`'s inline `--allowed-tools` list is hoisted into a step output consumed by both `claude_args` and the block, so the block quotes the exact resolved string by construction rather than a copy that could drift.

**CI summarizer**: `scripts/summarize-ci-checks.sh` renders one line per CI signal for the reviewed head — paginated `actions/runs?head_sha=` plus each run's `/jobs`, and `commits/{sha}/check-runs` — excluding DevFlow's own workflow, the duplicate `github-actions` app check-runs, and its own `Devflow Review` check. It always exits 0. A `null` conclusion renders its `status` (never coerced to `success`), a genuinely empty head renders `No CI signals reported for this commit`, and every failure renders the literal `CI status unavailable` with a query-specific breadcrumb — an unknown CI state is never rendered as a passing one. Check names are attacker-controlled text entering a `pull_request_target` prompt, so they are sanitized, truncated to 120 characters, and capped at 50 lines plus a dropped-count line.

**Tool profile and permissions**: the `review` profile grants `mkdir`, `tee`, `git cat-file`, `git checkout`, `mktemp`, `cmp`, `rm -f`, and `load-prompt-extension.sh` from any anchor directory. Phase 0.2's cached diff is now created (so Phase 3's `{DIFF_PATH}` resolves), and Phase 0.3.6's blocker-recheck fast path and Phase 3.1's dirty-tree backstop execute in cloud for the first time. `checks: read` lands in `devflow-runner.yml` and its caller job in `devflow-review.yml` as a coupled pair — a reusable workflow requesting a permission its caller did not grant aborts the run at graph-build time — plus `devflow.yml`'s `command` job. The false comment claiming `--permission-mode acceptEdits` widens an explicit `--allowed-tools` list is corrected.

**Observability**: `scripts/surface-execution-diagnostics.sh` raises a `::warning::` when a run records permission denials, and publishes `permission_denials_count` as a job output. Unknown is **not** zero: when the count cannot be established it publishes the literal `unavailable`, so `finalize_check`'s no-verdict `::error::` can never assert "the harness refused 0 command(s)" about a run it never measured. The three-way clause selection lives in `scripts/describe-denial-count.sh` so every arm is unit-tested rather than an untestable inline `case`.

**Review engine skill**: `skills/review/SKILL.md` gains an "Engine ground truth" section, conditioned on the block's presence so `/devflow:review-and-fix` (same phases, no block) is unaffected. It tells the engine the block's CI signals are the authoritative test evidence, that no command outside the injected list is to be attempted, and that every check **name** is untrusted data — while the **conclusions** beside them are API facts, never grounds to doubt the evidence. It also requires a terminal ❌ stamp on any path that reaches no verdict. The stale claim that the profile lacks `Bash(tee:*)` is removed.

**Drift pin**: `lib/test/extract-command-heads.py` extracts every Bash command head from the ```bash fences of `skills/review/SKILL.md` — quote-, comment- and heredoc-aware, descending into `$(…)`, splitting on Claude Code's separator set and stripping its process wrappers — and `lib/test/run.sh` asserts **both** allowlists grant each one. A command the skill invokes but a profile omits now turns the suite RED instead of being silently denied at runtime.

## Resolves
Resolves #363

## Test Plan
- [ ] `lib/test/run.sh` passes (4164 assertions, up from 3988 on `main`).
- [ ] Remove `Bash(mkdir:*)` from `devflow-runner.yml`'s `TOOLS='…'` line and confirm the suite goes RED naming `mkdir` as ungranted; restore it.
- [ ] Run `scripts/summarize-ci-checks.sh` with `sed` and `wc` removed from `PATH` and confirm no selection changes (the counts are derived with bash builtins).
- [ ] Feed `scripts/surface-execution-diagnostics.sh` an execution file recording 14 denials under a `PATH` without `sed`; confirm it publishes `permission_denials_count=14` and emits the `::warning::`.
- [ ] Feed it an unparseable execution file; confirm it publishes `unavailable` (never `0`) and emits no warning.
- [ ] `bash scripts/describe-denial-count.sh 0`, `… 14`, and `… unavailable` each print a distinct clause.
- [ ] `shellcheck --severity=warning -e SC1091` is clean on every changed `.sh`; `ruff check` is clean (CI runs it).
- [ ] After merge, watch one cloud `Devflow Review` run: its prompt-composition step resolves a non-empty grounding block, and the run reaches a verdict.

## Visual Changes
N/A

## Breaking Changes
None. `permission_denials_count` is a new job output; consumers that read it must treat the literal `unavailable` as unknown rather than coercing it to `0` (the reusable-workflow output description states this). The `checks: read` grant is additive and lands in the runner and its caller in the same change, so no review aborts at graph-build time.


The01Geek:
**Deferral record (receiving-code-review pass on the 👎 REJECT review):**

Addressed in the follow-up commit: the Critical self-contradicting comment (`devflow-review.yml` `PERMISSION_DENIALS_COUNT` env comment now describes `|| 'unavailable'` and the unknown≠0 distinction), the Important `Bash(git cat-file:*)` gap in `devflow.yml`'s hoisted TOOLS (granted + pinned against `$DEVFLOW_YML` in `lib/test/run.sh`, alongside `git checkout` — the same prose-invoked class), and both doc suggestions (untrusted-fence wording in `docs/DEVFLOW_SYSTEM_OVERVIEW.md` + `docs/cloud-setup.md` now describes the actual mechanism — a plain ```text fence beneath prose declaring the names untrusted; the two `###` sections that split the `1.`/`2.` list in `docs/cloud-setup.md` moved below the list).

**Deferred — test-coverage suggestions (pr-test-analyzer, non-blocking):**
- WHAT: (a) `finalize_check`'s `LAST_PHASE` jq is tested via a hand-copied mirror in `run.sh` rather than extracted from the YAML; (b) no dedicated fixture for `extract-command-heads.py`'s `_strip_case_patterns`; (c) untested arms in `summarize-ci-checks.sh`: `(unnamed check)` fail-closed name fallback, empty-`REPO`/`HEAD_SHA` entry guards, both-null `"unknown"` fallback.
- WHY deferred: advisory coverage hardening on paths that fail closed; out of scope for clearing the carve-out REJECT, and each needs its own RED/GREEN test work rather than a drive-by edit.
- REVISIT when: any of those code paths is next edited, or if a coupled-mirror drift between the `LAST_PHASE` jq in `devflow-review.yml` and its `run.sh` mirror is observed.

Issue #402, opened 2026-07-10
Close the deny-floor gap: parameterized tree-mutation entries (Write(**), Edit(...)) bypass the reviewer allowlist filter


The01Geek:
## Problem Statement

The consume-time deny-list floor that protects the cloud reviewer's allowlist strips the tree-mutation tools `Edit`/`Write`/`MultiEdit`/`NotebookEdit` by **exact string equality** against each `devflow_runner.allowed_tools` entry. A parameterized form — `Write(**)`, `Edit(src/**)`, `NotebookEdit(x)` — is not string-equal to the bare name and does not start with `Bash(`, so it bypasses the floor entirely and is appended to the reviewer's profile. The floor's own comment and `docs/DEVFLOW_SYSTEM_OVERVIEW.md` state the intent this violates: tree-mutation tools are *never* permitted on the reviewer's write-token job, regardless of what config says. The affected party is any consumer repo with `devflow_runner.provision_env: true` whose `allowed_tools` was LLM-enriched by `/devflow:init` or later hand-edited — exactly the actors the floor exists to defend against. (Found during the issue #401 design review; issue #401 deliberately places its scoped `Write` grants in the trusted workflow literal, not config, so this gap is orthogonal to that fix.)

## Current Behavior

In `.github/workflows/devflow-runner.yml` (review profile, `PROVISION_ENV` append path), the `DENY_NAMES='Edit Write MultiEdit NotebookEdit'` loop tests `[ "$entry" = "$n" ]`. `Write(**)` fails that equality test, falls through the `Bash\(*` case arm untouched, and lands in `FILTERED` — granted to the reviewer. The fast-feedback mirror (`denylisted` in `scripts/detect-project-tools.sh`) has the same blind spot for parameterized file-tool names. The filter logic is inline workflow YAML, so no suite test can exercise it.

## Desired Behavior

[...]

Issue #401, opened 2026-07-10
Eliminate cloud-review no-verdict stalls from permission-denial burn (matcher probe + targeted grants + command-shape discipline)


The01Geek:
## Problem Statement

The cloud review tier's required `Devflow Review` check fails nondeterministically with "Devflow review incomplete — re-run needed" when the review engine burns its run on commands the read-only tool profile silently denies. The maintainer merging DevFlow PRs is blocked on a coin-flip Re-run; each stalled run wastes roughly $10 and 10 minutes and yields no verdict.

Evidence of record — review run [29105381021](https://github.com/The01Geek/devflow-autopilot/actions/runs/29105381021) on PR #397: 22 permission denials, engine quit cleanly mid-Phase-3 (2 of 5 agents never dispatched), progress comment frozen at "🚀 Reviewing", `finalize_check` correctly failed closed. The prior *approving* run on the same PR recorded 13 denials — the friction is chronic and occasionally fatal. Precedent: PR #340 (7 of 14 denials, led to issue #363's head-extraction guard).

## Current Behavior

- The read-only `review` profile (`TOOLS` in `.github/workflows/devflow-runner.yml`) grants command *heads*; the permission matcher denies composite *shapes*. Observed denied in run 29105381021: `cd`-led compounds (~8), leading `VAR=` assignment compounds (~6), `python3` invocations (3), `Write`-tool calls (2), `cat > /tmp/… <<'EOF'` heredoc writes with the granted `cat` head (2), and one unexpanded `"${CLAUDE_SKILL_DIR:-…}"/../../scripts/load-prompt-extension.sh` anchor invocation.
- `skills/review/SKILL.md`'s own Phase 0.3.5 fence prescribes the assignment-led `printf`/`cat`-heredoc composite, and its inline comment claims `Bash(cat:*)`/`Bash(printf:*)` cover it — contradicted by the observed denials.
- `lib/test/extract-command-heads.py` (issue #363) validates that both allowlists grant every fence's command *head*, so it stays green on all of the above.
- The injected engine-ground-truth block (`scripts/render-grounding-block.sh`) lists the granted rules but says nothing about accepted command shapes, so the engine iterates denied variants (~6 formulations of one literal-presence check in that run) until it gives up.
[...]

PR #407, opened 2026-07-10
Record the cloud engine's execution transcript as a scrubbed artifact (issue 401 follow-up)

The01Geek:
Part of the #401 evidence trail — not a behavior change to the review itself.

**Why:** three no-verdict review stalls (runs 29105381021, 29118545733, 29126681079) each ended on a *voluntary* engine final message (`subtype: success`, mid-phase). That message — the engine's own stated reason for stopping — has never been readable: the execution file dies with the runner. Every diagnosis so far has been inference from denial counts.

**What:** the `Surface execution diagnostics` step exports its config gates — and the transcript upload sits behind its OWN opt-in key `devflow.execution_transcript_artifact_enabled` (default FALSE, fail-closed; this repo opts in via its tracked config.json, consumers are unaffected unless they opt in); a new fail-closed step scrubs token-shaped strings (GH tokens/PATs, Anthropic keys, Authorization headers — `Bash(env:*)` is granted, so a transcript can contain a dumped environment) and uploads the result with 7-day retention.

**Next use:** after merge, re-run PR #404's stalled review — if it stalls again, the artifact tells us exactly why; if it completes, #402 unblocks.

Suite: 4788 passed, 0 failed. actionlint: only the pre-existing info-level SC2086 in the untouched deny-floor block.


The01Geek:
## Deferred review findings — run 29127849497 (APPROVE with notes)

Verdict was non-blocking (17/17 checklist pass, no Critical, no documented-falsehood carve-out), so per the configured `critical` re-open threshold the following notes are **recorded as deferrals**, to be consolidated into a follow-up issue:

1. **[Important] No `lib/test/run.sh` pin for `execution_transcript_artifact_enabled`** (schema default-false, example-matches-default, the `[ "$TRANSCRIPT" = "true" ] || TRANSCRIPT=false` fail-closed clamp, workflow wiring — mirroring the `execution_diagnostics_enabled` pin family). *Why deferred:* test-coverage gap, not a live defect — the shipped gate was empirically verified in-review (VC-13). *Revisit:* the follow-up issue; must land before any edit touches the transcript gate.
2. **[Important] Scrub success path emits no breadcrumb and the blocklist is silently incomplete** — add a best-effort `::warning::` + a caveat header prepended into the scrubbed output naming the four patterns covered. *Why deferred:* limitation is disclosed in-source; default-false opt-in bounds exposure. *Revisit:* follow-up issue.
3. **[Suggestion] No `[ -s "$OUT" ]` check before advertising `path=`** — an empty scrub output uploads an empty artifact, masking transcript-preservation failure. *Revisit:* with item 2 (same step).
4. **[Suggestion] Base64 basic-auth `Authorization` form not redacted** — low reachability under the read-only review profile; defense-in-depth. *Revisit:* with item 2.
5. **[Suggestion] Schema `description` hardcodes "7-day"**, coupled to `retention-days: 7` — comment-rot hazard. *Revisit:* whenever retention is tuned; treat as a coupled pair.
6. **[Suggestion] Trailing newline dropped on the three config files** — cosmetic churn. *Revisit:* opportunistically on the next touch of those files.

Issue #357, opened 2026-07-09
Cloud writers push as github-actions[bot], not the App: checkout@v6's includeIf'd auth header outranks claude-code-action's remote-URL token


The01Geek:
## Summary

Every cloud-tier `/devflow:implement` and `/devflow:review-and-fix` run pushes as
`github-actions[bot]`, **not** as the configured `devflow-autopilot` App — so any push
touching `.github/workflows/` is rejected with:

```
refusing to allow a GitHub App to create or update workflow
.github/workflows/<file>.yml without workflows permission
```

Ordinary pushes succeed (`GITHUB_TOKEN` has `contents: write`), which is why this hid for
so long and why it was repeatedly misattributed to the App's permissions. The App is fine:
`devflow-autopilot` (app_id `3102164`) declares `workflows: write`, and a manual push with
its credentials works.

## Root cause

The credential `actions/checkout` persists — **not** the `github_token` handed to
`claude-code-action` — is what `git push` authenticates with.

`actions/checkout@v6` no longer writes its auth header into `.git/config`. It writes
`http.https://github.com/.extraheader` into an external config file under `RUNNER_TEMP` and
wires it in through `includeIf.gitdir:` directives (including a `.git/worktrees/*` glob).

`claude-code-action`'s `configureGitAuth()`
([`src/github/operations/git-config.ts`](https://github.com/anthropics/claude-code-action/blob/main/src/github/operations/git-config.ts))
tries to clear that header with `git config --unset-all http.<server>/.extraheader`, which
searches only `.git/config`. It matches nothing. It then rewrites `origin` to
`https://x-access-token:<app-token>@github.com/…` — but the surviving preemptive
`Authorization:` header outranks URL userinfo, so the App token in the remote URL is never
used.

Evidence from run `28987244791` (`claude` job):

```
01:19:12  git config --file /home/runner/work/_temp/git-credentials-….config \
              http.https://github.com/.extraheader AUTHORIZATION: basic ***
01:19:12  git config --local includeIf.gitdir:/home/runner/work/…/.git.path            …/git-credentials-….config
01:19:12  git config --local includeIf.gitdir:/home/runner/work/…/.git/worktrees/*.path …/git-credentials-….config
01:19:19  Removing existing git authentication headers...
01:19:19  No existing authentication headers to remove      <-- the unset matched nothing
01:19:19  ✓ Updated remote URL with authentication token     <-- App token embedded, but outranked
[...]

Issue #287, opened 2026-07-03
Harden the stall-backstop auto-resume against App-token expiry + lower this repo's implement effort


The01Geek:
## Problem Statement

The `/devflow:implement` run on #284 stalled twice at `🚀 Reviewing` and never finished. The auto-resume backstop fired once but then failed to fire again, and the run's git commits were attributed to `github-actions[bot]` rather than the `devflow-autopilot` App. Investigation of runs `28665669190` (initial, 39 min) and `28667699880` (auto-resume, 68 min) traced the failed second resume to a single root cause: **the stall-backstop step reuses the App installation token the `claude` job minted ~68 minutes earlier, and GitHub App installation tokens expire after 60 minutes.** By the time the backstop ran, the token was dead — `scripts/workpad.py status` (authenticated with the expired token) got `HTTP 401: Bad credentials`, a perfectly healthy workpad was misclassified as `unreadable`, the decision resolved to `fail-unreadable` (terminal, no resume), and the resume-comment POST also 401'd.

This is a structural defect, not bad luck: **the condition that triggers the need for a resume (a run long enough to stall) is the same condition that expires the token the resume depends on.** Any run that stalls after 60 minutes can never auto-resume. Compounding it, this repo runs `/devflow:implement` at `medium` effort, making each run slower and more expensive than necessary — which makes exceeding the 60-minute window more likely in the first place.

(Note: the `github-actions[bot]` commit attribution is a separate, cosmetic `claude-code-action` behavior — it commits as `claude[bot]` under the github-actions bot ID and uses the App token only for push auth — and is explicitly **out of scope** for this issue.)

## Current Behavior

- `.github/workflows/devflow-implement.yml` mints the optional App token once, at the start of the `claude` job (`id: app-token`, gated on `vars.DEVFLOW_APP_ID`). The `Stall backstop` step, which runs after a claude step that can last well over an hour, consumes that same now-stale token via `GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}`.
- `scripts/workpad.py status` returns exit 1 for **both** a `gh` api / transport / auth failure **and** a genuinely unparseable/absent Status line (see its docstring, lines ~254–261, and `_fail('repo lookup', …)`). The stall backstop maps exit 1 and exit 2 alike to the `unreadable` decision class → `fail-unreadable` (terminal, consumes the retry budget for the issue).
[...]

Issue #434, opened 2026-07-12
Scope the stale-prose lint to comment/prose lines (claim AND R4 permit referent)


The01Geek:
# Scope the stale-prose lint to comment/prose lines (claim AND R4 permit referent)

## Problem Statement

The deterministic stale counted-prose lint (issue #423, shipped by PR #424) **grades data that merely looks like a claim**. Run the helper over PR #424's own branch diff and it reports **14 STALE rows (measured at `73bb9a0`), every one of them in `lib/test/run.sh`** — the lint's own test fixtures, whose added lines *are* stale claims by construction:

```
printf '%s\n' '# Cases 19-32 are exercised by the block below' 'Case 19 alpha' 'Case 37 delta' > "$SPF"
assert_eq "#423 T2/R2 reconciled sibling (7 bullets) exits 0" "0" "$(spl_rc "$SPR")"
```

This is a **scope violation, not a missing feature**. Issue #423's own acceptance criterion says the rules detect claims "over **diff-added** comment/prose lines". The code does no such thing: `examine_file` iterates **every** diff-added line of **every** path — shell code, YAML, JSON, all of it. There is no comment gate and no file-type gate. (PR #424 corrected the module header to state the real, wider scope rather than ship a documented falsehood; that records the limitation, it does not fix it.) The lines above are *shell code*, not comments — under #423's contract they would never have been examined. **This change restores the contract #423 specified.**

Once the helper reaches the vendored path, any PR touching `run.sh`'s `#423` block draws a spurious Important finding per row, each entering Phase 4.2 verdict computation. That is wasted fix-loop iterations and possible REJECT rounds — precisely the churn #423 exists to eliminate. The feature currently reproduces the disease it was built to cure.

**Path-exclusion is the wrong remedy, on evidence.** Excluding `lib/test/**` would blind the lint to its own origin story: `run.sh` carries the *real* counted header `# Cases 19-37 (issue #327)` — literally PR #328's **healed** header, the defect that motivated this entire feature. Moving fixtures into data files fails identically: the claims relocate, they do not stop being scanned. The problem is **per-line**.

## Current Behavior

- `examine_file` examines every diff-added line of every path. The regexes are prose-*shaped*, so most code lines happen not to match — but fixture strings, assertion names, and any shell string carrying a counted claim do match, and are graded exactly like a real header.
[...]

PR #491, opened 2026-07-14
Keep writer-job push and gh credentials fresh past the App token's 60-minute lifetime

### LINKED ISSUE #487: Keep writer-job push and gh credentials fresh past the App token's 60-minute lifetime
## Problem Statement

GitHub App installation tokens expire exactly one hour after minting and cannot be renewed — only replaced by a fresh mint (official docs, recorded below). DevFlow's two writer jobs mint one App token at job start and ride it for the whole run. Any `claude` step that outlives that hour spends its remainder with dead credentials: `git push` and agent-side `gh` calls fail with `Bad credentials`, and the agent — with no recognition of the state — keeps iterating on the failures. Evidence: run 29299441781 ("DevFlow implement (issue 469)") ran its `claude` job for 75.7 minutes (01:47:52 → 03:03:36) and burned roughly $60 of agent budget fighting a dead `git push` credential it was unable to diagnose and unable to repair. Long runs are DevFlow's normal case — real tickets in real codebases — so this is a standing tax on exactly the runs DevFlow exists to ship.

## Current Behavior

- `.github/workflows/devflow-implement.yml` (`claude` job) and `.github/workflows/devflow.yml` (`command` job, non-review commands) each mint one installation token at job start (`actions/create-github-app-token@v3`), seed it into `actions/checkout`'s `token:` — persisting it as the `http.<server>/.extraheader` credential every in-run `git push` authenticates with (the #357 contract) — and pass the same token to `claude-code-action`'s `github_token:`.
- Past minute ~60 of that step, three credential surfaces go dead at once:
  1. **`git push`** — the persisted extraheader still carries the expired token (the observed $60 fight);
  2. **agent-side `gh`/REST calls** — `scripts/workpad.py`, `scripts/apply-labels.sh`, skill-fence `gh pr create` — the agent-visible token is the same expired job-start mint;
  3. **`claude-code-action`'s own API calls** (progress-comment updates) — the same static `github_token` input; tracked upstream as anthropics/claude-code-action#716 (open, no fix shipped).
- Nothing in the run detects the state; the agent retries push variants and burns budget until the run dies.
- The **post-run** instance of this bug is already fixed: the stall backstop mints its own fresh token just-in-time (issue #287), as does devflow.yml's review backstop. Only the **in-run** window is uncovered.

## Desired Behavior

Writer-job runs stay push- and API-capable for their entire duration. A background credential refresher — a `background: true` step started after checkout and before the `claude` step in both writer jobs — holds the App credentials, re-mints a fresh installation token on a 45-minute cadence (with a 2-minute backoff retry after any failed cycle), and rewrites the two repo-controlled credential surfaces in place: the checkout-persisted extraheader (so `git push` always authenticates with a token younger than 45 minutes) and a mode-0600 token file that agent-side `gh` reads at call time through an installed wrapper. When a credential is nonetheless rejected at run time, the run fails fast with a named cause instead of fighting it. With no App configured, everything behaves exactly as today.

## User Impact

Every `/devflow:implement` and `/devflow:review-and-fix` run longer than an hour — in this repo and in consumer repos — completes its pushes, workpad updates, labels, and PR creation instead of dying mid-lifecycle. The ~$60-per-incident budget burn from credential-fighting disappears, and the stall backstop stops being the only part of a long run that survives the hour mark.

## Technical Context

**Verified third-party facts (docs ladder, sources recorded):**

- Installation access tokens are minted by `POST /app/installations/{installation_id}/access_tokens` with an RS256 app JWT (installation id from `GET /repos/{owner}/{repo}/installation`) and "will expire after 1 hour", with no renewal mechanism — [GitHub docs: generating an installation access token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app).
- `actions/create-github-app-token`'s README states the 1-hour expiry, and the maintainer-recommended long-running pattern is to hold the App credentials in the long-lived process and mint installation tokens on demand — [README](https://github.com/actions/create-github-app-token) and maintainer guidance in actions/create-github-app-token#121 (comment 2043214796).
- The same failure is open upstream against our exact stack: anthropics/claude-code-action#716 ("The fixed timeout of github_token from Github APP is blocking long-running jobs") — no fix shipped, so surface 3 in Current Behavior is upstream-only and is disclosed here rather than fixed.
- `actions/checkout` persists the credential in a separate config file `git-credentials-<UUID>.config` under `$RUNNER_TEMP`, wired in via `includeIf.gitdir:`; the key is `http.<serverUrl>/.extraheader` and the value is `AUTHORIZATION: basic <base64 of x-access-token:TOKEN>` — verified against the action source ([src/git-auth-helper.ts](https://github.com/actions/checkout/blob/main/src/git-auth-helper.ts): `tokenConfigKey`, `x-access-token:${...authToken}`), agreeing with the in-repo #357 gotcha and `docs/cloud-setup.md`. The refresher derives the key's server URL from `GITHUB_SERVER_URL` and locates the persisted file at run time via `git config --show-origin --get-all http.<serverUrl>/.extraheader` rather than hardcoding a path.
- GitHub Actions steps can run as background services: `background: true` (with `wait`/`cancel` semantics) is generally available since June 25, 2026 — [GitHub changelog](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/).
- `claude-code-action` exports its `github_token` input as **both** `GITHUB_TOKEN` and `GH_TOKEN` into its own process env, inherited by every agent subprocess — verified against the action source (`src/entrypoints/run.ts`: "Set GITHUB_TOKEN and GH_TOKEN in process env for downstream usage"). This is why the wrapper must fingerprint-discriminate the ambient job-start token rather than defer to any preset `GH_TOKEN`.
- Community precedent for mint-on-demand at the git layer (credential helpers with ~55-minute token caches): AmadeusITGroup/gh-app-auth, bdellegrazie/git-credential-github-app, mackee/git-credential-github-apps. A stock credential helper is **not** usable here: the persisted extraheader — and the token `claude-code-action` embeds in the origin URL — both outrank credential helpers in git's precedence order (the exact fight #357 documents). The adapted mechanism therefore rewrites the extraheader itself, keeping it the credential-of-record.

- **Relevant Classes/Files** — `.github/workflows/devflow-implement.yml` (`claude` job: the `app-token` mint step, the seeded checkout, the claude step env), `.github/workflows/devflow.yml` (`command` job: same pattern; its mint `if:` additionally excludes `/devflow:review` commands), new `scripts/refresh-app-credentials.sh`, new `scripts/gh-fresh.sh` (the gh wrapper, a tracked repo file so shellcheck and the suite cover it), `skills/implement/SKILL.md` and `skills/review-and-fix/SKILL.md` (fail-fast rule), `lib/resolve-gh.sh` (the `DEVFLOW_GH` seam the wrapper plugs into), `lib/test/run.sh` (new arms; existing checkout-token pins), `docs/cloud-setup.md`.
- **Architecture Alignment** — mirrors the #287 precedent (a fresh mint beats stale reuse) and keeps the #357 seeded-extraheader contract untouched: the refresher rewrites the same credential-of-record that checkout created; git credential precedence is unchanged. The wrapper rides the existing `DEVFLOW_GH` override contract (honored verbatim, never probed — `lib/resolve-bin.sh`), and setting `GH_TOKEN` in a spawning environment is the pattern every existing workflow step already uses for `gh` auth.
- **Dependencies** — `openssl` and `curl` on the ubuntu GitHub-hosted runner for the JWT mint — verified: both are in the [actions/runner-images Ubuntu 24.04 manifest](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) (`openssl 3.0.13`, `curl 8.5.0`). No new Actions dependencies, no new config keys (`vars.DEVFLOW_APP_ID` / `secrets.DEVFLOW_APP_PRIVATE_KEY` reused as-is), and **zero new agent-facing tool grants** — the refresher and the wrapper install run in trusted workflow steps outside the agent allowlists, so there is no #363/#401 skew to manage.
- **Data/Schema Considerations** — none.
- **Cross-layer Impact** — workflows (two writer jobs), scripts (refresher + wrapper), skills prose (fail-fast rule), test suite, docs.

## Acceptance Criteria

- [ ] **Push freshness:** during a writer-job agent step that runs longer than 60 minutes, `git push` to the feature branch succeeds throughout the run — the refresher rewrites the checkout-persisted `http.<server>/.extraheader` credential with a freshly minted installation token every 45 minutes (first rewrite 45 minutes in), so the credential-of-record is never older than 45 minutes.
- [ ] **Agent API freshness:** agent-side `gh` invocations resolve their token at call time — the trusted step installs `scripts/gh-fresh.sh` as a `gh` wrapper that runs the real `gh` (absolute path resolved at install) with `GH_TOKEN` read from the refresher-maintained token file on each invocation, wired both as `DEVFLOW_GH` in the claude step's env and ahead of the real `gh` on `PATH` (via `GITHUB_PATH`) so helper calls and direct skill-fence calls resolve to it alike.
- [ ] **Ambient-vs-chosen credential discrimination (protects both the agent-side fix and the #287 backstops):** `claude-code-action` exports the job-start token as both `GITHUB_TOKEN` and `GH_TOKEN` into its process env, which every agent subprocess inherits — so a defer-to-any-preset-`GH_TOKEN` rule would leave every agent call on the expiring ambient token. The wrapper therefore discriminates by fingerprint: the install step records a hash of the job-start token, and at call time the wrapper **substitutes** the token-file credential when env `GH_TOKEN` is absent and when its hash matches the recorded job-start fingerprint (the ambient agent-session case — and devflow.yml's #356 flip step, whose `GH_TOKEN` is that same job-start mint, repairing its silent stale-flip on >60-minute dead runs), and **defers untouched** when the hash differs (the deliberately fresh #287 mints: devflow-implement.yml's stall backstop and devflow.yml's review backstop). The observability-persist step invokes no `gh` and pushes through the extraheader the refresher already maintains, so it needs no defer protection. All three post-claude surfaces are verified in the same change.
- [ ] **Key hygiene:** the App private key is delivered to the refresher via stdin; it is never exported into any environment variable visible to the agent session, never passed as a process argument, and never written to disk.
- [ ] **Unconfigured no-op:** with `vars.DEVFLOW_APP_ID` empty, the refresher step and the wrapper install are skipped by the same `if:` condition shape as each workflow's existing writer mint (devflow.yml's additionally excluding `/devflow:review` commands), and workflow behavior is byte-identical to today's `GITHUB_TOKEN` fallback path.
- [ ] **Loud degrade, never fatal:** a failed mint-and-rewrite cycle leaves the previous credential in place, emits a `::warning::` line naming which arm failed, and retries on a short backoff (every 2 minutes until a cycle succeeds, then back to the 45-minute cadence) — so a single transient mint failure never produces a dead-credential window; sustained failure is required. The refresher loop itself never exits non-zero: every failure is warn-and-continue, and it traps `TERM` and exits 0, so the job's conclusion never rides on background-step failure semantics. A probe run recorded on the shipping PR demonstrates that a killed background refresher step leaves the job conclusion green (the matcher-probe precedent: probe an action semantic before relying on it).
- [ ] **Fail-fast guard (two layers, one prose + one deterministic):** the always-resident cross-phase rules of `skills/implement/SKILL.md`, and `skills/review-and-fix/SKILL.md`, gain the rule: after two consecutive `git push`/`gh` failures whose output carries the bad-credential signature (HTTP 401 / `Bad credentials` / `Authentication failed`), the run stops retrying, records the cause in the workpad, and ends at a terminal status naming the expired-credential cause — mirroring the existing two-denials command-shape discipline, never open-ended retry iteration. The prose rule is best-effort under context compaction (a >60-minute run is the maximally compaction-likely population), so the deterministic sibling signal rides in the wrapper: when the underlying `gh` call fails with the bad-credential signature, the wrapper appends one distinctive diagnostic line to stderr naming the expired-credential cause — a compaction-immune, re-derivable signal at every failing call (`git push` failures remain covered by the prose rule alone).
- [ ] **Suite coverage:** the refresher logic lives in `scripts/refresh-app-credentials.sh` (single-cycle logic invocable without sleeping, so the suite can drive it); `lib/test/run.sh` drives exactly these nine arms — complete by construction: (1) missing inputs → clean exit 0 with a stderr breadcrumb, (2) mint success → fixture extraheader config rewritten and mode-0600 token file written, (3) mint failure → previous credential intact and a `::warning::` emitted, (4) a failed-then-succeeding cycle sequence → the backoff retry recovers and the fixture credential ends fresh, (5) wrapper substitutes when env `GH_TOKEN` is absent, reading the token file at call time (two calls observe two different tokens), (6) wrapper substitutes when env `GH_TOKEN`'s hash matches the recorded job-start fingerprint, (7) wrapper defers untouched when env `GH_TOKEN`'s hash differs from the fingerprint, (8) wrapper degrades to plain invocation when the token file is absent, (9) wrapper appends its expired-credential diagnostic line on a bad-credential failure on both the substitute and defer paths. The mint invocation honors a `DEVFLOW_`-prefixed env override that wins verbatim and is never probed (the `lib/resolve-bin.sh` `DEVFLOW_<TOOL>` stub contract), so the suite drives every arm with no network.
- [ ] **Coupled sites in one change:** both writer jobs gain the step in the same PR, and the existing checkout-token pins in `lib/test/run.sh` (#357) are reconciled in the same change.
- [ ] **Docs:** `docs/cloud-setup.md` documents the refresher (cadence, key handling, degrade behavior) and the disclosed residual — `claude-code-action`'s internal API calls still ride the static `github_token` input until anthropics/claude-code-action#716 lands upstream.
- [ ] **Upstream engagement:** the shipping PR body links anthropics/claude-code-action#716 and cites run 29299441781 as the evidence run; posting the comment on that upstream issue is a named human follow-up recorded in the PR body (a cloud implement run's repo-scoped token cannot comment on a foreign repo, so the posting itself is out of the run's reach).

## Implementation Notes

**Approach** — one decided design:

1. **`scripts/refresh-app-credentials.sh`** (new): reads the PEM private key from stdin into shell memory; exposes a single-cycle entry point (mint once, rewrite once) that a thin loop drives on a 45-minute cadence — dropping to a 2-minute backoff after a failed cycle until one succeeds — writing its PID to a pidfile at startup, trapping `TERM` to exit 0, and never exiting non-zero (every failure is warn-and-continue). A cycle: build the RS256 app JWT (`openssl dgst -sha256 -sign` with the key supplied over a file descriptor, never a disk path); resolve the installation id via `GET /repos/{owner}/{repo}/installation` (JWT bearer, `curl`); mint via `POST /app/installations/{installation_id}/access_tokens`; parse the response with jq routed through sourced `lib/resolve-jq.sh` (the repo's new-jq-caller pin); derive the server URL from `GITHUB_SERVER_URL` and locate the persisted credential file via `git config --show-origin --get-all http.<serverUrl>/.extraheader`; rewrite the header value (`AUTHORIZATION: basic <base64 of x-access-token:TOKEN>`) with `git config --file <located file>` (git's own locking makes the rewrite atomic against a concurrent push); write the raw token to a mode-0600 token file under `RUNNER_TEMP` for the wrapper. Every failure path emits a `::warning::` naming the arm and keeps the previous credential.
2. **Workflow steps** (both writer jobs, same PR): a `background: true` step placed after checkout and before the claude step, gated by the same `if:` shape as each job's writer mint, feeding the key via stdin from `secrets.DEVFLOW_APP_PRIVATE_KEY`; a sibling foreground step installs `scripts/gh-fresh.sh` as the `gh` wrapper into a fresh directory (resolving the real `gh` by absolute path first) and prepends that directory via `GITHUB_PATH`; the claude step's `env:` gains `DEVFLOW_GH` pointing at the wrapper; and a final `if: always()` step best-effort kills the refresher via its pidfile — so job completion never depends on background-step auto-cancel semantics, which the workflow-syntax docs do not yet specify.
3. **Fail-fast rule** (prose): the two-strikes bad-credential rule added to `skills/implement/SKILL.md`'s always-resident cross-phase rules and to `skills/review-and-fix/SKILL.md` (via the `writing-skills` discipline, per repo convention).
4. **Upstream linkage:** the shipping PR body links anthropics/claude-code-action#716, cites run 29299441781, and names the upstream comment as a human follow-up.
5. **`scripts/gh-fresh.sh` semantics:** defer only when env `GH_TOKEN` is set **and** its hash differs from the recorded job-start fingerprint (run the real `gh` untouched — the deliberately fresh backstop mints); in every other case (env token absent, and the hash-match ambient case) run the real `gh` with `GH_TOKEN` from the token file; pass output through on both paths, and on a bad-credential failure — whichever path was taken — append the one-line expired-credential diagnostic to stderr, preserving the real `gh` exit code. The install step records the job-start token's hash (a mode-0600 fingerprint file beside the token file) from `steps.app-token.outputs.token`.

**Code Patterns** — `scripts/ensure-label.sh` / `scripts/apply-labels.sh` for the best-effort, always-exit-0, stderr-breadcrumb contract; the #287 stall-backstop mint for the fresh-mint-over-stale-reuse precedent; `lib/resolve-bin.sh` for the `DEVFLOW_GH`-honored-verbatim contract the wrapper plugs into; existing `env: GH_TOKEN:` workflow steps as the in-repo proof that `gh` honors a spawning environment's token.

**Testing Strategy** — the change's automatable boundary is the refresher helper's observable outputs (files it writes, warnings it emits, exit codes) plus the wrapper's invocation behavior; both are unit-drivable from `lib/test/run.sh` (gh-stubbed, no network), and the skills-prose rule is pinned at the surface-presence level. Test-first: the nine suite arms in the AC above are written before the helper exists and observed RED (missing helper / missing wrapper), then GREEN after implementation; the mint-failure and stale-credential arms reproduce the defect class at unit scale (a fixture extraheader carrying token A stays token A when the mint fails — the degrade contract — and becomes token B when it succeeds — the fix). Assertions map one-to-one onto the AC arms; fixtures are a temp git config file carrying a known extraheader value and a stub mint endpoint; the real `gh`, real network, and real key are never used. Two properties have no unit harness and take reproducible-verification stand-ins recorded on the shipping PR: (1) the live >60-minute property (AC "Push freshness") — on the first natural writer run exceeding 60 minutes after merge, confirm in the run log that at least one refresh cycle fired and that a `git push` after minute 60 succeeded; (2) the background-step conclusion property (AC "Loud degrade") — a probe run on the PR branch demonstrating that a killed background refresher step leaves the job conclusion green. No behavioral-fix pins are claimed (the new prose pins are surface-presence class; the suite arms are functional tests). governing conventions consulted: `CLAUDE.md` (the six-shape adversarial config matrix does not apply — the change adds no config parser; the refresher consumes only workflow-provided env and secrets), `CONTRIBUTING.md`, `docs/` testing guidance.

**Documentation Needed** — `docs/cloud-setup.md` (mechanism, cadence, key handling, degrade behavior, upstream residual); a line in `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s workflow architecture section; a `.changeset/*.md` (patch bump — engine surface).

**Potential Gotchas**

- Never remove the checkout `token:` seeding — the refresher depends on the persisted extraheader existing (#357); this change rewrites it, never replaces the mechanism.
- The wrapper must exec the real `gh` by an absolute path captured at install time, before the `GITHUB_PATH` prepend takes effect — a name-based lookup would recurse into the wrapper itself.
- Never log the minted token; never log the key. The refresher runs with `set +x` semantics and prints only arm names.
- `actions/create-github-app-token`'s post-step revokes only the job-start token at job end; refresher-minted tokens expire naturally within an hour of their mint and need no revocation handling.
- `background: true` is three weeks old at drafting time — confirm the repo's actionlint version accepts the new keywords and upgrade it in the same PR if it lags. Job-end auto-cancel semantics for a still-running background step are not yet documented; the explicit pidfile-kill step above is what makes them non-load-bearing.
- The refresher parses REST responses with jq — it must source `lib/resolve-jq.sh` like every jq-calling helper, and it is a new gh-adjacent shell helper, so the `lib/test/run.sh` resolver pins apply to it.
- The refresher is cloud-only (ubuntu-latest); its tool checks still fail closed with a `::warning::` when a tool is missing (guard-class 2 discipline), never silently.
- Trust posture: the writer jobs are authorization-gated and check out trusted refs, so running a checkout-tree helper there matches those jobs' existing posture (contrast the review runner's #402/#458 trusted-source floors — the read-only review tier is deliberately out of scope here).
- The claude step's existing `DEVFLOW_APP_ID` env signal (Phase 1 Pass 5) is untouched.
- `lib/test/run.sh`'s `DEVFLOW_GH` stub contract is unaffected: the wrapper exists only in the cloud jobs' env; local suite runs never see it.
- The `GITHUB_PATH` prepend outlives the claude step: every post-claude step that invokes `gh` resolves it through the wrapper. The fingerprint rule keeps the fresh #287 backstop mints authoritative (hash differs → defer) while repairing devflow.yml's #356 flip step, whose `GH_TOKEN` is the job-start mint (hash matches → substitute); the observability-persist step invokes no `gh` and pushes via the extraheader the refresher maintains. Treat all three as coupled surfaces and verify them in the same change.

The01Geek:
## Summary
- Keeps `/devflow:implement` and `/devflow:review-and-fix` writer-job runs push- and `gh`-capable for their whole duration, so runs that outlive the GitHub App installation token's 60-minute lifetime stop dying mid-lifecycle on `Bad credentials` (the ~$60 evidence run 29299441781).
- A detached background refresher re-mints a fresh, **repo-scoped** installation token on a 45-minute cadence (2-minute backoff on failure) and rewrites the two repo-controlled credential surfaces in place — the checkout-persisted `http.<server>/.extraheader` push credential and a mode-0600 token file an installed `gh` wrapper reads at call time.
- With no App configured (`vars.DEVFLOW_APP_ID` empty) everything is skipped and behavior is byte-identical to today's `GITHUB_TOKEN` fallback.

## Changes
**Refresher (`scripts/refresh-app-credentials.sh`)**: new best-effort helper — single-cycle mint+rewrite plus a loop driver (45-min cadence, 2-min backoff, `TERM`-trap exit 0, pidfile, per-arm `::warning::`). Private key via stdin→shell memory→openssl over an fd (never env-visible-to-agent, argv, or disk). Mint scoped to the current repo (least privilege). jq via `lib/resolve-jq.sh`; extraheader locator uses bash builtins (no `sed`/`head` — BSD-portable, guard-class-2).

**`gh` wrapper (`scripts/gh-fresh.sh`)**: resolves the token at call time from the refresher-maintained file; discriminates the ambient job-start token (substitute) from a deliberately-fresh #287/#356 backstop mint (defer untouched) by sha256 fingerprint; degrades to a plain invocation when the token file is absent; appends a distinctive `devflow-gh-fresh:` diagnostic to stderr on a bad-credential failure. Wired both as `DEVFLOW_GH` and ahead of the real `gh` on `PATH`.

**Stop helper (`scripts/stop-refresher.sh`)**: retires the detached refresher and surfaces its health honestly — warns only on a genuine defeat (never-started/crash, or a sustained failure) and stays silent on a recovered transient or an upstream early-abort.

**Workflows (`devflow-implement.yml`, `devflow.yml`)**: both writer jobs gain start/install/stop steps gated on the same `DEVFLOW_APP_ID` condition as their App-token mint (`devflow.yml` additionally excludes `/devflow:review`).

**Skills prose**: two-strikes bad-credential fail-fast rule added to `skills/implement/SKILL.md` and `skills/review-and-fix/SKILL.md`.

**Tests / docs**: `lib/test/run.sh` gains 18 gh-stubbed arms (no network) + workflow-wiring pins, and reconciles the existing #357 checkout-token pins; `docs/cloud-setup.md` and `docs/DEVFLOW_SYSTEM_OVERVIEW.md` document the mechanism; `.changeset/*` (patch/Added).

## Resolves
Resolves #487

## Test Plan
- [ ] CI `lib + python tests` is green (18 `#487` arms + wiring pins; full suite 6744 passed, 0 failed in-env).
- [ ] `actionlint`, `shellcheck`, and `ruff` pass (no `background:` keyword; workflows lint-clean).
- [ ] With `DEVFLOW_APP_ID` unset, confirm the refresher/install/stop steps are skipped (unconfigured no-op).

## Post-Merge Verification
- [ ] **Push freshness:** on the first natural writer-job run exceeding 60 minutes after merge, confirm in the run log that at least one refresh cycle fired and that a `git push` after minute 60 succeeded — the refresher rewrites the checkout-persisted `http.<server>/.extraheader` credential every 45 minutes so the credential-of-record is never older than 45 minutes.
- [ ] **Loud degrade, never fatal:** a probe run on the PR branch demonstrating that a killed detached refresher process leaves the job conclusion green (the matcher-probe precedent), and that a failed mint cycle leaves the previous credential in place, emits a `::warning::`, and recovers on the 2-minute backoff.

## Upstream engagement
- This PR discloses a residual it cannot fix in-run: `claude-code-action`'s own internal API calls (its progress-comment updates) still ride the static `github_token` input until [anthropics/claude-code-action#716](https://github.com/anthropics/claude-code-action/issues/716) lands upstream. Evidence run: [29299441781](https://github.com/The01Geek/devflow-autopilot/actions/runs/29299441781) (75.7-minute `claude` job that burned budget fighting a dead push credential).
- **Named human follow-up:** post a comment on anthropics/claude-code-action#716 linking this PR as a downstream mitigation. A cloud `/devflow:implement` run's repo-scoped token cannot comment on a foreign repo, so the posting itself is out of the run's reach and is left to a maintainer.

## Visual Changes
N/A

## Breaking Changes
None. With `DEVFLOW_APP_ID` unset, workflow behavior is byte-identical to before this change.


The01Geek:
## Advisory review findings — resolution (commit 4bbb7503)

Addressing the four remaining non-gating findings from the last review pass.

### Fixed
- **Suggestion 1 (refresher TERM latency).** `cmd_loop` now sleeps in the background and `wait $!`s on it, so `trap 'exit 0' TERM` interrupts the wait and `stop-refresher.sh`'s kill retires the process promptly instead of waiting out the full 45-min interval. The `|| true` preserves the loop's never-exit-nonzero contract; arm4's `DEVFLOW_REFRESH_SLEEP` stub assertions (backoff→interval ordering) still hold.
- **Suggestion 3 (`--replace-all`).** The extraheader rewrite in `run_cycle` now uses `git config --replace-all`, so a key that ever held multiple values collapses to the one fresh credential instead of failing the set and going stale. New **arm21** seeds a fixture config with two extraheader values and asserts the cycle collapses them to the single fresh value (mutation-checked: without `--replace-all` the set fails and two values survive → RED).
- **Suggestion 5 (test coverage — the highest-value item).** The previously override-bypassed `real_mint()` production path is now desk-tested, no network:
  - **arm22** — real mint with a stub `curl` + real `openssl`/`jq`: the RS256 JWT header + claims decode correctly (`{"alg":"RS256",…}`, `iss`=app id), and the access-token POST body scopes to **this repo only** (`{"repositories":["<repo_name>"]}`) — converting the least-privilege key AC from attestation to evidence — plus the fresh token is written to both surfaces. (mutation-checked: stripping the repo scoping empties the asserted body → RED.)
  - **arm23** — mint returns an empty `.token` → previous credential intact + `::warning::`, exit 0.
  - **arm24** — `locate_extraheader_file` failure (no extraheader anywhere) → warn, no crash, exit 0.
  - **arm25** — `gh-fresh.sh` `resolve_real_gh` returns 127 when nothing resolves → the could-not-resolve breadcrumb.
  - **arm26** — the `Authentication failed` regex alternative in the wrapper's bad-credential scan (arm9 only exercised `Bad credentials`). (mutation-checked: dropping the alternative → diagnostic absent → RED.)

Suite: **6914 passed, 0 failed, 1 skipped** (the 1 skip is the `#434` stale-prose self-scan, which self-skips on a dirty working tree and runs live in CI). `shellcheck --severity=warning -e SC1091` clean on the touched script.

### Deferred
- **Suggestion 4 — wrapper stderr buffering.** `gh-fresh.sh` captures the real gh's stderr (`err="$( … 2>&1 1>&3 )"`) rather than streaming it, so stderr is delayed/reordered until the call returns.
  - **Why deferred:** the capture is load-bearing — it is exactly how the bad-credential signature (`HTTP 401`/`Bad credentials`/`Authentication failed`) is scanned to append the compaction-immune diagnostic while preserving the real gh exit code. A streaming `tee`-based rewrite would have to re-thread both the scan and the exit-code preservation, adding real complexity for no benefit to DevFlow's workload: every wrapped `gh` call is a short API/push, not a long-running streaming invocation, so the buffering window is sub-second and no output is lost — only briefly reordered.
  - **Revisit condition:** if DevFlow ever routes a long-running or streaming `gh` invocation through the wrapper (e.g. `gh run watch`, a long log tail), reopen this and switch to a streaming tee that still scans stderr and preserves the exit code.

PR #507, opened 2026-07-15
Persist explicit unavailable telemetry

### LINKED ISSUE #499: iter-<N>.json telemetry: persist an explicit "unavailable" instead of dropping the key (persist-time stamp, record normalization, durable-store backfill)
## Problem Statement

On a large fraction of recorded review-loop iterations, the `iter-<N>.json` workpad's `telemetry` block is simply absent instead of saying that measurement never happened. An absent key is not self-describing: a reader cannot distinguish "this iteration measured nothing" from "this iteration recorded zero" from the record alone. That violates the repo's load-bearing rule — *"Unknown is not zero — never collapse an unestablished measurement onto a real value"* (`CLAUDE.md`) — on the very dataset the weekly retrospective and experiment analyses consume. It has already caused a real misreading: during PR #438's skill-prose pressure testing, an agent read a key-absent record as "telemetry block is all zeros" and built on it the argument that the review loop "wasn't capturing cost anyway"; a second agent caught it only by opening the actual file (issue #452, item 1 — the provenance of this ticket).

The writer mandate already exists as prose (`skills/review-and-fix/SKILL.md` Step 3 item 7: the block is "mandatory") and as a best-effort warning (`--self-check`). Both are cooperative mechanisms, and the observed data shows they do not close the gap:

- **Verified (2026-07-14), telemetry branch, iter workpads:** 43 of 111 `iter-*.json` blobs lack the `telemetry` key — including one written the same day (`.devflow/logs/review/worktree-pr-480/local-20260714T044610Z-1/iter-1.json`). Command: enumerate `git ls-tree -r --name-only origin/devflow-telemetry | grep 'iter-.*\.json'`, test each blob with `git show origin/devflow-telemetry:<path> | jq -e 'has("telemetry")'` → total=111, absent=43.
- **Verified (2026-07-14), telemetry branch, derived run records:** 25 of 56 `.devflow/logs/efficiency/*.json` blobs carry at least one `telemetry[].phases: null` entry (the downstream image of the absent workpad key). Command: same enumeration over the efficiency prefix, testing `jq -e '(.telemetry // []) | map(select(.phases == null)) | length > 0'`.
- **Verified (2026-07-14), committed working tree (legacy archive):** 81 of 214 tracked `iter-*.json` under `.devflow/logs/review/` lack the key (grown from the 79 of 196 recorded when issue #452 was filed, so new key-absent files are still landing), and 43 of 98 tracked `.devflow/logs/efficiency/*.json` carry a `phases: null` entry — same test commands over `git ls-files`.

## Current Behavior

- The writer spec (`skills/review-and-fix/SKILL.md` Step 3 item 7) declares the `telemetry` block mandatory and names the wholesale-absent block as "the failure mode to eliminate" — prose only; observed compliance is ~60%.
- `--persist`'s durable workpad copy (`persist_one` in `lib/efficiency-trace.sh`) stages `cp -p "$dir"/*.json` byte-verbatim, so an absent key stays absent on the `devflow-telemetry` branch.
- The derived run record (`lib/efficiency-trace.jq`) maps `phases: .telemetry`, so an absent key becomes `phases: null`; the comment declares the pass-through "verbatim (unnormalized; null when absent)". Once derived, a record is never re-derived — `persist_one`'s presence-based idempotency deliberately leaves an existing branch record untouched forever — so a `null` that lands in the durable record store stays there permanently unless explicitly migrated.
- `--self-check` emits a best-effort `::warning::` per missing expected field — a breadcrumb, not a fix.
- Consumers collapse the shapes: `scripts/build-experiment-records.py`'s `_telemetry_complete` returns `False` on any falsy `phases`, so never-measured and recorded-none are indistinguishable downstream.
- No emit site anywhere produces the literal `unavailable` for this field, though the vocabulary already exists in the repo: `permission_denials_count` publishes the literal `unavailable` (`.github/workflows/devflow-review.yml`), with `scripts/describe-denial-count.sh` owning the three-way unestablished/none/N selection.

## Desired Behavior

An unestablished telemetry block reads as the explicit JSON string `"unavailable"` on the durable surfaces, applied by a deterministic path that runs regardless of the writing agent's cooperation (the `--persist` backstop: the Stop hook locally, the cloud-workflow wrapper in CI — per `skills/review-and-fix/SKILL.md`'s Common Mistakes persistence contract). Four decided applications:

1. **Stamp at persist time (durable workpad copy) — new branch paths only.** `--persist`'s durable-copy staging normalizes each staged `iter-<N>.json` whose `telemetry` key is absent-or-null to `"telemetry": "unavailable"`, per the closed input matrix below — **gated per file on the durable path not yet existing on the telemetry branch** (`devflow_telemetry_blob_exists`). A blob already on the branch is exclusively the backfill's territory: without this gate, the first post-upgrade Stop-hook persist on every machine would re-stage stamped bytes for each *retained legacy run dir* (this repo's `.devflow/tmp/review/` holds run dirs going back months), turning session-end into an uncoordinated distributed rewrite of historical branch paths — spurious commits from every machine, outside the migration's sequencing. Files needing no stamp keep their `cp -p` bytes untouched. The stamp itself modifies no tmp-run-dir file — the agent's as-written artifacts remain evidence; the pre-existing `synthesize_shadow_markers` in-place rewrite of promoted-shadow workpads is a separate, unchanged mechanism (its changed-bytes re-staging of an existing path is untouched by the stamp gate, which conditions only the *stamp*, never the copy).
2. **Normalize the derived run record.** The record's `telemetry[].phases` carries `"unavailable"` (not `null`) when the workpad's block is absent-or-null.
3. **Backfill the telemetry branch — both blob families.** An idempotent, **re-runnable-by-design**, maintainer-run migration script stamps the existing absent-or-null **iter workpad blobs** and rewrites `telemetry[].phases: null` entries to `"unavailable"` in the existing **derived run-record blobs** (`.devflow/logs/efficiency/*.json`) — the store `scripts/build-experiment-records.py` actually consumes, which the presence-based record idempotency would otherwise leave `null` forever. Repeat runs are the convergence mechanism for blobs written by pre-change code after the first migration (see the stated residual below) — a re-run selects only what needs rewriting and no-ops otherwise.
4. **Normalize the tracked legacy archive (this repo, one-time).** The same PR applies the same two rewrites to the tracked working-tree files — `.devflow/logs/review/**/iter-*.json` (absent-or-null key) and `.devflow/logs/efficiency/*.json` (`phases: null`) — as plain edits committed with the change, closing the surface the #452 misreading actually happened on.

Plus the aligned prose layer, at **both** writer sites of this block: `skills/review-and-fix/SKILL.md` Step 3 item 7 (the fix loop) and `skills/review/SKILL.md` Phase 4.5's workpad-shaped-object assembly (the standalone review's own `"telemetry": {…}` emit into `iter-1.json`) each instruct the writer to emit `"telemetry": "unavailable"` when the block's figures were never established, rather than omitting the key.

**Stated residual (accepted, named — two re-pollution channels):** (a) in-flight legacy branches created before this change can still merge pre-existing unstamped files into the committed tree after the migration, and clones/forks retain old shapes indefinitely; (b) **persists executed by pre-change code after the migration** — a stale worktree's Stop hook running an old checkout of `lib/efficiency-trace.sh`, and a consumer whose vendored tree is pinned to an older `devflow_version` — write fresh unstamped blobs onto the branch (observed live: between this issue's drafting probe and its audit, hours apart, the branch grew from 43/111 to 44/113 key-absent). So every reader keeps tolerating the legacy absent/`null` shapes; the deterministic stamp makes every persist **executed by post-change code** uniform; and re-running the idempotent backfill is the convergence mechanism for channel (b). Landing order: land the stamp; let in-flight writers drain (locally: refresh worktrees; consumers: bump the vendored `devflow_version` **before** backfilling); **invalidate diverged local telemetry refs across the fleet's machines/worktrees** (a diverged local ref is deliberately never overwritten by `do_persist`, so a machine that ever persisted offline holds a full pre-backfill snapshot that a later push-retry union would resurrect — see the union-monotonicity criterion, AC 12); then run the backfill.

The stamp condition is exactly absent-or-null **on a JSON-object workpad**. Classification is two-gated: gate 1 — the file parses and is a JSON object (a parse failure and a valid non-object — including a whole-file `null` — both route to M7); gate 2 — the key-presence/null test, applied to objects only. This two-gate order is load-bearing: a single combined predicate misroutes a whole-file `null` into the stamp arm, where jq's `null | .telemetry = "unavailable"` *fabricates* `{"telemetry": "unavailable"}` — destroying the evidence of a truncated write (verified by execution at drafting time). An explicit `{}` or all-zeros object is a real recorded value and is preserved verbatim (zero is not unknown, exactly as unknown is not zero); a wrong-type `telemetry` value is preserved verbatim as established-but-malformed data.

**Input matrix (closed set — exactly these seven, complete by construction: the `CLAUDE.md` six-shape adversarial matrix mapped onto this surface, plus the marker row M6):**

| row | input (`telemetry` in a staged/branch `iter-*.json`) | outcome |
|---|---|---|
| M1 | populated object | preserved byte-verbatim |
| M2 | explicit `{}` (likewise an all-zeros object) | preserved byte-verbatim — established-empty is not unknown |
| M3 | key absent | stamped `"telemetry": "unavailable"` |
| M4 | value JSON `null` | stamped `"telemetry": "unavailable"` |
| M5 | wrong-type value (array, number, boolean, non-marker string) | preserved byte-verbatim |
| M6 | already the marker `"unavailable"` | preserved byte-verbatim (idempotent) |
| M7 | file is not a JSON object — malformed JSON, likewise valid non-object JSON (whole-file `null`, bare array, bare string) | copied/left byte-verbatim + named `::warning::` breadcrumb — not stamped, not dropped, not fabricated |

**Record-family input matrix (closed set — exactly these seven, complete by construction: the same convention applied to the run-record surface's own shape space, which M1–M7 does not cover):**

| row | input (a `.devflow/logs/efficiency/*.json` blob) | outcome |
|---|---|---|
| R1 | `telemetry` array of objects, an entry with `phases: null` | that entry's `phases` rewritten to `"unavailable"`; all other bytes preserved |
| R2 | an entry carrying **no `phases` key at all** | **not selected — the key is never fabricated** (the record-family sibling of the whole-file-`null` fabrication guard) |
| R3 | `telemetry` absent, `null`, or `[]` at the record level | preserved byte-verbatim — record-level absence is not an entry-level `null` |
| R4 | `telemetry` wrong-type at the record level (object, string, number, boolean) | preserved byte-verbatim + named breadcrumb |
| R5 | an entry that is not a JSON object (a scalar in the array) | whole record preserved byte-verbatim + named breadcrumb — never a jq abort |
| R6 | an entry with `phases` already `"unavailable"` (likewise any non-null `phases`) | preserved byte-verbatim (idempotent) |
| R7 | file is not a JSON object (malformed, likewise valid non-object JSON) | left byte-verbatim + named breadcrumb — not rewritten, not dropped, not fabricated |

## User Impact

DevFlow maintainers, the weekly retrospective, experiment analyses, and any human or agent reading a durable iter record can distinguish "never measured" from "measured zero"/"recorded none" from the record alone — in the iter workpads *and* in the derived run-record store `build-experiment-records.py` actually reads. The misreading class behind #452's evidence — an argument built on a vanished key — is eliminated for every persist executed by post-change code and, after the migration, for both durable stores and this repo's tracked legacy archive (the two re-pollution channels in the stated residual above are the named exceptions, converged by backfill re-runs). Consumer repos inherit the same guarantee through the shipped helper and can normalize their own telemetry branches with the same in-tree migration script.

## Technical Context

- **Relevant Classes/Files**
  - `lib/efficiency-trace.sh` — `persist_one`'s durable-copy staging block (the stamp seam; the staging comment "Content-idempotent: the branch write's tree-equality no-op guard emits no commit when the bytes are unchanged" is a coupled prose site, as is the record-idempotency comment "A record already on the branch (a prior --persist) is left untouched").
  - `lib/efficiency-trace.jq` — the two `.telemetry` read sites: the record's `telemetry:` map (`($iters | map({iter: .iter, phases: .telemetry}))`, comment "null when absent") and the per-iteration normalization (`telemetry: ($it.telemetry // null)`); both updated with the same explicit predicate so no newly derived record emits a bare `null` for an absent block.
  - `lib/telemetry-branch.sh` — the staged-tree write path the backfill reuses (`devflow_telemetry_list_blobs`, `devflow_telemetry_show_blob`, `devflow_telemetry_persist_tree`), and `commit_union_on` (the AC 12 marker-monotonic collision rule).
  - `scripts/backfill-telemetry-unavailable.sh` — NEW: the re-runnable migration (both blob families).
  - `.devflow/logs/review/**/iter-*.json` + `.devflow/logs/efficiency/*.json` (tracked) — the legacy archive normalized by the same PR.
  - `skills/review-and-fix/SKILL.md` — Step 3 item 7 (record shape) and the Run telemetry summary capture rules — and `skills/review/SKILL.md` — Phase 4.5's workpad-shaped-object assembly, the standalone review's independent writer of the same `telemetry` block (the two writer-vocabulary sites).
  - `scripts/build-experiment-records.py` — consumer pins only; no code change expected (see Verified).
  - `docs/efficiency-trace.md`, `lib/test/run.sh`.
- **Architecture Alignment** — the stamp rides the existing #441 persist model (staging + tree-equality no-op guard + CAS push); the marker reuses the repo's existing unestablished-vocabulary literal `unavailable`; the migration reuses the telemetry-branch plumbing rather than adding a write path.
- **Dependencies** — none new: bash, jq (via sourced `lib/resolve-jq.sh`), and the git plumbing already inside the persist path. No PyYAML, no `gh`, no new command heads.
- **Data/Schema Considerations** — `telemetry` in a durable `iter-<N>.json` becomes `object | "unavailable"` (plus preserved legacy/wrong-type shapes); the run record's `telemetry[].phases` becomes `object | "unavailable" | <verbatim wrong-type>`. `ITER_EXPECTED_FIELDS` is unchanged (the marker satisfies the presence check); `ITER_SYNTH_EXPECTED_FIELDS` validation is unaffected (the marker is an extra key on a synthesized record, not a required one). No `schema_version` bump — the field vocabulary is additive and consumers already tolerate the shapes (see Verified).
- **Cross-layer Impact** — `lib/` (stamp + jq predicate + union rule), `scripts/` (new backfill; consumer pins), `skills/` (writer prose), tracked logs archive, docs, tests. The stamp applies wherever `--persist` stages — the writable tiers, the local Stop hook, and the read-only review tier's staging-only persist (#469), whose staged artifacts the #489 relay would land already-stamped.
- **Provenance** — refines issue #452 item 1 (an explicitly-unrefined placeholder; item 2 there is unrelated and stays its own ticket). Related but distinct open issues, confirmed non-overlapping: #475 (cloud-only `harness_cost` floor — its own text declares it "deliberately leaves open" this contract; see Gotchas for the shared-function coordination) and #489 (transport for read-only-tier records).

**Verified (drafting-time probes, 2026-07-14):**
- **Counts** — branch iter blobs 43/111 key-absent; branch run records 25/56 with a `phases: null` entry; tracked tree 81/214 iter files and 43/98 run records (commands in Problem Statement).
- **Stamp seam** — `persist_one` stages the durable copy via `mkdir -p "$durable" && cp -p "$dir"/*.json "$durable"/` into `_TELEMETRY_STAGE/.devflow/logs/review/<slug>/<run-id>`, and the branch write's tree-equality no-op guard makes unchanged bytes commit-free (code read, `lib/efficiency-trace.sh`).
- **Record derivation reads the tmp workpads, not the staged copies** — `emit_jq record` consumes `collect_valid_files "$dir"` (code read), which is why the record arm needs its own jq normalization and is not covered by the stamp.
- **Record idempotency is presence-based** — an existing branch record is "staged for neither derivation nor write" (code read of `persist_one`'s `devflow_telemetry_blob_exists` gate), which is why existing `phases: null` records can only be fixed by the explicit migration, never by the new derivation predicate.
- **Whole-file-`null` fabrication hazard** — executed at drafting time: the combined predicate `type == "object" and has("telemetry") and (.telemetry != null)` exits 1 on a whole-file `null` (indistinguishable from M3/M4 by exit status), and `printf 'null' | jq '.telemetry = "unavailable"'` succeeds, emitting `{"telemetry":"unavailable"}` — the observed output grounding the two-gate classification.
- **Consumer safety** — `_accumulate_cost` recurses only into dict/list nodes: a string `"unavailable"` contributes nothing and returns `False`, so `_telemetry_complete` returns `False`, `_run_cost` counts nothing, and no exception is raised (code read, `scripts/build-experiment-records.py`).
- **Backfill write path** — `lib/telemetry-branch.sh`'s staging guard admits exactly `.devflow/logs/*`-shaped paths (code read of the staging case arm), so both blob families' rewrites are inside the accepted set.
- **Vocabulary precedent** — the literal `unavailable` is published today by `permission_denials_count` (`.github/workflows/devflow-review.yml`), selection owned by `scripts/describe-denial-count.sh` (code read).
- **Persist-time workpad mutation precedent** — `synthesize_shadow_markers` rewrites tmp-run-dir iter files in place (jq → temp file → `mv`) when recovering a dropped-but-promoted shadow block (code read); the stamp deliberately diverges by mutating only the *staged copy*.
- **Second writer site** — `skills/review/SKILL.md` Phase 4.5 independently assembles a workpad-shaped object carrying its own `"telemetry": {…}` block and writes it to `.devflow/tmp/review/<slug>/<run-id>/iter-1.json` (prose read) — review-mode records persist through the same `persist_one` path (the #441 unification), so both writer sites carry the emit rule.
- **Wild-type sweep** — a type sweep of all 113 branch iter blobs at drafting time found `telemetry` only `absent` (44) or `object` (69) — zero falsy/wrong-type instances — grounding AC 4's accepted approximation. The same probe window observed live re-pollution (43/111 → 44/113 in hours), grounding the stated residual's channel (b).
- **Union overlay is local-wins on collision** — `commit_union_on` overlays every blob of the local tip onto the fetched remote base via `update-index --add --cacheinfo`, its correctness comment resting on "Per-run filenames are unique … a pure union"; and `do_persist` deliberately never overwrites a diverged local telemetry ref (code read, `lib/telemetry-branch.sh` + `lib/efficiency-trace.sh`) — together the mechanism behind AC 12.
- **Retained legacy run dirs + unconditional durable staging** — `.devflow/tmp/review/` holds run dirs from months-old runs, and `persist_one` stages the durable copy unconditionally per discovered run dir (only the *record* is presence-gated) (code read + directory listing) — the mechanism behind AC 1's blob-exists gate.
- **Live legacy-merge channel** — `chore: persist review-and-fix observability artifacts` commits landed on `main` via legacy branches as recently as the day before drafting (`git log` read) — the mechanism behind AC 5a's deliberate no-standing-gate choice.

## Acceptance Criteria

- [ ] 1. **Stamp arm — new branch paths only.** `lib/efficiency-trace.sh --persist`'s durable-copy staging implements the M1–M7 matrix over staged `iter-*.json` (and only `iter-*.json` — sibling files such as `deferrals.json` are untouched), using the two-gate classification (object-parse gate before key/null gate), **and only for a file whose durable path does not yet exist on the telemetry branch** (`devflow_telemetry_blob_exists` gate — an existing blob is exclusively the backfill's territory): M3/M4 files are staged with `"telemetry": "unavailable"`; M1/M2/M5/M6 files are staged byte-identical to their source; M7 files — including the whole-file-`null` fixture — are staged byte-identical with a named `::warning::` breadcrumb. The stamp itself modifies no tmp-run-dir file: byte-identity of the run dir is asserted on a fixture without the promoted-shadow shape that the pre-existing `synthesize_shadow_markers` legitimately rewrites in place (that mechanism is out of scope and unchanged — the blob-exists gate conditions the stamp, never the copy, so shadow-marker updates of existing paths still flow).
- [ ] 2. **Record arm.** A newly derived run record's `telemetry[]` entries carry `phases: "unavailable"` for M3/M4 workpads and the verbatim value for M1/M2/M5/M6 — pinned per matrix row, including both type-boundary pairs: explicit `{}` yields `phases: {}` (not the marker), and boolean `false` yields `phases: false` (not the marker — the predicate tests key-presence and null explicitly rather than using jq's `//` operator, which also fires on `false`).
- [ ] 3. **Idempotency — both directions.** A second `--persist` over an already-stamped run makes no new telemetry-branch commit (it ends at the existing tree-equality no-op guard) — pinned by a re-run fixture comparing the ref before/after. And a `--persist` discovery pass over a **legacy M3 run dir whose durable blob already exists unstamped on the branch** likewise makes no new commit (the AC 1 blob-exists gate leaves the staged bytes equal to the branch blob) — pinned by a legacy-run-dir fixture; this is the assertion that session-end persists never become an uncoordinated rewrite of historical paths.
- [ ] 4. **Backfill script — both blob families, re-runnable by design.** A new `scripts/backfill-telemetry-unavailable.sh` (SPDX header; best-effort exit-0 with a specific stderr breadcrumb on every non-happy branch; jq via sourced `lib/resolve-jq.sh`) rewrites, in one new branch commit through `lib/telemetry-branch.sh`'s staged-tree write path followed by a push: (a) every `iter-*.json` blob on the telemetry branch in class M3/M4 to carry the marker, and (b) every `.devflow/logs/efficiency/*.json` run-record blob per the **R1–R7 record-family matrix** — only an entry that *carries* the `phases` key with value `null` (R1) is rewritten; an entry with no `phases` key is never fabricated one (R2); record-level absent/null/empty `telemetry`, wrong-type shapes, and non-object entries are preserved byte-verbatim with a breadcrumb and never abort the run (R3–R5, R7). For R1, the accepted approximation is stated in the script's header: a legacy record `null` can in principle mask a falsy-but-established workpad value (the old `//`-based derivation nulled every falsy value, not only absent/null), and the rewrite treats all of them as unestablished — accepted because the `null` already erased the distinction, and a drafting-time type sweep of all branch iter blobs found only `absent` and `object` values in the wild (zero falsy/wrong-type instances). Iter-family classification uses the same two-gate discipline as AC 1; M1/M2/M5/M6 blobs and R2/R3/R6-class records are byte-untouched; M7/R7-class blobs are left in place with a per-blob breadcrumb; a run that selects nothing makes no commit — which is what makes repeat runs the convergence mechanism for the stated residual's channel (b); an absent or unresolvable telemetry branch draws a named breadcrumb and exit 0. The script is invoked by no skill, workflow, or hook — a maintainer-run migration kept in-tree so consumer repos can normalize their own branches — asserted in the same change by a whitespace-normalized `git grep` of the script name over `skills/`, `.github/`, and the Stop-hook target list, cited in the PR.
- [ ] 5. **Migration is a maintainer post-merge step — not an implement-run obligation.** The implement run's tickable obligation for this criterion is complete when the script and its fixtures ship (AC 4) and the docs state the migration procedure (AC 9): running the backfill against this repo's live `devflow-telemetry` branch, re-running the count probes from the Problem Statement (drafting-time observed: 43 of 111 iter blobs and 25 of 56 run records), and recording the outputs is performed **by a maintainer after merge** (none of the backfill invocation, `git ls-tree`, or `git show` probe forms is granted in `devflow_implement.allowed_tools`, and an autonomous run must not go Blocked on them). The docs name this step and its expected before/after counts so the maintainer action is auditable.
- [ ] 5a. **Legacy archive normalized (tracked tree) — one-shot, deliberately NOT a standing gate.** The same PR rewrites the tracked `.devflow/logs/review/**/iter-*.json` files in class M3/M4 (drafting-time observed: 81 of 214) and the tracked `.devflow/logs/efficiency/*.json` files per the R1–R7 matrix (drafting-time observed: 43 of 98 with a null-phases entry), as plain committed edits in a mechanically-generated, separately-reviewable commit; the zero-remaining verification is a **one-shot probe recorded in the PR description at review time**, not a `lib/test/run.sh` assertion. The no-standing-gate choice is deliberate and load-bearing: `lib + python tests` is this repo's *required* merge check, and the stated residual's channel (a) is alive (legacy-branch merges landed key-absent files on `main` as recently as the day before drafting) — a standing cleanliness assertion would go RED on `main` at the next such merge and freeze every open PR on a failure unrelated to its diff. M7/R7-class tracked files are left byte-verbatim and named in the PR description.
- [ ] 6. **Writer spec — both writer sites.** `skills/review-and-fix/SKILL.md` Step 3 item 7 states the emit rule — when the block's figures were never established, the loop writes `"telemetry": "unavailable"` instead of omitting the key — with the Run telemetry summary's capture rules carrying the same vocabulary; and `skills/review/SKILL.md` Phase 4.5's workpad-shaped-object assembly (the standalone review's independent writer of the same block) states the same rule for its `"telemetry"` emit. Both files edited under the repo's `writing-skills` discipline.
- [ ] 7. **Consumer pins.** `lib/test/run.sh` pins `scripts/build-experiment-records.py` on a marker-bearing record: `_telemetry_complete` is `False`, `_run_cost` contributes no figures from the marker entry, and the script exits cleanly. The drafting-time code read shows today's code already behaves this way — the pins guard regression; no consumer code change is expected.
- [ ] 8. **Synthesized records.** The stamp applies uniformly: a synthesized (`synthesized: true`) workpad staged without the key is stamped like any other (its telemetry is by construction never established); `--self-check`'s `ITER_SYNTH_EXPECTED_FIELDS` validation output is unchanged by the extra key (asserted by a synthesized-workpad fixture).
- [ ] 9. **Coupled prose reconciled in the same change.** `docs/efficiency-trace.md`'s pass-through sentence ("verbatim (unnormalized; `null` when the workpad recorded none)") and its writer-gap paragraph ("the writer gap … is closed … never on a normal full-engine run"); the jq comment above the record's `telemetry:` map; `persist_one`'s durable-copy comment and its record-idempotency "left untouched" comment; `skills/review/SKILL.md` Phase 4.5's telemetry-shape example (the second writer site from AC 6); `lib/telemetry-branch.sh`'s "Per-run filenames are unique … a pure union" comment (superseded by the AC 12 monotonic collision rule); the maintainer migration procedure with its expected before/after counts (the AC 5 auditability requirement); and any `lib/test/run.sh` pins asserting the superseded prose — all updated together, enumerated by a whitespace-normalized search cited in the PR.
- [ ] 10. **Mutation evidence.** Every new behavioral pin goes through `assert_pin_red_under` with a recorded mutation run (e.g. a `sed -E` mutation replacing the explicit key-presence/null predicate with a bare `//` must turn the M5 boolean-`false` pin RED, and one collapsing the two gates into the combined predicate must turn the whole-file-`null` M7 pin RED).
- [ ] 11. **No new tool grants; changeset.** No agent-facing skill or phase file gains a new command head (the stamp lives inside the already-granted `lib/efficiency-trace.sh`; the backfill appears in no skill/workflow fence per AC 4; the AC 6 SKILL.md edit is prose around the existing Write-tool emit, not a new fenced command). The change adds a `.changeset/*.md` (`bump: patch`).
- [ ] 12. **Union monotonicity (the un-migration guard).** `lib/telemetry-branch.sh`'s `commit_union_on` gains a **marker-monotonic collision rule**: when a path exists in both the fetched remote base and the local overlay, and the local bytes are in a rewrite-selected class (an M3/M4-shaped iter blob; a record blob with an R1-selected entry) while the remote bytes are not, the **remote side is kept**; every other collision keeps today's local-wins behavior byte-for-byte. Without this, the existing pure union — whose stated correctness premise, "Per-run filenames are unique … a pure union", this issue is the first to break — lets any machine holding a diverged pre-backfill local ref silently overwrite every migrated path on its next push-retry (a stale snapshot is retained forever by design: `do_persist` never overwrites a diverged local ref). Pinned by a two-writer fixture: writer B holds a stale unstamped snapshot of paths writer A's backfill migrated; B's union re-parent must leave the migrated bytes in place (fails RED under today's overlay). The superseded "pure union" comment is reconciled in the same change (AC 9).

## Implementation Notes

**Approach.** Three seams, smallest first:

1. **Stamp** (`lib/efficiency-trace.sh`, `persist_one`): after the `cp -p` staging, iterate the staged `iter-*.json`; skip any file whose durable path already exists on the branch (`devflow_telemetry_blob_exists` — the AC 1 gate); classify the rest with the two-gate classification — gate 1: the file parses and `jq -e 'type == "object"'` holds (a parse failure and a non-object, including whole-file `null`, route to M7: breadcrumb, bytes untouched); gate 2: on objects only, `jq -e 'has("telemetry") and (.telemetry != null)'` — false selects the stamp (`.telemetry = "unavailable"` into a temp file, then move into place). Every non-selected file keeps its `cp -p` bytes. **Test-then-rewrite, never an unconditional re-serialize** — re-serializing every copy would churn whitespace/key-order bytes on all legacy runs, defeating the tree-equality no-op guard and forcing a spurious branch commit per historical run on its next persist. No new exec edge (jq is already resolved via the sourced resolver — the helper is a #458 Stop-hook trusted-closure target).
2. **Record normalization** (`lib/efficiency-trace.jq`): change the two `.telemetry` read sites to the same explicit key-presence/null predicate emitting `"unavailable"` — not the `//` operator (it also fires on boolean `false`, which M5 preserves).
3. **Backfill** (`scripts/backfill-telemetry-unavailable.sh`): resolve the telemetry ref (`devflow_telemetry_ref`), enumerate both blob families (`devflow_telemetry_list_blobs` over `.devflow/logs/review/` and `.devflow/logs/efficiency/`), classify each blob against its family's matrix (`devflow_telemetry_show_blob` — iter family: the M1–M7 two-gate discipline; record family: R1–R7, rewriting only R1-selected entries and never fabricating a `phases` key), rewrite selected blobs into a staging tree, write one commit through `devflow_telemetry_persist_tree`, push the branch the same way `--persist` does. The AC 12 marker-monotonic union rule is what keeps these rewrites durable against a stale diverged local ref's later push-retry. Idempotence falls out of the predicates: a run over an already-converged store selects nothing — which is what makes the script re-runnable as the convergence mechanism for late-arriving pre-change-code blobs. The tracked-tree normalization (AC 5a) is the same classification applied as plain working-tree edits in the PR — no plumbing involved.

Then the writer prose (SKILL.md item 7 + capture rules), the coupled doc/comment sites, and the pins.

**Code Patterns.** Persist-time workpad enrichment (`synthesize_shadow_markers` — with the deliberate staged-copy divergence noted in Verified); best-effort exit-0 + specific stderr breadcrumbs (`scripts/ensure-label.sh` contract); unknown-is-not-zero three-way vocabulary (`scripts/describe-denial-count.sh`); decisive values via jq/bash builtins only, never `tr`/`sed`/`wc`/`cut` (guard-class 2); jq through sourced `lib/resolve-jq.sh` (the pinned resolver convention).

**Testing Strategy.** Boundary: CLI/exit-code + JSON-contract surfaces, all drivable by `lib/test/run.sh` — unit level for the two-gate classifier and the backfill script (against a fixture telemetry branch in a scratch repo, the suite's existing pattern), integration level for `--persist` end-to-end staging.
governing conventions consulted: `CLAUDE.md` (the six-shape adversarial input matrix — `iter-<N>.json` is agent-authored, hand-corruptible input, so the matrix applies in full, mapped as M1–M7; the run-record surface is a distinct input type under the same convention, mapped as its own R1–R7); `CONTRIBUTING.md` and `docs/` searched — no additional governing matrix found.
Named assertions (each maps to the AC of the same number; every AC maps to at least one):
- A1: fixtures per matrix row — M3/M4 staged output carries the marker; M1/M2/M5/M6 staged bytes equal source bytes (`cmp`); M7 fixtures are specifically whole-file `null`, bare array, bare string, and malformed JSON, each asserting the named breadcrumb plus the verbatim copy (the whole-file-`null` fixture is the fabrication regression guard); plus the run-dir-untouched assertion (`cmp` before/after, on a fixture without the promoted-shadow shape) and the siblings-untouched assertion (`deferrals.json` byte-identical). The M3 fixture is production-realistic: a captured key-absent blob from the branch (e.g. `worktree-pr-480/local-20260714T044610Z-1/iter-1.json`), not only hand-built tokens. Each stamp assertion is written test-first and observed failing RED against today's `cp -p`-verbatim staging — the failure reproduces the reported defect (the staged copy lacks the key).
- A2: record-derivation fixtures per matrix row, including the `{}` and boolean-`false` type-boundary pairs asserted to pass through verbatim (each must fail RED under a `//`-based implementation).
- A3: the re-run fixture asserts an unchanged branch ref (`git rev-parse` before/after the second persist).
- A4: the backfill driven against a fixture branch holding one iter blob per M-row *plus* one record blob per R-row — selected blobs rewritten (R1 entries only; the R2 no-fabrication fixture must fail RED under a `select(.phases == null)`-style predicate that cannot distinguish key-absent from null), all others byte-identical (`git cat-file` + `cmp`), no jq abort on the R5 scalar-entry fixture, exactly one new commit, second run a no-op, absent-branch arm asserts the named breadcrumb and exit 0.
- A12: the two-writer un-migration fixture — writer B's stale unstamped snapshot unions against writer A's migrated tip; the migrated bytes survive (fails RED under today's local-wins overlay); plus a control asserting a non-selected-class collision still resolves local-wins byte-for-byte.
- A5/A5a: both are one-shot, human-audited verifications, deliberately outside the standing suite (the AC 5a rationale): A5a's zero-remaining probe over `git ls-files` is run at review time and recorded in the PR description; A5 is the documented maintainer post-merge procedure — the docs name the steps and expected before/after counts (drafting-time: 43/111 branch iter, 25/56 branch records), with the actual probe re-runs performed and recorded by the maintainer after merge.
- A6: prose pins on the SKILL.md emit-rule sentence (surface-presence class — plain `assert_pin_unique`, no mutation obligation).
- A7: a marker-bearing record fixture through `scripts/build-experiment-records.py` asserts `telemetry_complete` false, no cost contribution from the marker entry, clean exit.
- A8: a synthesized-workpad fixture asserts the stamp lands and `--self-check` output is unchanged.
- A9: the superseded-prose pins reconciled; the new docs sentence pinned to the new contract.
- A10: each behavioral pin records its `assert_pin_red_under` mutation run.

**Documentation Needed.** `docs/efficiency-trace.md` — the pass-through and writer-gap paragraphs, the marker vocabulary, the two-gate classification, the backfill's two-family scope and re-runnable convergence role, and the maintainer migration procedure with its landing-order rule (stamp first, drain/bump writers, then backfill; consumers upgrade the vendored `devflow_version` before backfilling) and expected before/after counts; `docs/DEVFLOW_SYSTEM_OVERVIEW.md` where it describes the record shape (trace it per the scope note).

**Potential Gotchas.**
- **#475 coordination (shared functions, append-only posture).** #475's merge arm and this change's backfill both mutate previously-immutable branch paths, and both issues edit `persist_one`/`do_persist` and the union machinery — #475's AC 5a makes `commit_union_on` merge-aware for `harness_cost` exactly as this issue's AC 12 makes it marker-monotonic; the two collision rules must compose, and whichever lands second reconciles against the other's changes in the same functions. Note the AC 12 rule is the actual guard here — a "run the backfill when nothing is in flight" precaution does **not** close the window, because the diverged local snapshot pre-exists the backfill and surfaces on an arbitrarily later push-retry.
- **Byte-churn hazard.** An implementation that re-serializes all staged copies instead of test-then-rewrite silently converts every legacy run's durable copy into a fresh branch commit on its next persist.
- **Fabrication hazard (the M7/null boundary).** The single-predicate shortcut mis-sorts a whole-file-`null` workpad into the stamp arm and fabricates a marked record from a truncated write — the two-gate order in the Approach is the guard, and the A1/A10 whole-file-`null` pins are its regression evidence.
- **#458 Stop-hook closure.** `lib/efficiency-trace.sh` is a Stop-hook target; the stamp must add no new exec edge (pure jq/bash through the already-sourced resolver).
- **Guard-class 2.** The stamp/backfill selection predicate is derived via jq and bash builtins, never `tr`/`sed`/`wc`/`cut` — a missing PATH tool must not fail the selection open.
- **Marker literal is coupled vocabulary.** The string `unavailable` matches the `permission_denials_count` publication; keep the literal byte-identical, and enumerate with a whitespace-normalized search when touching it.
- **Offline/shallow clones.** The backfill needs the telemetry ref resolvable; the absent-branch arm is the fail-closed breadcrumb path, and the plumbing's existing bounded fetch is the recovery route.
- **Tracked-archive edits are data edits, not code edits.** The AC 5a normalization touches 100+ tracked JSON files; keep it a mechanically-generated, separately-reviewable commit inside the PR so the reviewer can verify it against the classification rather than reading each file.

The01Geek:
## Summary
- Persist an explicit `"unavailable"` marker when iteration telemetry was never established, without collapsing valid falsy values.
- Provide a re-runnable telemetry-branch backfill and prevent stale writers from undoing migrated blobs during union retries.
- Normalize the tracked legacy archive in a separate mechanical commit and align writer guidance and migration documentation.

## Changes
**Persistence and records**: Stamp new durable iteration blobs after object/key gates, preserve existing historical paths, and derive marker-aware efficiency records.

**Migration and concurrency**: Add the maintainer-run backfill for iteration and efficiency-record families plus marker-monotonic collision handling in telemetry branch unions.

**Archive and contracts**: Normalize 81 tracked iteration blobs and 43 tracked efficiency records; update both writer skills, comments, documentation, and the patch changeset.

## Resolves
Resolves #499

## Test Plan
- [x] Run `lib/test/run.sh` on the committed tree: 7,258 passed, 0 failed.
- [x] Run `bash -n` and `shellcheck` over the modified shell scripts; only informational existing-source warnings remain.
- [x] Verify tracked archive probes report `iter_remaining=0` and `record_remaining=0`.
- [x] Exercise the absent telemetry-ref backfill path and confirm exit 0 with the named breadcrumb.
- [x] Confirm mutation guards turn RED when the falsy-safe predicate or whole-file object gate is collapsed.

## Post-Merge Verification
The following item can only be verified after this PR is merged. Tick it after performing the check.
- [ ] Run `scripts/backfill-telemetry-unavailable.sh` against the live `devflow-telemetry` branch, rerun the documented before/after probes, and record the outputs.

## Visual Changes
N/A

## Breaking Changes
None. Legacy telemetry is normalized through an explicit, re-runnable maintainer migration.


The01Geek:
Review reception update for `640d3ee9`:

Addressed the two Important findings with behavioral coverage:
- stale legacy-vs-normalized union collisions now prove both directions: normalized remote migration data wins, while ordinary established collisions remain local-wins;
- classifier rc>=2 now has an integration test proving the union is refused and breadcrumbed.

Also fixed the demonstrable correctness edge: an established durable telemetry value can no longer be downgraded to `"unavailable"` by a retained/stale source whose telemetry key is absent or null. The `existing_class` local declaration was cleaned up in the same change.

Deferred below the configured Important threshold:
- a dedicated non-object durable-overlay fixture: current coverage verifies the non-object gate and the new stale legacy union path, but does not isolate that exact overlay shape; revisit if persist classification/overlay rules change or a regression is observed.
- a dedicated M7 backfill byte-preservation/breadcrumb fixture: production behavior is covered by the shared object gate and reviewed implementation, but not a standalone M7 case; revisit when the backfill classifier changes.

Verification: `GITHUB_ACTIONS=true bash lib/test/run.sh` → 7,318 passed, 0 failed. Mutation probes independently turned the new monotonic-union and classifier-refusal assertions red.

PR #525, opened 2026-07-16
Harden the retrospective-weekly loop: reflection-severity gate, state-PR stdout hygiene, and stale-scratch cleanup

### LINKED ISSUE #519: Harden the retrospective-weekly loop: reflection-severity gate, state-PR stdout hygiene, and stale-scratch cleanup
## Problem Statement

Running `/devflow:retrospective-weekly` end-to-end on a 42-PR backlog surfaced three defects in the loop and its helpers. They hurt, in order: **cost** (the cheap gate forces expensive LLM analysis on PRs whose only "signal" is an informational note), **correctness** (the state-PR helper's stdout is polluted, breaking the unattended report step), and **robustness** (per-run scratch is never cleaned, so it accumulates without bound and can be read stale). The maintainer running the weekly loop — and the unattended `claude -p` cron variant — hits all three every run.

## Current Behavior

**1. No reflection-severity discrimination (cost).** `lib/cheap-gate.jq` (final `elif`, the reflection check) trips on `($reflections | length) > 0`: **any** non-empty `## Devflow Reflection` block forces a Stage-A LLM analysis, regardless of the bullets' kind. The reflection contract in `skills/implement/SKILL.md` already groups bullets into three severity sections — `### ⚠️ Action required` (kinds `blocked`/`deferred`/`dropped-failed`), `### 💡 Improvements` (kind `improvement`), and `### ℹ️ Notes` (kinds `note` and `issue-accuracy`) — but that structure is discarded at extraction: `lib/fetch-pr-context.sh`'s reflection parser captures only bullet lines (`^\s*[-*]\s+…`) and drops the `### ` sub-headings, so `cheap-gate.jq` receives a flat string array with no severity tag. On the 42-PR run, a large fraction of analyzed PRs were mechanically clean (Complete, APPROVE, 0 post-bot commits, 0 CI failures) and were analyzed **solely** because they left an informational `ℹ️` `note`-kind bullet (the `note` kind is for "an informational deviation you worked around" — `skills/implement/SKILL.md`'s reflection-style contract). Seven independent Stage-A subagents flagged this and proposed a severity discriminator (from PRs #508, #470, #495, #482, #483, #486, #509). (The contract already says a *frictionless* clean confirmation belongs in `## Progress` as a `--note`, not in Devflow Reflection — this change targets the residual informational `note`-kind reflections that legitimately exist for worked-around deviations, and it keeps every *actionable* kind, including `issue-accuracy`, forcing analysis.)

**2. `open-state-pr.sh` leaks git and `gh` output to stdout (correctness).** `lib/open-state-pr.sh`'s stdout contract is "print only the PR number" — the retrospective skill's state-PR step relies on it via `STATE_PR=$(bash open-state-pr.sh)` and later feeds `$STATE_PR` to `jq` as `--argjson state_pr`. The script is careful to send `gh pr create`'s URL to stderr (the `gh pr create … >&2` call, with an explicit "keep it off our stdout" comment). But three other commands leak to stdout:
- `git checkout -B "$BRANCH" "$BASE"` (via the `_run` helper) prints carried-over `M<TAB>file` modification lines;
- `git commit -m …` prints its `[branch hash] subject` / `N files changed` summary;
- on the **branch-reuse path** (`[ -n "$EXISTING_PR" ]`, taken whenever the per-run branch `devflow/learnings-<date>` already has an open PR — i.e. any second run on the same UTC day), `"$DEVFLOW_GH" pr edit "$EXISTING_PR" --title "$SUBJECT"` is **not** redirected and `gh pr edit` prints the edited PR's URL to stdout (the same URL-to-stdout convention the create path already guards against), immediately before `echo "$EXISTING_PR"`.

The captured `$STATE_PR` therefore becomes a multiline blob rather than a bare number; in an unattended run the downstream `--argjson state_pr` fails and the status report never renders or posts. Observed live this session on the create path — the number had to be extracted by hand; the reuse path fails identically on the next same-day run.

**3. Stale per-run scratch is never cleaned (robustness).** `/devflow:retrospective-weekly` Step 1 (scratch prep) runs `mkdir -p .devflow/tmp` and `rm -f .devflow/tmp/new-entries.jsonl` only. The per-PR artifacts `result-*.json` (Stage-A outputs) and `pr-*.context.json` (fetched context bundles) are never removed, so they accumulate across every run without bound (measured after this run: 131 `result-*.json` + 135 `pr-*.context.json`). Two consequences: unbounded growth makes any "count the result files" completion check unreliable, and a stale `pr-*.context.json` left at a PR's path can be read as current if that PR's fetch fails on a later run — silently analyzing another run's data.

## Desired Behavior

**1.** `cheap-gate.jq` treats a reflection bullet as **exempt only when it is a `note`-kind bullet** — rendered `- ℹ️ …` under `### ℹ️ Notes`. **Every other bullet is friction** and forces analysis: `issue-accuracy` (rendered `- 📝 **Issue accuracy:** …`, *also* under `### ℹ️ Notes` — actionable "wrong/underspecified issue claims" feedback, deliberately **not** exempt), every `### ⚠️ Action required` or `### 💡 Improvements` bullet, every bullet under an *unrecognized* `### ` heading (a future subsection, or an adversarially hand-edited workpad), and every bullet before any `### ` heading. This **exempt-list** direction (exempt exactly one kind; everything else is friction) means the taxonomy can gain a section or kind without silently making its bullets non-friction — anything not explicitly the informational `note` kind fails closed to friction. Because `note` and `issue-accuracy` share the `### ℹ️ Notes` section, the classifier distinguishes them within that section by their per-bullet glyph — `ℹ️` (note, exempt) vs `📝` (issue-accuracy, friction) — both from `scripts/workpad.py`'s `_REFLECTION_KINDS`. A PR whose reflection block contains only `note` (`ℹ️`) bullets, and is otherwise clean, is treated as clean and routed through the cheap `clean-entry` path instead of Stage A. Any non-exempt bullet — alone or mixed with `note` bullets — still forces analysis. When the new friction signal is **absent** from a bundle (an older bundle that predates this change, or a bundle whose emission failed), `cheap-gate.jq` falls back to its current behavior — trip on any non-empty `reflections[]` — so the change **fails closed** (a missing signal over-analyzes, never silently skips a friction PR), mirroring the existing `reflections // []` default the header already documents for older bundles. The `reflections[]` bundle field keeps its existing flat-string-array shape so every current consumer reads it unchanged.

So that an exempted `note` is not silently lost from the learnings, **`clean-entry.jq` records the bundle's reflection bullets in the clean entry** it emits (today it drops them — verified: it copies no reflection text). A PR routed to the clean path because its only reflections are `note` bullets keeps those bullets verbatim in its `retrospectives.jsonl` entry (recorded, just not LLM-analyzed), and its summary no longer claims "no retrospective signal" when a note is present.

**2.** `open-state-pr.sh` writes **only** the resulting PR number (or `DRYRUN` on the dry-run path, or the existing-PR number on the reuse path) to stdout on **every** real path; the `git checkout`, `git commit`, and reuse-path `gh pr edit` progress output is redirected off stdout. A `STATE_PR=$(bash open-state-pr.sh)` capture yields a single bare token on both the create path and the branch-reuse path.

**3.** `/devflow:retrospective-weekly` Step 1 removes prior-run `.devflow/tmp/result-*.json` and `.devflow/tmp/pr-*.context.json` before scanning, so a run starts from clean per-PR scratch and can never read another run's stale bundle or output. The removal is **shell-agnostic** — it is a safe no-op when no such files exist **and when only one of the two patterns matches** (the common first-of-week state, where `result-*.json` exists but no `pr-*.context.json` has been written yet), on both bash and zsh.

## User Impact

The maintainer (and the unattended cron variant) gets a cheaper weekly run (informational-only PRs stop consuming an LLM analysis each), a state-PR/report step that actually completes unattended instead of silently breaking on a polluted capture, and a scratch directory that stays bounded and cannot feed stale data into a run.

## Technical Context

**Scope note:** The files and details below are the known starting points, not the full list. Before implementing, trace the change through the codebase to find every affected call site, consumer, and layer — this issue maps the work, it does not bound it.

- **Relevant Classes/Files**
  - `lib/cheap-gate.jq` — the reflection check (final `elif`, currently `reflection_count > 0`) is the gate to change.
  - `lib/fetch-pr-context.sh` — the python reflection parser (the `## Devflow Reflection` extraction loop) that currently drops the `### ` sub-section headings; and the `jq --slurpfile` bundle assembly where a new friction-signal field would be emitted.
  - `scripts/workpad.py` — **the renderer that owns the section-heading literals** (`_REFLECTION_SUBSECTIONS`) AND the per-bullet kind→glyph map (`_REFLECTION_KINDS`: `note→ℹ️`, `issue-accuracy→📝`, …). The parser couples to both — the section heading (`### ℹ️ Notes`) and the note/issue-accuracy glyphs that distinguish exempt from friction *within* that section. This is the authoritative source of both, not `skills/implement/SKILL.md` (which only documents them).
  - `skills/implement/SKILL.md` — documents the reflection-kind → section contract (`### ⚠️ Action required` / `### 💡 Improvements` / `### ℹ️ Notes`) for humans; not the render source.
  - `lib/open-state-pr.sh` — the `_run git checkout -B`, `git commit -m`, and reuse-path `gh pr edit "$EXISTING_PR"` calls that leak to stdout, and the `gh pr create … >&2` call that already demonstrates the correct redirect.
  - `skills/retrospective-weekly/SKILL.md` — Step 1 (scratch prep), where the cleanup is added.
  - `lib/clean-entry.jq` — the cheap-path entry emitter; **edited to record the bundle's reflection bullets in the clean entry** (today it drops them, so an exempted `note` would vanish from learnings) and to not claim "no retrospective signal" when a note is present.
  - `skills/retrospective/SKILL.md` — the Stage-A subagent skill; its "Source priority" section documents an invariant ("a run that left even one reflection … the clean-gate already forces it into analysis") that this change falsifies for `ℹ️ Notes`-only runs — a coupled doc site to reconcile (via `superpowers:writing-skills`).
  - `lib/test/run.sh` — the suite where the new pins land.
- **Architecture Alignment** — Item 1 keeps the discrimination on the read side (`fetch-pr-context.sh` bundle assembly + `cheap-gate.jq`), mirroring how `workpad_final_status` is normalized once at extraction and consumed as a simple field by the gate. Item 2 mirrors the existing `gh pr create … >&2` redirect already in the same file. Item 3 mirrors the existing `rm -f .devflow/tmp/new-entries.jsonl` scratch-prep line.
- **Dependencies** — none beyond the existing `python3` (already a hard preflight prerequisite for `fetch-pr-context.sh`'s parser) and `jq` (resolved via the existing resolver).
- **Data/Schema Considerations** — Item 1 adds a **new** top-level bundle field (a friction-reflection count) and leaves the existing `reflections[]` string array byte-identical, so it is additive to the bundle schema; no existing consumer field changes.
- **Cross-layer Impact** — Entirely within the retrospective loop's shell/jq/skill layer. No frontend, API, or database surface.

Verified premises (read against HEAD while drafting):
- `cheap-gate.jq`'s reflection check is `($reflections // []) | length` compared `> 0` — a flat count, no severity awareness. (Verified: read `lib/cheap-gate.jq`.)
- `fetch-pr-context.sh` captures reflection bullets with `re.match(r'^\s*[-*]\s+(.*\S)\s*$', line)` and terminates the region at the next `## ` heading, so `### ` sub-headings are neither captured nor treated as terminators — they are silently skipped. (Verified: read `lib/fetch-pr-context.sh` reflection parser.)
- The section-heading literals and the per-bullet kind→glyph map are **rendered by `scripts/workpad.py`** — `_REFLECTION_SUBSECTIONS` = `{action: "### ⚠️ Action required", improvements: "### 💡 Improvements", notes: "### ℹ️ Notes"}`, and `_REFLECTION_KINDS` maps `blocked→⛔`, `deferred→⏭️`, `dropped-failed→❗` (all under `action`), `improvement→💡` (under `improvements`), `issue-accuracy→📝` and `note→ℹ️` (both under `notes`). **`⚠️` never appears on a bullet — only on the `action` section heading — and the informational `issue-accuracy` bullet carries `📝`, not `ℹ️`.** (Verified: read `scripts/workpad.py` `_REFLECTION_KINDS` / `_REFLECTION_SUBSECTIONS`.)
- `open-state-pr.sh` redirects `gh pr create` to stderr but not `git checkout -B`, `git commit -m`, or the reuse-path `gh pr edit "$EXISTING_PR"`; the git commands print to stdout by default (`git commit` prints its summary; `git checkout -B` prints carried-over `M<TAB>file` lines when the working tree has modified tracked files) and `gh pr edit` prints the PR URL to stdout like `gh pr create`. (Verified: read `lib/open-state-pr.sh` create and reuse paths; reproduced the create-path leak live — the state-PR capture returned a multiline blob.)
- `retrospective-weekly` Step 1 cleans only `new-entries.jsonl`. (Verified: read the skill's Step 1.)

## Acceptance Criteria

- [ ] **(Item 1, exempt arm)** A context bundle whose `## Devflow Reflection` block contains only `note`-kind bullets (rendered `- ℹ️ …` under `### ℹ️ Notes`), with every other cheap-gate signal clean, yields `{"clean": true}` from `cheap-gate.jq`.
- [ ] **(Item 1, issue-accuracy is friction)** A context bundle whose reflection block contains an `issue-accuracy` bullet (rendered `- 📝 **Issue accuracy:** …` under `### ℹ️ Notes`) — even if it is the *only* reflection and every other signal is clean — yields `{"clean": false}`. `issue-accuracy` is actionable feedback, not an exempt note, despite sharing the `### ℹ️ Notes` section with `note`.
- [ ] **(Item 1, trip arm)** A context bundle containing at least one `### ⚠️ Action required` bullet (kind `blocked`/`deferred`/`dropped-failed`) or one `### 💡 Improvements` bullet, otherwise clean, yields `{"clean": false}` with a reason naming the friction reflection.
- [ ] **(Item 1, mixed arm)** A context bundle whose reflection block mixes a `note` (`ℹ️`) bullet with an `⚠️`, `💡`, or `📝` bullet yields `{"clean": false}` — any non-exempt bullet forces analysis.
- [ ] **(Item 1, unknown-heading arm)** A context bundle with a reflection bullet under an **unrecognized** `### ` heading (a heading that is not one of the three current sections), otherwise clean, yields `{"clean": false}` — an unrecognized section fails closed to friction, so a future taxonomy addition cannot silently make its bullets non-friction.
- [ ] **(Item 1, fail-closed on absent signal)** A bundle that carries a non-empty `reflections[]` but **lacks** the new friction signal (an older-format bundle, or one whose emission failed) yields `{"clean": false}` — `cheap-gate.jq` falls back to the legacy "any reflection trips" behavior rather than treating the absent signal as zero friction. A missing signal never routes a friction-bearing PR to the clean path.
- [ ] **(Item 1, compatibility)** The emitted bundle's `reflections[]` field keeps its existing flat-string-array shape and contents; the Stage-A retrospective subagent and any other `reflections[]` consumer read it unchanged.
- [ ] **(Item 1, clean-entry preserves notes)** `clean-entry.jq` records the bundle's reflection bullets in the clean retrospective entry it emits (today it drops them), so a PR routed to the clean path on `note`-only reflections keeps those bullets verbatim in its `retrospectives.jsonl` entry, and its summary does not claim "no retrospective signal" when a note is present. The entry stays schema-valid for every `retrospectives.jsonl` consumer.
- [ ] **(Item 1, coupled doc)** `skills/retrospective/SKILL.md`'s "Source priority" invariant — currently "a run that left even one reflection is, by construction, not a frictionless run (the clean-gate already forces it into analysis for exactly this reason)" — is reconciled with the new gate (a run whose *only* reflections are `note`-kind bullets is no longer forced into analysis; every actionable kind, including `issue-accuracy`, still is), edited via the `superpowers:writing-skills` discipline.
- [ ] **(Item 2)** On **every** real (non-dry-run) path — the create path and the branch-reuse path — `lib/open-state-pr.sh` writes only the resulting PR number to stdout (and `DRYRUN` on the dry-run path); the `git checkout`, `git commit`, and reuse-path `gh pr edit` progress output goes to stderr. A `X=$(bash lib/open-state-pr.sh …)` capture yields a single bare token with no git or `gh` porcelain, on both a first run (create) and a same-day re-run (reuse).
- [ ] **(Item 3)** `/devflow:retrospective-weekly` Step 1 removes prior-run `.devflow/tmp/result-*.json` and `.devflow/tmp/pr-*.context.json` before the scan; the removal is a safe no-op when **neither or only one** pattern matches (no fatal unmatched-glob abort), works on both bash and zsh, and never touches the current run's own freshly-written files (Step 1 runs before any fetch).
- [ ] **(Tests)** `lib/test/run.sh` gains, at minimum: cheap-gate fixtures pinning the item-1 exempt (`note`-only → clean) / **issue-accuracy-is-friction** / trip / mixed / **absent-signal-fail-closed** arms; a `clean-entry.jq` fixture asserting a `note`-only bundle's reflection text is recorded in the entry; an `open-state-pr.sh` bare-PR-number test on **both** the create path and the branch-reuse path (reproducing each leak first); a coupled-site pin asserting the parser's heading + note/issue-accuracy glyph literals match `scripts/workpad.py`'s `_REFLECTION_SUBSECTIONS` and `_REFLECTION_KINDS`; and a coupled-site pin on the retrospective-weekly Step-1 cleanup literal. The item-1 and item-2 tests fail against today's code (reproducing each defect) before the fix and pass after. (The Testing Strategy below sweeps additional arms — empty and malformed/sectionless reflection blocks, and an end-to-end real-workpad-body fixture — beyond this floor.)

## Implementation Notes

**Approach.**
- *Item 1* — In `fetch-pr-context.sh`'s reflection parser, track the current `### ` sub-section **heading** while iterating (the loop already goes line by line), and classify each captured bullet with **exempt-list** semantics — exempt exactly one kind, `note`; **every other bullet is friction**. A bullet is exempt only when **both** hold: it is under `### ℹ️ Notes` **and** it is a `note` bullet (leading glyph `ℹ️`), not an `issue-accuracy` bullet (leading glyph `📝`, which shares that section). Everything else is friction — `### ⚠️ Action required`, `### 💡 Improvements`, an `issue-accuracy` (`📝`) bullet, any *unrecognized* `### ` heading, and any bullet before a `### ` heading. This default-friction direction is what makes a future taxonomy section or kind fail closed instead of silently non-friction. **Do the section split by heading, not by hunting for a section glyph on the bullet** — `⚠️` never appears on a bullet, so an action bullet is recognized by its `### ⚠️ Action required` *heading*, not a per-bullet `⚠️`. The **only** per-bullet glyph the classifier reads is the note-vs-issue-accuracy distinction *within* `### ℹ️ Notes` (`ℹ️` = exempt note, `📝` = friction issue-accuracy). The parser is an inline `python3` heredoc that **cannot `import` `workpad.py`**, so the `### ℹ️ Notes` heading literal and the `ℹ️`/`📝` glyphs are hard-copied into the heredoc **and** pinned by a `run.sh` coupled-site assertion (source↔source) against `_REFLECTION_SUBSECTIONS` / `_REFLECTION_KINDS` — this **relocates** the drift risk to a source pin, it does not eliminate an end-to-end normalization divergence, so also keep an end-to-end fixture that feeds a **real rendered workpad body** and asserts the friction field. Emit a **new** top-level bundle field — a friction-reflection count — alongside the unchanged `reflections[]` string array. In `cheap-gate.jq`, change the final reflection `elif` to gate on that new friction field, and refine the `reason` string to name friction reflections; **when the field is absent, fall back to the current `reflections | length > 0` trip** (fail-closed for older/failed-emission bundles, mirroring the existing `reflections // []` default).
- *Item 1 (clean-entry)* — In `clean-entry.jq`, add the bundle's `reflections` to the emitted clean entry (a `reflections` field carrying the verbatim bullets) and make the hardcoded `summary` acknowledge them when present, so an exempted `note` is recorded in `retrospectives.jsonl` rather than dropped. Keep the entry shape a superset of today's (additive field), so existing `retrospectives.jsonl` consumers are unaffected.
- *Item 2* — Redirect the three leaking commands' stdout to stderr: `git checkout -B … 1>&2` (through `_run`), `git commit -m … 1>&2`, and the reuse-path `gh pr edit "$EXISTING_PR" --title "$SUBJECT" 1>&2`. Stderr (not `/dev/null`) preserves the output for debugging while keeping stdout clean, matching the intent of the existing `gh pr create … >&2` line.
- *Item 3* — In `retrospective-weekly` Step 1, remove the prior-run per-PR scratch next to the existing `rm -f .devflow/tmp/new-entries.jsonl`, using a **shell-agnostic** form that cannot fatally abort on an unmatched pattern — a single `find .devflow/tmp -maxdepth 1 -type f \( -name 'result-*.json' -o -name 'pr-*.context.json' \) -delete 2>/dev/null` (BSD/macOS-safe, no GNU-only flags). A bare `rm -f <glob>` is **not** safe here: under zsh (the local Bash tool in at least one supported environment) an unmatched glob is a fatal `no matches found` that aborts the whole `rm`, so if `result-*` matches but `pr-*` does not, *neither* is removed. This is a `SKILL.md` body edit, so make it through the `superpowers:writing-skills` discipline per the repo convention.

**Code Patterns.**
- Item 1's "normalize once at extraction, consume a simple field at the gate" mirrors `WORKPAD_FINAL_STATUS` handling in `fetch-pr-context.sh` + `workpad_ok` in `cheap-gate.jq`.
- Item 2 mirrors the existing `gh pr create … >&2` redirect in the same file.
- Item 3 mirrors the existing scratch-prep `rm -f` line; new `lib/test/run.sh` pins follow the repo's `assert_pin_red_under` / coupled-site pin conventions.

**Testing Strategy.**
- *Boundary & level (Move 1)* — `cheap-gate.jq` is a pure jq filter with a clear bundle-in → `{clean,reason}`-out contract: unit-test it with crafted context-bundle fixtures. `fetch-pr-context.sh`'s classification is exercised end-to-end by feeding a workpad body containing each section and asserting the emitted friction field. `open-state-pr.sh`'s stdout is testable in a throwaway git repo with the existing `gh` stub (the suite already builds temp git repos and stubs `gh`). `retrospective-weekly` Step 1 is a skill-prose change whose automated boundary is the coupled-site pin on the cleanup literal.
- *Item 1 cases* — fixtures build workpad bodies with the **real** glyphs under the real section headings (`### ℹ️ Notes` with an `ℹ️` note *and* a `📝 **Issue accuracy:**` bullet; `### ⚠️ Action required` with a `⛔`/`⏭️`/`❗` bullet; `### 💡 Improvements` with a `💡` bullet). Arms: exempt (`ℹ️` note-only → clean), **issue-accuracy-is-friction** (`📝`-only → not clean — guards the note-vs-issue-accuracy split), trip (a single action-required bullet, and separately a single `💡` → not clean), mixed (`ℹ️` note + an action bullet → not clean; and `ℹ️` note + `📝` → not clean), **unknown-heading** (a bullet under a `### 🔥 Something` heading that is not one of the three → fail-closed to not clean), empty reflection block (→ clean, unchanged), malformed/sectionless block (bullets before any `### ` heading → fail-closed to not clean), and **absent-signal** (a bundle with non-empty `reflections[]` but no friction field → fail-closed to not clean). Assert both the `clean` boolean and, for the trip arms, the `reason`. Include an **end-to-end** fixture: a real rendered workpad body (from `workpad.py`) → `fetch-pr-context.sh` → the friction field, so a normalization divergence the source pin can't see is still caught.
- *Item 1 compatibility & coupling* — a fixture bundle asserts `reflections[]` is byte-identical to the pre-change extraction for the same workpad body. A coupled-site pin asserts the parser's `### ℹ️ Notes` heading literal and the `ℹ️`/`📝` glyphs equal `scripts/workpad.py`'s `_REFLECTION_SUBSECTIONS` / `_REFLECTION_KINDS` values, so a future rename in `workpad.py` turns the pin RED rather than silently mis-bucketing.
- *Item 1 clean-entry* — a fixture asserts `clean-entry.jq` emits an entry that (a) is schema-valid and (b) **contains the `note` bullet text** for a `note`-only bundle, and that its summary does not say "no retrospective signal" when a note is present. Reproduce-first: against today's `clean-entry.jq` the note text is absent from the entry.
- *Item 2 (bug-fix, reproduce first, both real paths)* — in a temp git repo with a modified tracked file staged and the `gh` stub returning a known PR number, run `open-state-pr.sh` and assert **stdout is exactly that number** — once with **no** existing PR (create path) and once with the stub reporting an existing PR for the branch (reuse path, exercising `gh pr edit`). Against today's code each capture contains the git / `gh pr edit` porcelain, so both tests fail first exhibiting the exact leak, then pass after the redirects.
- *Item 3* — the reproducible verification: run the loop's Step-1 prep twice with a planted stale `result-999.json` / `pr-999.context.json`, and assert those files are gone after the second Step 1; plus the coupled-site pin that the Step-1 block names both globs.
- Every AC above maps to at least one of these assertions; no orphan assertions.

**Documentation Needed.** Update `cheap-gate.jq`'s header comment (the reflection-check description and the `clean` bullet list) to state the severity discrimination. Note the new bundle field wherever the context-bundle schema is documented (`docs/` if the bundle schema is described there; otherwise the `fetch-pr-context.sh` header). Update the `retrospective-weekly` Step 1 description to mention the scratch cleanup.

**Potential Gotchas.**
- Item 1: **use exempt-list, not whitelist, classification.** Exempt = exactly the `note` kind; friction = everything else (issue-accuracy, Action-required, Improvements, unknown/future `### ` headings, sectionless bullets). A friction *whitelist* would fail **open** on a future fourth section — `scripts/workpad.py` already tolerates an unknown `### ` heading — silently routing those bullets to the clean path.
- Item 1: **`note` and `issue-accuracy` share the `### ℹ️ Notes` section** — the section heading alone cannot separate them. Split them by the per-bullet glyph *within* that section: `ℹ️` = note (exempt), `📝` = issue-accuracy (friction). Do **not** exempt the whole `### ℹ️ Notes` section (that would wrongly exempt actionable issue-accuracy feedback). Separately, do **not** try to recognize Action-required bullets by a per-bullet `⚠️` — `⚠️` appears only on the section heading; recognize the section by its heading.
- Item 1: the `### ℹ️ Notes` heading literal **and** the `ℹ️`/`📝` glyphs are a **coupled invariant with `scripts/workpad.py`** (`_REFLECTION_SUBSECTIONS` + `_REFLECTION_KINDS`, the render source) — and the inline `python3` heredoc parser can't `import` them, so they are hard-copied. The `run.sh` coupling pin (source↔source) catches a rename there; it does **not** catch an end-to-end emoji-normalization divergence (e.g. a `U+FE0F` variation selector dropped somewhere in the `gh api`→jq→heredoc pipeline) — the end-to-end real-workpad-body fixture is what guards that. The parser fail-closes to friction on any bullet it can't positively match as a `note`, so a drift over-analyzes rather than silently skips.
- Item 1: `cheap-gate.jq` must **fail closed** when the new friction field is absent (older bundle, or a bundle whose emission failed) — fall back to the legacy `reflections | length > 0` trip. Gating on a bare `.friction_count > 0` would make an absent field read as zero friction (`null > 0 → false`) and silently route a friction PR to the clean path.
- Item 1: bullets emitted before any `### ` heading (a malformed block) must fail-closed to friction — never silently exempt a real problem. The existing malformed-block guard in `fetch-pr-context.sh` (parse failure → `[]`) must keep producing a clean-gate-safe result.
- Item 1: PRs that used to reach Stage A (`note`-only) now flow through `clean-entry.jq`. Today that path **drops all reflection text** (verified: `clean-entry.jq` copies none), so without the clean-entry edit an exempted `note` would vanish from `retrospectives.jsonl` and its summary would falsely read "no retrospective signal." The clean-entry edit (record the bullets + fix the summary) is **in scope**, not optional — it is what keeps the exemption a cost saving rather than a signal loss.
- Item 1: the exemption trades away analysis of *worked-around informational deviations* only. `issue-accuracy` (wrong/underspecified issue claims) and every actionable kind still force analysis by design — do not widen the exemption to them.
- Item 2: redirect the three leaking commands (`git checkout -B`, `git commit`, reuse-path `gh pr edit`); do not touch the `gh pr list` (captured into a variable) or `gh pr create` (already `>&2`) calls. The dry-run path's `DRYRUN:` diagnostics are intentional and out of scope.
- Item 3: **do not use a bare `rm -f <glob>`** — under zsh an unmatched glob is a fatal `no matches found` (verified: `zsh -c 'rm -f x/result-*.json'` → rc 1), and a two-pattern `rm` aborts entirely if only one matches, so *neither* pattern is cleaned in the common first-of-week state. Use the `find … -delete` form (verified no-op, rc 0, on no match), which is shell- and OS-agnostic.

The01Geek:
## Summary
- Stops the weekly retrospective loop from spending an LLM analysis on PRs whose only signal is an informational `note`-kind reflection, by teaching `cheap-gate.jq` to discriminate reflection *severity*.
- Fixes `open-state-pr.sh` leaking git/`gh` porcelain to stdout, which broke the unattended state-PR/report step's `STATE_PR=$(...)` capture.
- Bounds retrospective scratch growth by cleaning prior-run per-PR bundles before each scan.

## Changes
**Reflection-severity gate (`lib/cheap-gate.jq`, `lib/fetch-pr-context.sh`)**: The reflection parser now emits a new top-level `reflections_friction_count` bundle field alongside the byte-identical `reflections[]`. A bullet is exempt (non-friction) only when it is a `note`-kind `ℹ️` bullet under `### ℹ️ Notes`; every other kind — `issue-accuracy` `📝` (which shares that section), Action-required, Improvements, any unrecognized/missing `### ` heading, and any pre-heading bullet — fails closed to friction. `cheap-gate.jq` gates on the friction count, and **fails closed** to the legacy "any reflection trips" behavior when the field is absent (older/failed-emission bundle). A present-but-unparseable reflection block is substituted with a friction sentinel so it over-analyzes rather than silently reading as clean.

**Note preservation (`lib/clean-entry.jq`)**: A PR routed to the clean path on `note`-only reflections now records those bullets verbatim in its `retrospectives.jsonl` entry (previously dropped), and its summary no longer claims "no retrospective signal" when a note is present.

**State-PR stdout hygiene (`lib/open-state-pr.sh`)**: `git checkout -B`, `git commit`, `git push -u`, and the reuse-path `gh pr edit` progress output are redirected to stderr, so stdout carries only the resulting PR number on both the create and branch-reuse paths.

**Stale-scratch cleanup (`skills/retrospective-weekly/SKILL.md`)**: Step 1 removes prior-run `result-*.json` / `pr-*.context.json` via a shell-agnostic `find … -delete` (no fatal unmatched-glob abort under zsh), before any fetch.

**Coupled doc (`skills/retrospective/SKILL.md`)**: The "Source priority" invariant is reconciled with the new gate — a run whose only reflections are `note`-kind bullets is no longer forced into analysis; every actionable kind still is.

**Docs & tests**: `docs/DEVFLOW_SYSTEM_OVERVIEW.md`, `docs/implement-skill.md`, and `docs/external/release-notes.md` document the new field and exemption; `lib/test/run.sh` gains cheap-gate arms, an end-to-end real-rendered-workpad friction fixture, a source↔source coupled-site pin, parser fail-closed assertions, a parse-failure friction-sentinel pin, clean-entry note-preservation, and an `open-state-pr` bare-PR-number test on both paths.

## Resolves
Resolves #519

## Test Plan
- [ ] `bash lib/test/run.sh` passes (cheap-gate/parser/clean-entry/open-state-pr arms, coupled pins).
- [ ] `git ls-files '*.sh' | grep -v '^lib/test/' | xargs -r shellcheck --severity=warning -e SC1091` is clean.
- [ ] `git ls-files '*.py' | xargs -r ruff check` is clean.
- [ ] A note-only reflection bundle yields `{"clean":true}` from `cheap-gate.jq`; an `issue-accuracy`/action bundle yields `{"clean":false}`.
- [ ] `STATE_PR=$(bash lib/open-state-pr.sh …)` yields a single bare PR number on create and same-day reuse.

## Visual Changes
N/A

## Breaking Changes
None. The `reflections[]` bundle field keeps its existing flat-string-array shape; the new `reflections_friction_count` field is additive and the gate fails closed when it is absent, so older bundles behave as before.


The01Geek:
### Receiving-code-review — deferral record (iteration 2)

Blocker fixed; the remaining findings are all below the `critical` re-open threshold and are recorded here (not fixed this pass), with rationale:

- **🟠 Important #1 (`docs/implement-skill.md:504`)** — **FIXED.** Replaced "irrelevant to `cheap-gate.jq`'s non-empty check" with "…friction check". Verified against HEAD: `lib/cheap-gate.jq:81` gates on `$friction_count > 0`, not a non-empty check.
- **🟡 Suggestion #1 (`capture-implement-session.py` dirty_tree)** — **Reviewer's proposed fix rejected; deliberate design documented at the site instead.** `run_git` returns `value or None`, so a clean tree (empty `git status` output) and a failed `git status` are *both* `None`. The `and head_sha is None` guard is the intended disambiguator (HEAD resolves ⇒ git works ⇒ `None` status means clean). The suggested `None if status is None` would misreport **every clean tree** as `unknown` — a regression. Added an explanatory comment; deeper `run_git` empty-vs-failure redesign deferred (dev-only, fail-benign metadata). **Revisit if:** the capture bundle ever needs to distinguish clean-tree from status-failure.
- **🟡 Suggestion #2 (`analyze-implement-runs.py` type-design)** — Deferred. Latent `TypeError` is only "if ever re-serialized"; no current failing path. Dataclass/`schema_version` gate is a design improvement. **Revisit if:** bundle metadata is ever re-serialized or the schema version bumps.
- **🟡 Suggestion #3 (`analyze-implement-runs.py` `--safe-mode` CLI flags)** — Deferred; not verifiable from the diff (suite uses a fake `claude`). Dev-only, manual analyzer. **Revisit with:** a real-CLI smoke check (highest-value follow-up).
- **🟡 Suggestion #4 (`COHORT_SIZE` constant)** — Deferred; pure drift-prevention refactor, no behavioral defect.

Verification: `lib/test/run.sh` → 8039 passed, 0 failed, 1 skipped (the `#434` self-scan self-skips on a dirty working tree; runs live in CI). `ruff` clean.

_Note for maintainer:_ commit `f04c5a27` ("feat: add implement flight recorder") is in this PR's diff but is not in `origin/main`, so the flight-recorder files appear in the #519 diff. Structural/scope decision left to you — not addressed here.

The01Geek:
### Update — all review findings now addressed (iteration 3)

Per request, the previously-deferred Suggestions are now **fixed** (not just deferred):

- **🟡 Suggestion #1 (`capture-implement-session.py` `dirty_tree`)** — Fixed with a dedicated `git_status_dirty()` that reads git's own exit status, so clean / dirty / failed-`git status` are now distinct (`False` / `True` / `None`). The old `head_sha` proxy misreported a clean **no-commit** repo as `unknown`; that discriminator case is now tested and mutation-checked.
- **🟡 Suggestion #2 (`analyze-implement-runs.py` type-design)** — Fixed. Frozen `Bundle` / `IssueDraft` dataclasses replace the stringly-typed dict/tuple; the parsed capture timestamp is a typed field instead of being injected into the JSON dict (removing the latent `TypeError`-on-reserialize); and a **`schema_version` gate** now rejects unsupported bundles rather than misreading them (tested with a rejection + attribution pair, mutation-checked).
- **🟡 Suggestion #3 (`--safe-mode` / `--permission-mode dontAsk` / `--allowedTools` / `--print`)** — **Verified** against `claude` CLI **2.1.211** (`--help` lists all four; `dontAsk` is a valid `--permission-mode` choice). No code change — the suite uses a fake `claude` binary by design and cannot drive the real CLI. Disposition recorded here rather than a smoke test.
- **🟡 Suggestion #4 (`COHORT_SIZE` constant)** — Fixed; the two hard-coded `3`s in `--last` handling now reference `COHORT_SIZE`.

Verification: `lib/test/run.sh` → **8046 passed, 0 failed, 0 skipped** on clean committed HEAD (the `#434` stale-prose self-scan ran green); `ruff` clean. Both new guard tests mutation-checked (observed RED for the pinned reason; working tree never mutated).

PR #526, opened 2026-07-16
feat: workflow flight recorder and native transcript instrumentation

The01Geek:
Bundles the commits that had accumulated on local `main` into a reviewable branch.

## What's here

**Recorder core**
- `scripts/workflow_flight_recorder.py` — the recorder itself
- `scripts/capture-workflow-manifest.py` — captures workflow start manifests
- `scripts/inventory-workflow-transcripts.py` / `import-workflow-transcript.py` — native transcript inventory + import
- `scripts/analyze-workflow-runs.py` — performance-cohort analysis
- `scripts/workflow-flight-recorder-registry.json` + two analysis prompt files

**Tests**
- `lib/test/test_workflow_flight_recorder.py`, `lib/test/test_workflow_analyzer.py`
- `lib/test/run.sh` extended to drive them

**Docs**
- `docs/workflow-flight-recorder.md`
- `docs/DEVFLOW_SYSTEM_OVERVIEW.md` updated
- `.changeset/native-workflow-transcript-instrumentation.md`

**Touched existing**
- `scripts/analyze-implement-runs.py`, `scripts/capture-implement-session.py`
- `.claude/settings.json`

Versioning follows the changeset convention: no direct edits to `.claude-plugin/plugin.json` or `CHANGELOG.md` — `version-consolidate` owns those on merge.

## Note for reviewers

This branch is currently behind `origin/main` and needs updating before CI is meaningful. `lib/test/run.sh` is modified on both sides, so expect a conflict there.


The01Geek:
## Review findings addressed

All nine findings triaged against HEAD. Suite: **8153 passed, 0 failed, 0 skipped** (clean tree, so the `#434` stale-prose blocking gate armed and ran rather than self-skipping). `ruff` + `shellcheck` clean.

### Fixed

**Important-1 (REJECT driver)** — confirmed: `capture_prompt_manifest` took `requested_model`/`requested_effort` from the payload alone, so the documented launch-line declaration was silently dropped on the hooked path. Fixed in the code rather than by retreating in the docs, since `DEVFLOW_RECORDER_OUTPUT_STYLE` already survives the launch and leaving its siblings import-only would ship an inconsistent user-facing contract. A declaration now fills what the session did not report; an observation from the host still wins and is sourced `user_prompt_submit_payload` vs `explicit_recorder_environment`. Doc updated to state that precedence exactly.

**Important-2** — SPDX headers added to all six files (all nine new sources now carry one).

**Important-3** — `_discover()` now names the bundle it drops instead of a bare `continue`.

**Suggestion-1** — `absent` vs `corrupted` vs `present` is now carried on `SessionBundle`, surfaced per-occurrence in the cohort document, and the analyst prompt is told not to read the two as the same thing.

**Suggestion-2** — analyst subprocess bounded at 900s, `DEVFLOW_CLAUDE_TIMEOUT`-overridable; a malformed/non-positive override degrades to the default with a breadcrumb rather than disabling the guard.

**Suggestion-3** — `os.execv` failure reported in the `devflow: <tool>: <reason>` convention.

**Suggestion-4** — intent confirmed and documented: a manifest is captured at UserPromptSubmit when only the root occurrence exists, so it can only fingerprint that workflow's surfaces.

**Suggestion-5** — added: Stop-observer forced-exception fail-open, inventory `OSError`, the recurrence threshold at its exact boundary, non-object/empty transcripts, naive-datetime rejection, and the `_analyst_timeout` shape matrix. The brittle cumulative post-loop assertion in the entry-point test is replaced with per-case attribution plus a positive control, so a rejection from an unrelated precondition can no longer masquerade as the one under test.

**Suggestion-6 (partial)** — `mode`/`boundary_confidence` are now `Literal`; the value sets were verified to match every literal assigned or compared in the module.

### Deferred, with reasons

- **`Occurrence` → frozen.** Recorded as a comment at the dataclass. Boundary resolution discovers end/timing/model-effort facts across three later passes over the event stream; rebuilding a frozen instance per pass would spread the correlated-field updates wider, not narrower, for immutability the single producer does not currently need. Revisit if a second producer appears.
- **`SessionBundle.metadata`/`event_summary` defensive copies.** Both producers build these from a fresh `json.loads` per bundle, so no aliasing exists today; a shallow copy would not protect the nested contents anyway. Revisit if a caller starts mutating them.

### Correction to my own fix

The first pass of the doc edit asserted both sources carry `explicit_recorder_environment` — false, an observed value carries `user_prompt_submit_payload`. Caught during the diff-claim check and corrected in 66c2c3e2.

Every new test was mutation-checked: each was observed RED for the reason it pins (e.g. reverting the env wiring → `AssertionError: None != 'declared-model'`; `>= 2` → `>= 3` on the recurrence threshold → `False is not true`), and each mutation was reverted with the file verified byte-identical afterward.

Issue #530, opened 2026-07-16
[Wave 2] Split Review-and-Fix into durable step references


The01Geek:
## Problem Statement

Review-and-Fix loads its schema, loop, shadow, fixing, convergence, and Loop Exit procedures before the active step. Durable state exists, but procedure lacks a reloadable boundary after compaction and resume.

## Current Behavior

`skills/review-and-fix/SKILL.md` plus its repo extension measure 1,330 lines, 38,634 words, 266,868 bytes, and approximately 66,717 tokens using bytes divided by four. The monolith owns the full lifecycle. `iter-<N>.json` persists execution state; `sweep_defs_read` records one authoritative external read.

## Desired Behavior

A thin root owns invocation, schema, loop state, routing, terminal mapping, and fail-closed reference handling. Step procedures load at entry and re-anchor after nested returns. Named run-scoped operands support resume without recall. The plugin root is at most 3,000 words; this repository's root plus current live extension is at most 5,500 words. Maximum root-plus-live-extension-plus-active-step load is at most 15,000 words in this repository. Consumer extension size is reported separately. The full lifecycle stays present.

## User Impact

Maintainers gain context headroom while consumers retain outcomes and telemetry.

## Technical Context

> **Scope note:** The files and details below are the known starting points, not the full
> list. Before implementing, trace the change through the codebase to find every affected
> call site, consumer, and layer — this issue maps the work, it does not bound it.

- **Relevant Classes/Files:** `skills/review-and-fix/SKILL.md`, its repo extension, Review, receiving-code-review, Implement phases 2 and 3, efficiency-trace files, `scripts/workflow-flight-recorder-registry.json`, recorder import/tests, `docs/workflow-flight-recorder.md`, `lib/test/run.sh`, and writable workflows.
- **Architecture Alignment:** extend `sweep_defs_read`: procedure comes from the executing bundle; evidence comes from the run-scoped workpad.
- **Dependencies:** use shared Review, Write-tool workpads, telemetry, base checkpoints, and existing grants. No new runtime helper is introduced; grants stay unchanged.
[...]

PR #625, opened 2026-07-20
[Wave 2 follow-up] Reconcile sweep_defs_read/sweep_evidence and add the reference_reads evidence field

### LINKED ISSUE #541: [Wave 2 follow-up] Reconcile sweep_defs_read/sweep_evidence and add the reference_reads evidence field
## Dependencies
Blocked by #530 — the parent issue's /devflow:implement run (the review-and-fix thin-root split) must land before this deferred schema-reconciliation chunk can start.

## Problem Statement
Follow-up to #530. The #530 split delivered the structural thin-root + step-references restructure and the durable continuation operands (`current_step`/`current_substep`/`pending_dispatch`), but deferred the run-scoped **evidence-schema** reconciliation to keep the split PR reviewable. This chunk reconciles the existing sweep-evidence fields and formalizes the new reference-read evidence field.

## Current Behavior
`sweep_defs_read` and `sweep_evidence` are pure orchestrator-prose fields written into `iter-<N>.json` by `skills/review-and-fix/references/fixing.md` (Step 3 items 3b/6/7). They are pinned only by the #478 assertions in `lib/test/run.sh`; there is no separate writer script and no `efficiency-trace.jq` consumer, and they are absent from the root `### Schema` JSON block. `ITER_EXPECTED_FIELDS` (`lib/efficiency-trace.sh`) exists (13 fields) but does not list them. The #530 fail-closed fix-delta outcome is delivered behaviorally (not_verified + prohibits clean approve) without a formal `reference_reads.fix_delta` schema field.

## Desired Behavior
The `sweep_defs_read`/`sweep_evidence` producers and consumers are reconciled across the authoritative writer prose, the root `### Schema` block, `ITER_EXPECTED_FIELDS`, and the tests, and the new `reference_reads` reference-read evidence field is added on top of that reconciled base with explicit `unknown/unrecoverable` provenance on synthesized records.

## User Impact
Maintainers and the cross-run efficiency analyzer get a single, consistent run-scoped evidence schema; fix-commit-only synthesized records never masquerade absent evidence as real.

## Technical Context
- **Relevant Classes/Files** — `skills/review-and-fix/references/fixing.md`, the root `### Schema` in `skills/review-and-fix/SKILL.md`, `lib/efficiency-trace.sh` (`ITER_EXPECTED_FIELDS`), `lib/efficiency-trace.jq`, `lib/test/run.sh` (#478 pins + the LR_SCHEMA divergence guard).
- **Architecture Alignment** — extend the existing evidence-schema mirror-site discipline; procedure comes from the executing bundle, evidence from the run-scoped record.
- **Cross-layer Impact** — the iter-<N>.json writer prose, the schema example, the single-source field set, and the desk pins.
- **Issue-accuracy note carried from #530** — the parent's premise that these fields reconcile across "the authoritative writer, top schema, ITER_EXPECTED_FIELDS, JQ consumer" over-states the current surface: `ITER_EXPECTED_FIELDS` exists but there is no separate writer script and no JQ consumer of these fields today. Scope the reconciliation to the surfaces that actually exist.
- Parent issue #530 was scoped to a single PR by its /devflow:implement run; see the workpad on #530 for the full scope decision.

## Acceptance Criteria
- [ ] Existing `sweep_defs_read` and `sweep_evidence` producers and consumers are reconciled before the new reference evidence is added. Fix-commit-only synthesized records persist explicit `unknown/unrecoverable` provenance; they never serialize absent evidence as `[]`/`not-run`. Producer-consumer tests cover each evidence field.

## Implementation Notes
- **Approach** — first reconcile the existing evidence schema across the authoritative writer, root schema, `ITER_EXPECTED_FIELDS`, and tests; then add the `reference_reads` field on the reconciled base.
- **Code Patterns** — mirror `sweep_defs_read`, the LR_SCHEMA divergence guard, and the fix-commit-fused write.
- **Potential Gotchas** — schema additions require complete producer-consumer updates; synthesized records mark fields `unknown/unrecoverable` instead of inventing no-sweep evidence.

The01Geek:
## Summary
- Reconciles the run-scoped evidence schema so the `iter-<N>.json` contract agrees across all four surfaces that define it, and adds the conditional `reference_reads` field that formalizes the fix-delta gate's outcome (issue #530 shipped that outcome behaviorally, with no durable record).
- Makes fix-commit-only synthesized records stamp explicit `{"status":"unrecoverable","reason":…}` provenance instead of `[]` / `{"status":"not-run"}` — those are the *legitimate* values of a real no-fix iteration, so emitting them here would assert something about an iteration the synthesis floor never observed.
- Adds a consumer guard for the value, not just the key: `--self-check` previously compared field *names*, so a regression to a real-looking `[]` kept every key present and validated in silence.

## Changes

**Schema (`skills/review-and-fix/SKILL.md`)**: the root `### Schema` block gains `sweep_defs_read` and `sweep_evidence` as unconditional fields and `reference_reads` as a conditional one (`{"fix_delta": {status, outcome, reason}}`), positioned like `shadow`.

**Authoritative writer prose (`references/fixing.md`, `references/fix-delta-gate.md`)**: item 7 gains the `reference_reads` specification — conditional contract, registry keying, `status`/`outcome`/`reason` semantics with both `not_verified` arms named, and the APPROVE-family prohibition; Step 3.5 gains the matching producer obligation at the gate it governs.

**Field sets and synthesis (`lib/efficiency-trace.sh`)**: `ITER_EXPECTED_FIELDS` gains the two sweep fields; `ITER_SYNTH_EXPECTED_FIELDS` gains all three; `synthesize_iter_workpads` stamps unrecoverable provenance via a single `unrec()` emitter; `do_self_check` gains a value-shape arm with every deref type-guarded and an else arm so a check that cannot run says so.

**Tests (`lib/test/run.sh`)**: producer and consumer cells for each field in both record classes; an adversarial malformed-shape matrix (string / array / number / null / scalar `fix_delta` / missing status), each row paired with a regressed sweep field to pin attribution; vacuity positive controls on both derived-set extractions; a converse assertion catching a constant-side deletion; and `LR_CONDITIONAL_FIELDS` as a single source for both the subtraction and the presence pins — which retroactively closes four fields that were subtracted with nothing pinning their deletion.

**Docs and artifacts**: `docs/efficiency-trace.md` (synthesized shape, the `.sweep_defs_read | length` type-check trap, compatibility boundary), a re-measured `docs/review-and-fix-budget.md`, a growth cutover artifact recording the one deferred site, plus a changeset and regenerated cloud-writer contract / prompt-mass baseline.

## Resolves
Resolves #541

## Test Plan
- [ ] `lib/test/run.sh` passes with an empty skip tally (observed in-env: 10,849–10,850 passed, 0 failed, 0 skipped across runs).
- [ ] `--self-check` warns on a synthesized record whose evidence field is present but carries `[]` / `not-run`, and stays silent on a non-synthesized record where those are legitimate.
- [ ] A malformed `reference_reads` (string, array, number, scalar `fix_delta`) does not suppress the sibling `sweep_defs_read` warning — the guard does not fail open.
- [ ] `shellcheck` and `ruff` clean; `cloud_writer_contract.py verify` matches closure; stale-prose self-scan exits 0.

## Visual Changes
N/A

## Breaking Changes
None. The three fields are additive. A pre-change record written by an earlier build lacks them and `--self-check` now emits a missing-field warning for each — intended fail-loud behavior, narrow in blast radius because `--workpad-dir` is a per-run ephemeral directory.


The01Geek:
# Devflow Review — PR #625

**Status:** 🎉 APPROVE with notes
**Diff profile:** `engine_self_modifying` + `detect_all_audit` → full checklist + always-on agents + forced Phase 3.1.5 completeness critic
**Run:** _(local run)_
**Reviewed HEAD:** 98b797cf319e7fe2eddd6193d692dbdcaa83ebc1
**Last updated:** Phase 4 complete

## Blueprint
- [x] Generate verification checklist (Phase 1) — 31 items
- [x] Verify checklist (Phase 2)
- [x] Review agents (Phase 3) — 5/5 returned, 0 failed
- [x] Aggregate & verdict (Phase 4)

## Verdict: APPROVE with notes (0 Critical, 0 Important, 7 Suggestions — Phase 4.2 rule 6)

**0 Critical · 0 Important · 7 Suggestions.** All five Phase-3 agents returned; the dirty-tree backstop confirms none mutated the working tree.

---

## Phase 0 — setup
- 14 changed files, 548 added / 77 removed lines. Related issue **#541**.
- `has_new_types` = false → `type-design-analyzer` gated out (correctly — shell/markdown only). `pr-test-analyzer` dispatched.
- Phase 0.3.6 blocker-recheck **not** taken: the prior REJECT's `commit_id` is `null`, so precondition-2's join could not be established. Full path run instead.
- **Phase 0.6 stale-prose lint:** exit 0 — **0 STALE**, 12 `UNRESOLVABLE` (informational; never gate the verdict). No adjudication carry-forward needed.

## Phase 1/2 — checklist (31 items)
Two batches; the deduper found no cross-batch duplicates. Under the 100 cap — no `cap_drops`.

Verified by **direct execution** rather than verifier-agent dispatch:

| Check | Result |
|---|---|
| Full suite | **10,961 passed, 0 failed, 0 skipped** |
| `cloud_writer_contract.py verify` | matches closure |
| `prompt-mass-baseline.json` | all six byte counts exact |
| Budget doc (one-pass re-measure) | every figure exact — root 3,197 · ext 2,473 · `shadow-review.md` 11,262 · cumulative 43,366 · peak 16,932 · reduction 33,004 · growth +4,692 (12.1%) |
| `#541` pin literals | all three resolve **exactly once** in the assembled 9-member bundle (`MAXI_SKILL="$MAXI_BUNDLE"`) |
| `else`-arm absolute claim | **TRUE** — fires zero times through the real `--self-check` path on malformed JSON, `[]`, `null`, `"x"`, empty file |

Two checklist items alleging REJECT-grade pin defects (wrong target file; `the`/`this` literal mismatch) were **refuted** — both rested on the assumption that the pins search `SKILL.md` alone.

## Phase 3.1.5 — completeness critic (forced by `detect_all_audit`)
- **Population:** the conditional-field set `LR_CONDITIONAL_FIELDS` subtracts from `LR_SCHEMA`.
- **Independent signal (stated for auditability):** derived the true conditional set as (schema-block fields) − (`ITER_EXPECTED_FIELDS`), *not* the audit's own hard-coded list.
- **Result:** exact match, **8/8** — zero over-subtraction, zero under-declaration. Fail-closed in **both** directions: a new schema-block conditional omitted from the list turns the equality RED; a stale list entry turns the presence pin RED. **No finding.**
- Complete *with respect to this signal only* — not a proof of exhaustiveness.

## Findings

### Suggestions (7)

1. **`skills/review-and-fix/references/fix-delta-gate.md`** — the "Gate-subagent failure" arm ends with *"record … and **proceed** to Step 4"*, while the **"Also persist that result durably"** paragraph sits *after* it. Read sequentially, the `not_verified` arms exit before the persist obligation — leaving `reference_reads` absent, which the schema defines as *"the gate did not run,"* collapsing the exact distinction this field exists to preserve. **Mitigated** (hence Suggestion, not Important): `fixing.md:102` independently anchors it — *"the re-Write is Step 3's last Write — Step 3.5 appends the conditional `reference_reads.fix_delta` after it, per item 7."* Cheap fix: restate the persist inline in each failure arm. `logic_error`
2. **`skills/review-and-fix/references/fixing.md:151`** — no verdict-time surface *explicitly enumerates* reading `.reference_reads.fix_delta.status`, unlike shadow `not_verified` which is named at `SKILL.md:323`. **Mitigated:** `SKILL.md:300` (correctly rewritten by this PR) states the prohibition, and the verdict table routes *"any fail-closed reference outcome above"* to non-clean. Suggest naming the field alongside shadow in the caveat row. `logic_error`
3. **`lib/efficiency-trace.sh` (`do_self_check`)** — `2>&1` merges jq's stderr into the same capture as its stdout field names, so a partial-then-abort run interpolates real violation names into an error breadcrumb. Not a fail-open (the record is correctly declared unchecked), but misattributing. Separate the streams. `api_misuse`
4. **`lib/efficiency-trace.sh:795-797`** — "the `promotion_provenance` arm … whose first predicate is likewise a type check": its literally-first predicate is the skip guard. Substantive claim holds; suggest "first *validating* predicate". `style`
5. **`lib/test/run.sh:23059-23080`** — the malformed-shape matrix rows all set `synthesized: true`; no row asserts a **non**-synthesized record with a malformed `reference_reads` stays silent. Safe twice over today (arm short-circuits, derefs type-guarded); closes cheaply with one fixture. `test_gap`
6. **`skills/review-and-fix/SKILL.md:201`** — the root Schema shows `{"status": "run"}` and `verified|not_verified`, narrower than the shipped domain. Explicitly **not** graded `documented_falsehood`: it is a value-template like every other field, and `fixing.md` is the authoritative surface stating the full domain. `comment_drift`
7. **`docs/efficiency-trace.md`** — pre-existing real iter records now emit two `::warning::` lines each under `--self-check`. Correct and warn-only; worth a backward-compatibility sentence. `other`

### Rejected as false positives
- **`lib/test/run.sh:22892` "derived through grep/sed/tr" allegedly contradicts the "never `tr`" comment.** It does not: the "never `tr`" comment scopes to `LR_COND_RE` (bash parameter expansion), while `LR_CONST` at 22878 genuinely uses `tr ' ' '\n'`. Two different variables.
- **"The worktree has uncommitted modifications."** Not true of the reviewed checkout — this review ran against a clean detached tree at exactly `98b797cf`.

### Corroboration
All findings are **single-source** (N=1) and flagged as such. `devflow:code-reviewer` returned no issues at or above threshold and independently corroborated the budget arithmetic, the jq short-circuit safety, and the word-split safety.

## Acceptance criterion
Issue #541's AC is **met**: `sweep_defs_read`/`sweep_evidence` are reconciled across all four defining surfaces before `reference_reads` is added; synthesized records stamp explicit `{"status":"unrecoverable","reason":…}` via the single `unrec()` emitter (never `[]`/`not-run`); and the producer/consumer matrix — {3 fields} × {producer, consumer} × {synthesized, real} — is **12/12 covered**.

## ⚠️ Stale REJECT not cleared (merge-gate housekeeping)
`dismiss-stale-rejections.sh 625` exited **0 but dismissed nothing**, and the PR remains at `reviewDecision: CHANGES_REQUESTED`. Cause: the prior REJECT (review `4733092178`, `devflow-reviewer[bot]`) has the **legacy** body prefix `## Devflow Review: REJECT`, whereas the script matches the current `## Verdict: REJECT` prefix. This is a **tooling marker-format gap, not a defect in this PR**, and per Phase 4.4 a dismissal failure never downgrades the verdict — but the PR stays wedged until that review is dismissed manually.

## Notes
- Two agents graded findings 1–2 as HIGH. Both were demoted on **located evidence** of mitigating surfaces the agents did not find (`fixing.md:102`, `SKILL.md:300`, `SKILL.md:321/325`) — not on judgment alone.
- Reviewer caveat: this was a read-only pass, so pin mutations were assessed by shape rather than observed RED. The full suite passing at 0 failed is the standing evidence.

The01Geek (commented):
## Verdict: APPROVE with notes — full report in PR comment

> The complete review report (checklist results, findings, details) is in the
> Devflow Review progress comment on this PR.

**0 Critical · 0 Important · 7 Suggestions.** Reviewed HEAD: `98b797cf`.

Issue #690, opened 2026-07-21
Make install-gh-wrapper.sh output 5/7's fingerprint-mode gate platform-aware — it is unsatisfiable on Windows runners


The01Geek:
## Problem Statement

A consumer running DevFlow's writer tiers on a self-hosted **Windows / Git-Bash** runner cannot start an implement run at all once a GitHub App is configured. `scripts/install-gh-wrapper.sh` output 5/7 asserts that the token-fingerprint file's POSIX mode is exactly `600`. On a native-Windows `python3`, that value is not reachable, so the assertion fails on every run, the step exits 1, and the `claude` job aborts **before the agent starts** — `Resolve model provider` through `Run Claude Code` are all skipped.

Reported against plugin version 2.19.9, with a captured failing run: GitHub Actions run 29850313725, job `claude`, step `Install fresh-gh wrapper (optional)`, canary issue 3614 in their repo. Observed step output:

```
install-gh-wrapper: output 5/7 FAILED: fingerprint file
E:\GitHub-3\_work\_temp/devflow-gh-fingerprint is mode 666, not 0600 (fingerprint-mode)
##[error]Process completed with exit code 1.
```

The step is labelled `(optional)` in the workflow, but it carries no `continue-on-error`, so a non-zero exit fails the job.

## Current Behavior

`scripts/install-gh-wrapper.sh` validates seven setup outputs in order and exits 1 on the first failure. Output 5/7 writes the job-start token's sha256 under `( umask 077 … )`, then reads the resulting mode back through `python3` and compares it to the literal `600`:

```bash
( umask 077
  printf '%s' "$APP_TOKEN" \
    | python3 -c 'import hashlib,sys; sys.stdout.write(hashlib.sha256(sys.stdin.buffer.read()).hexdigest())' \
    > "$FINGERPRINT_FILE"
)
…
_fpmode="$(python3 -c 'import os,sys; print(oct(os.stat(sys.argv[1]).st_mode & 0o777)[2:])' "$FINGERPRINT_FILE" 2>/dev/null || true)"
[ "$_fpmode" = "600" ] \
  || fail "output 5/7 FAILED: fingerprint file $FINGERPRINT_FILE is mode ${_fpmode:-unreadable}, not 0600 (fingerprint-mode)"
```

[...]

Issue #619, opened 2026-07-20
Add lib/test/regenerate-artifacts.py — one batched generated-artifact pass before every re-verify suite run


The01Geek:
## Problem Statement
Fix and implement loops in this repo pay N full-suite runs to serially discover drift in suite-owned generated artifacts. In one observed fix loop, three of the failures were mechanical regeneration chores with known commands — the cloud-writer runtime manifest (`python3 lib/test/cloud_writer_contract.py generate`, #543), the prompt-mass census baseline (#551), and the review-bundle budget-record figures (#529) — each discovered by a full suite run, fixed, and re-verified by another full run. The full suite is the slowest verification step in the repo, so each avoidable rerun is the dominant cost of a loop iteration. The drift is induced by the loop's own edits (editing a reached skill asset drifts the cloud-writer manifest; adding prompt prose drifts the census baseline; editing the capability manifest drifts the five workflow literals; adding review-bundle prose stales the budget record), so it is predictable before the rerun starts.

## Current Behavior
Each generated artifact carries its own per-artifact "regenerate in the same commit" rule — CLAUDE.md's coupled-invariant and capability-manifest bullets, and the census section of `.devflow/prompt-extensions/implement.md` — but no surface instructs a batched pass and no single regeneration entry point exists. Verified: a drafting-time docs-verify sweep of CLAUDE.md, the prompt extensions, and `skills/**` found no batching guidance and no regenerate-all wrapper (2026-07-19). An agent therefore discovers each drifted artifact one full-suite run at a time.

## Desired Behavior
A suite-owned batch entry point, `lib/test/regenerate-artifacts.py`, owns the enumeration of suite-owned generated artifacts. After applying edits and before each full-suite re-verify run, the loop runs `python3 lib/test/regenerate-artifacts.py` once; the helper regenerates the mechanically-safe artifact unconditionally, runs each judgment-gated artifact's non-writing check, and prints the resulting judgment items together in one report (a single invocation with several drifted artifacts reports them all). The agent commits the mechanical regeneration with its edits and resolves each judgment item under that artifact's named governing policy, so the next suite run verifies a tree whose generated artifacts are already reconciled.

## User Impact
[...]

Issue #591, opened 2026-07-19
Populate the focused-test-module registry: coverage map, module-authoring convention, coverage ratchet, and a capability-profiles seed module


The01Geek:
## Dependencies
Blocked by #561 — the capability-profile generator and its `lib/test/run.sh` assertion coverage (PR #588) must merge to main before the seed module can extract them.

## Problem Statement
The selectable focused-test-module runner works: `lib/test/run-module.sh <module-id>` selects a module from the `test_modules` section of `scripts/workflow-flight-recorder-registry.json`, and the review-and-fix prompt extension mandates focused modules for tight RED/GREEN fix iteration. But the registry declares exactly three modules — `workflow-flight-recorder`, `review-and-fix-contract`, `create-issue-contract` — all skill/workflow contract modules. The rest of the tested surface — the `lib/` retrospective and build/CI helpers, the `scripts/` Python and shell CLIs, and the config/jq consumers — has no registered module, so the extension's fallback rule ("When no registered module covers the fix, use the full suite during iteration") sends every fix touching that majority to the full `lib/test/run.sh` suite (44,385 lines) on every iteration, at roughly 3–7 minutes per run (an observed-run estimate, not a repo-derivable figure). Concretely, the recent `/devflow:implement` 561 resume ran the full suite about five times (20+ minutes of suite wall-clock, same estimate basis) because no module covers the capability-profile generator. That cost recurs for every future ticket touching `lib/` and `scripts/`.

**Relationship to #586 (deliberately orthogonal):** #586 hardens the runner/boundary signal-cleanup mechanics of the machinery itself; this issue populates what that runner can select. The two share no acceptance criteria, and modules added here inherit whatever runner hardening #586 ships.

## Current Behavior
- The registry's `test_modules` section maps three module ids to `{path, minimum_assertions, description}` (floors 68/68/206). All three are skill/workflow contract modules.
- The review-and-fix prompt extension reads: "When no registered module covers the fix, use the full suite during iteration. Cloud-tier runs continue using the already-permitted complete suite without requesting new permissions." (`.devflow/prompt-extensions/review-and-fix.md`, "Focused test modules accelerate fix iteration only".)
- The focused runner is granted nowhere on the cloud tiers: `.devflow/config.json` grants `Bash(lib/test/run.sh:*)` in both `devflow_implement.allowed_tools` and `devflow.allowed_tools`, and `run-module.sh` appears in neither.
[...]

Issue #544, opened 2026-07-16
scripts: deferred review findings in scripts/gh-fresh.sh (carried from #533)


The01Geek:
Carried forward from the /implement run on #533 (PR #538).

The following review-agent findings were surfaced during PR review but deferred under the Scope-Acknowledged Findings contract. They are tracked here for follow-up resolution. Closing this issue invalidates the related deferral and forces re-verification on the next /devflow:review run.

### Suggestion — devflow:silent-failure-hunter
**File**: scripts/gh-fresh.sh:92-126
**Symbol**: sha256_of
**Kind**: logic_error

gh-fresh.sh's call-time fingerprint comparison (sha256_of) still depends on sha256sum/shasum+awk — the tool family the installer was migrated off for guard-class-2 reasons. On a hash-tool-less runner the freshness feature degrades per-call (disclosed defer breadcrumb, fail-safe direction) instead of failing loudly at install. Converge sha256_of onto python3 hashlib (writer/reader symmetry) in a follow-up.

**Why deferred**: out-of-scope — gh-fresh.sh executable lines are unmodified by this PR (comment-only diff); AC16 explicitly preserves the wrapper's existing decision outcomes including the no-hash-tool defer arm; converging the reader is follow-up material. Widens-surface check: the PR's gh-fresh.sh hunk touches lines 9-18 (header) only — no overlap with lines 92-126 within ±10.

---
Filed automatically by devflow-implement.

The01Geek:
Resolved directly in PR #538 (commit 599bbbb3) after the cloud review surfaced it as a Suggestion and rejected the deferral (untrusted-filer): `gh-fresh.sh`'s `sha256_of` now tries the preflight-guaranteed python3 `hashlib` first (writer/reader symmetry with `scripts/install-gh-wrapper.sh`), keeping `sha256sum`/`shasum` as fallbacks. The disclosed could-not-establish defer arm is preserved and both directions are suite-tested (a hash-tool-less match via the python3 arm, and the defer breadcrumb when every method is defeated). PR #538's body now carries `Resolves #544`, so this closes on merge.

PR #697, opened 2026-07-21
Extract the run.sh installer/workflow-wiring region into a focused test module and enforce coverage-map block ownership

### LINKED ISSUE #695: Extract the run.sh installer/workflow-wiring region into a focused test module and enforce coverage-map block ownership
## Problem Statement

Iterating on installer and `gh`-wrapper coverage costs a full `lib/test/run.sh` run. `run.sh` is over 51,000 lines; a reporting agent measured roughly five minutes per run and worked around it by hand-rolling a throwaway extraction harness that ran the same coverage in roughly thirty seconds. The supported way to get that loop is a focused module under `lib/test/run-module.sh`, but no module covers the installer region, so the fast loop had to be rebuilt ad hoc and discarded.

The same reporter read the coverage map's `"533"` entry, saw `owner: "unmodularized"`, and concluded that no focused target exists. The conclusion is correct; the inference that produced it is not, and that is the second, larger problem. `owner` is not a signal today: every `run_sh_blocks` entry reads `unmodularized`, so the field distinguishes nothing. The map is also incomplete, and it is blind to the modules that already exist — labels whose assertions live wholly inside extracted modules have no entry at all. Nothing detects any of this, because `lib/test/coverage_map_guard.py` has no completeness or attribution arm for `run_sh_blocks`. The map is documented as "a ranked to-do list" for future extractions, and it cannot serve that purpose while its only ownership signal is constant.

## Current Behavior

**Measurement snapshot.** Every figure in this section was measured against commit `93e5cd13` (`main`). They are a dated snapshot, not a live contract: `run.sh` grows continuously, so an implementer re-deriving them later will find larger counts. The acceptance criteria below are written against the derivation, never against these literals.

Verified: `python3` read of `lib/test/modules/coverage-map.json` — `run_sh_blocks` holds 195 entries and their owner values are `Counter({'unmodularized': 195})`; `files` holds 113 entries with `Counter({'unmodularized': 112, 'capability-profiles': 1})`.

Verified: `lib/test/coverage_map_guard.py` arm 3 is the only arm that reads `run_sh_blocks`, and it checks owner referential validity alone — `for label in sorted(run_sh_blocks): owner = run_sh_blocks[label].get("owner")` feeding `_valid_owner`, which returns `True` for the literal `unmodularized` unconditionally. Arms 1, 2, 5 and 6 read `files` and `non_code_exempt` exclusively. No arm compares `run_sh_blocks` membership against `run.sh`.

Verified: `lib/test/modules/coverage-map.json`'s `generated_by` is a checked-in `python3` one-liner that derives the label set mechanically, but the guard only type-checks it as a string (`if not isinstance(map_value.get("generated_by"), str)`), so it is never executed and is an unverified provenance claim.

Verified: running that `generated_by` scanner over `93e5cd13` yields 230 distinct labels from `run.sh` assertion names against 195 map keys, so at least 36 labels present in `run.sh` have no map entry. Zero map keys are stale in the other direction. This count is a floor, not a total, because the scanner's own head list is incomplete — see the next bullet.

Verified: the `generated_by` scanner matches only the heads `assert_eq`, `assert_true`, `assert_pin_unique`, `assert_pin_red_under`, `assert_pin_red_on_removal` and `check`. Extracted modules assert predominantly through other heads — counted across `lib/test/modules/*.sh`: `devflow_module_pin_unique` 209, `assert_eq` 201, `_ra_has` 59, `_cap_fail` 46, `_raf_pin_unique` 45, `_ra_live_unchanged` 33, `devflow_module_pin_red_under` 27. The namespaced harness API and the module-private wrappers are therefore invisible to the current scanner, which is why any label set derived from it under-reports module coverage.

Verified: scanning the modules with a head set that also covers the namespaced pin API and the module-private wrappers yields 20 module-carried labels at `93e5cd13`. Eleven of them are absent from `run.sh` entirely, meaning their coverage is fully extracted: 462, 464, 465, 467, 522, 559, 561, 593, 613, 614 and 624. All eleven are absent from the coverage map. Per-module counts under that wider scan are `create-issue-contract` 15, `regenerate-artifacts` 3, `capability-profiles` 2, `review-and-fix-contract` 0, `workflow-flight-recorder` 0.

Verified: the installer coverage is **not** delimited by its labelled assertions. The first `#533`-named assertion sits inside the body of `for _wf487 in devflow-implement devflow; do`, whose `done` closes 27 lines later, and the last `#533`-named assertion is a two-line statement followed by the `rm -f "$_t533j"` / `rm -rf "$D533"` fixture teardown. A slice taken at the first and last labelled assertion is therefore not valid shell — it carries an orphan `done` and an unassigned `_wf487`. The extractable unit is the **syntactic** one: the enclosing `for _wf487` loop header through the `rm -rf "$D533"` teardown.

Verified: measured at `93e5cd13`, that syntactic unit spans 613 lines and holds 87 assertions across six labels: 533 (40), 487 (22), 690 (12), 599 (6), 491 (5), 544 (2). Occurrences of `#533` elsewhere in `run.sh` are comments citing issue #533 as history, correctly excluded by a names-only scanner.

Verified: the unit does not exhaust every label it contains. Counting assertion names outside it, `#487` retains 86 assertions and `#491` retains 32 in the monolith, while `#533`, `#690`, `#599` and `#544` retain zero. Extraction therefore fully removes four of the six labels and partially removes two.

Verified: the unit calls three helpers defined in `run.sh` and absent from the harness — `mint_blk` 13 times (defined roughly 28,000 lines above the unit, and used 28 times in `run.sh` overall so 15 uses remain behind), `probe_tmp` 4 times, and `probe_assert` twice. A `grep` for all three across `lib/test/module-harness.sh` returns zero matches, so none is part of the exported module API.

Verified: the unit's external dependencies are not only functions. It expands the `run.sh` global `WF` 14 times (as `_WFF487="$WF/$_wf487.yml"` and as direct `"$WF/devflow.yml"` reads). `WF` is assigned once, at `WF="$LIB/../.github/workflows"`, roughly 28,000 lines above the unit and outside it, and neither `lib/test/run-module.sh` nor `lib/test/module-harness.sh` assigns it. Both the focused-runner path and the full-suite boundary run the module body under `set -u`, so a verbatim extraction aborts on the unit's first statement with `WF: unbound variable` before any assertion executes. A variable sweep over the unit found `WF` to be the only live external: every other bare-looking token sits inside a single-quoted heredoc or `grep -F` literal, and the unit's own locals are assigned within it. `run.sh` retains its own `WF` assignment, which the 15 remaining `mint_blk` uses and 86 remaining `#487` assertions depend on.

Verified: `run.sh` is never shellchecked — `.github/workflows/ci.yml` lints `git ls-files '*.sh' | grep -v '^lib/test/'`, and its second, explicit invocation lints `module-harness.sh`, `run-module.sh`, `summary.sh` and each module at `--severity=warning`. Moving this code into a module and the three helpers into the harness therefore submits previously-unlinted code to a gating linter for the first time, and it does not currently pass: `shellcheck --severity=warning` over the unit reports `SC2069` on `bash "${1:-$INSTALL533}" 2>&1 1>/dev/null`, an idiom whose stderr-only capture is deliberate.

Verified: `lib/test/modules/regenerate-artifacts.sh` registers the coverage map as a **judgment** artifact in the batched generated-artifact pass. Its `RA_ROW_NAMES` includes `coverage-map-ratchet`; the row is classified `by-hand` (`_ra_class_is coverage-map-ratchet by-hand`); and the pass asserts the artifact is never mechanically written — `_ra_cmp "#619 A3 write scope: the coverage map is byte-unchanged" "$RA_A3_COVMAP" lib/test/modules/coverage-map.json`.

Verified: `lib/test/coverage_map_guard.py`'s `evaluate(tracked_files, map_value, registry_value, *, map_read_error=None, registry_read_error=None)` performs no file access — zero `open(`, `glob` or `Path(` occurrences in its body — and its docstring states "Never raises". Every read lives in `main()`, and read failures are injected as the `map_read_error` and `registry_read_error` keyword arguments. `lib/test/test_coverage_map_guard.py` calls `evaluate(` at 35 sites.

Verified: `grep -n 'DEVFLOW_GH' lib/test/run-module.sh` returns no match, and the same search across `lib/test/module-harness.sh` and every existing module returns none. The inherited-`DEVFLOW_GH` clearing that the installer override-contract assertions depend on lives only in `run.sh`'s preamble, so a module sourced by the focused runner inherits whatever value the caller exported.

Verified: `python3` count of `skip` invocations inside the unit returns zero, so it is compatible with the harness rule that a module may not self-skip.

## Desired Behavior

The installer and workflow-wiring coverage runs as a registered focused module, so `lib/test/run-module.sh installer-wiring` exercises it without a full-suite run, and the complete suite runs the identical module through the shared boundary.

The three helpers the unit depends on — `mint_blk`, `probe_tmp` and `probe_assert` — move into `lib/test/module-harness.sh` as shared helpers that `run.sh` and the modules both obtain by sourcing, so no second copy of any of them exists.

The inherited-`DEVFLOW_GH` fixture isolation that `run.sh` performs in its preamble is performed by `lib/test/module-harness.sh`, so any module the harness sources runs under the same isolation the monolith's preamble guarantees.

`lib/test/coverage_map_guard.py` gains an arm that derives issue labels mechanically from `lib/test/run.sh` and from `lib/test/modules/*.sh` through one shared implementation, and turns the suite RED when the map's `run_sh_blocks` half disagrees with that derivation. It fails on a label present in `run.sh` with no map entry, and on a label whose coverage is **fully** extracted — carried by a module and absent from `run.sh` — whose entry is absent or does not name the module carrying it. A label that is partially extracted, carried by a module while assertions remain in `run.sh`, keeps `unmodularized` and is not an attribution violation, because a single `owner` string cannot truthfully describe split coverage.

The guard gains an explicit, hand-invoked `--fix` mode that repairs the map to satisfy its own arm, so the recurring obligation the arm creates has a mechanical remedy. The coverage map remains a `by-hand` judgment artifact in the batched generated-artifact pass, never written mechanically by it.

## User Impact

Anyone changing `scripts/install-gh-wrapper.sh` or the workflow install steps gets a focused verify loop instead of a full-suite run. That population includes autonomous `/devflow:implement` runs touching those surfaces, which pay the full-suite cost on every verification attempt inside a single run.

Everyone extracting a future module gets a coverage map whose `owner` field means something for fully-extracted coverage, so the "which block should I extract next" question is answerable from the artifact that claims to answer it.

This change also **broadens** a per-PR obligation rather than removing one, and that trade is deliberate. Today an author who extracts a module can silently skip the ownership update; after this change the guard fails the suite instead. But the same arm also means any PR adding an assertion named for a new issue must give that label a map entry, which was previously unenforced. The `--fix` mode is the remedy that keeps that obligation cheap, and the documentation criterion below puts the obligation where an assertion author will actually meet it rather than only in the module-authoring checklist.

## Technical Context

- **Relevant Classes/Files**
  - `lib/test/run.sh` — the monolith; the unit to extract, the `mint_blk`/`probe_tmp`/`probe_assert` definitions, the `#591 coverage-map guard: shipped tree + map is clean` assertion, and the `devflow_run_full_suite_module` boundary calls.
  - `lib/test/run-module.sh` — the focused runner and its registry-driven selector.
  - `lib/test/module-harness.sh` — the shared harness and namespaced pin API; recipient of the `DEVFLOW_GH` clearing and of the three promoted helpers.
  - `lib/test/modules/` — destination for the new module and its sibling inventory file.
  - `lib/test/modules/coverage-map.json` — `run_sh_blocks`, `files`, `generated_by`.
  - `lib/test/coverage_map_guard.py` — `evaluate`, `main`, `_valid_owner`, `_registry_module_ids`, `_map_shape_error`, and the new `--fix` mode.
  - `lib/test/test_coverage_map_guard.py` — the guard's focused Python suite and its 35 existing `evaluate(` call sites.
  - `lib/test/modules/regenerate-artifacts.sh` — the `coverage-map-ratchet` judgment row and the `#619 A3` write-scope assertion that the map is byte-unchanged.
  - `scripts/workflow-flight-recorder-registry.json` — `test_modules`.
  - `lib/test/test_module_runner.py` — cross-checks each registry floor against its `run.sh` call-site literal.
  - `.github/workflows/ci.yml` — the explicit shellcheck file list, which does not glob module paths.
  - `CONTRIBUTING.md` — "Authoring a new focused module", plus the assertion-authoring surface the new obligation must be documented on.
  - `docs/DEVFLOW_SYSTEM_OVERVIEW.md` — "Coverage map and ratchet (issue #591)".

- **Architecture Alignment** — this follows the established extraction path: `CONTRIBUTING.md`'s seven-item same-PR checklist, the `module-harness.sh` module contract, and the namespaced pin API that replaces the monolith's pin helpers inside a module. The new guard arm follows the existing arm shape — append a message to `violations`, run after arm 4's fail-closed map admission — and preserves the file's established boundary in which `main()` performs I/O and `evaluate` stays a pure, never-raising function over injected values.

- **Dependencies** — `python3` and `git` only, both preflight-guaranteed. The guard already declares itself "driven by the complete suite through `git` and `python3` only" and the new arm adds no third dependency.

- **Data/Schema Considerations** — `coverage-map.json` keeps `schema_version` unchanged: no key is added or removed, and `owner` stays a single string. The partial-extraction rule exists precisely so a single-`owner` entry is never asked to describe split coverage, which is what would have forced a schema change.

- **Cross-layer Impact** — repo-internal test infrastructure only. Verified: `install.sh`'s workflow copy loop is `for w in devflow devflow-runner devflow-implement devflow-review telemetry-push`, and no `lib/test/` path is shipped by any channel, so no consumer repository is affected and the `install.sh`-versus-`devflow_version` two-artifact skew cannot arise.

- **Cloud allowlist** — no new tool grant is required. Verified against `CLAUDE.md`'s tier-2 bullet: `Bash(lib/test/run.sh:*)` and `Bash(lib/test/run-module.sh:*)` are granted in both `devflow_implement.allowed_tools` and `devflow.allowed_tools` as **direct leading-token** forms, and a new module id is an argument to the already-granted `run-module.sh` head. The `bash <path>` wrapper form stays deny-floored, so the criteria below invoke these scripts directly. Because both grants exist on the default branch today, no in-run verification obligation here depends on a trigger-time-resolved config change this PR would ship (the in-PR-inert channel of issue #593).

- **Coupled mirror sites** — Verified by a repo-wide whitespace-normalized search (`tr -s '[:space:]' ' '` over `git ls-files`, matching `run_sh_blocks`): the five hits are `CONTRIBUTING.md`, `docs/DEVFLOW_SYSTEM_OVERVIEW.md`, `lib/test/coverage_map_guard.py`, `lib/test/modules/coverage-map.json`, `lib/test/test_coverage_map_guard.py`. The same search for `unmodularized` returns those five minus `CONTRIBUTING.md`. `lib/test/run.sh` is not a hit — it invokes the guard without naming the key. Two of those hits state a contract this change falsifies and must be rewritten in the same PR:
  - `CONTRIBUTING.md` item 6: "The `run_sh_blocks` half of the map is hand-maintained (mechanical `run.sh` block scanning is a disclosed non-goal), and ownership is a curated claim the guard checks for referential validity, not semantic correctness."
  - `docs/DEVFLOW_SYSTEM_OVERVIEW.md`: "The `run_sh_blocks` half is hand-maintained (mechanical `run.sh` block scanning is a disclosed non-goal)."

- **Mechanism dependencies the design relies on** — Verified: `_registry_module_ids` returns `None` when the registry is absent, unreadable, or wrong-shape, and `_valid_owner` then returns `True` unconditionally with the comment "Arm 8 already recorded the registry failure; do not double-report every owner here." The new arm's attribution half reads the same registered-id set and therefore inherits that fail-open, which is why a criterion below requires it to stand down when the id set is unavailable. Verified: `_map_shape_error` already rejects a non-object `run_sh_blocks` and any entry that is not an object with a string `owner`, so the new arm may assume that shape only because arm 4 runs first. Verified: `evaluate` takes no path argument and performs no I/O, so the arm's inputs must be produced by `main()` and injected, exactly as the existing read-error keyword arguments are. Verified: `probe_assert`'s purpose in `run.sh` is to run an assertion whose RED verdict is the probed value rather than a suite failure, so promoting it must preserve that the probed failure never reaches the module tally — the same tally the floor probe measures.

## Acceptance Criteria

- [ ] `scripts/workflow-flight-recorder-registry.json` `test_modules` contains an `installer-wiring` entry whose `path` is `lib/test/modules/installer-wiring.sh`, with a `description` and a positive integer `minimum_assertions`.
- [ ] `lib/test/module-harness.sh` defines `mint_blk`, `probe_tmp` and `probe_assert`, `lib/test/run.sh` obtains all three by sourcing the harness rather than defining them, and no second definition of any of the three exists anywhere in the tree.
- [ ] A promoted `probe_assert` keeps its existing contract that a probed RED verdict is returned to the caller and never counted as a failure in the tally of whichever runner is executing, so the module's executed-assertion count reflects only real assertions.
- [ ] `lib/test/modules/installer-wiring.sh` carries the assertions moved out of the **syntactic** unit described in Current Behavior — the enclosing `for _wf487` loop header through the `rm -rf "$D533"` teardown — not a slice cut at the first and last labelled assertion. The extracted text is valid shell on its own: it contains no unmatched loop keyword, and it references no variable that it neither assigns itself nor derives from the harness-provided `LIB`.
- [ ] `lib/test/modules/installer-wiring.sh` re-derives `WF` from `LIB` at module top rather than inheriting it from `run.sh`, and `lib/test/run-module.sh installer-wiring` passes with the module's only inherited inputs being the harness-provided `LIB` and `RESULTS_FILE`. `lib/test/run.sh` keeps its own `WF` assignment for the coverage that stays behind.
- [ ] The module covers every label the extracted unit contains, as derived at extraction time. At `93e5cd13` that set was 533, 487, 690, 599, 491 and 544; the criterion is satisfied against the set derived when the extraction is performed, not against that snapshot.
- [ ] The `minimum_assertions` floor is established with the over-floor probe described in `CONTRIBUTING.md` item 2 — set a deliberately over-high floor, run the suite, read the boundary's `executed N assertions; minimum is M` line, and set the floor to the observed `N`. The floor is the observed executed count, not a statically counted number of assertions in the unit.
- [ ] The same floor literal appears at the `devflow_run_full_suite_module` call site in `lib/test/run.sh`, and `lib/test/test_module_runner.py`'s existing registry-versus-call-site cross-check passes for `installer-wiring`.
- [ ] `lib/test/run-module.sh installer-wiring` exits 0 on the clean working tree and reports an executed assertion count greater than or equal to its floor.
- [ ] The module calls `skip` nowhere and invokes no helper that is defined in `lib/test/run.sh` and absent from `lib/test/module-harness.sh`, satisfying the module contract documented in the harness header.
- [ ] `lib/test/modules/installer-wiring.inventory.md` records the module's provenance — its source baseline and its coverage groups.
- [ ] `.github/workflows/ci.yml`'s explicit shellcheck file list names `lib/test/modules/installer-wiring.sh`.
- [ ] `shellcheck --severity=warning -e SC1091` passes over `lib/test/modules/installer-wiring.sh` and over `lib/test/module-harness.sh` after the move. Each warning newly surfaced by moving previously-unlinted code under the gating linter is resolved behavior-preservingly — in particular the `SC2069` site is fixed by brace-grouping the redirect (`{ cmd >/dev/null; } 2>&1`), never by reordering `2>&1` last, which would capture a different stream and silently change the assertion it feeds.
- [ ] `lib/test/module-harness.sh` clears an inherited `DEVFLOW_GH` before a module body is sourced, so a focused run started with `DEVFLOW_GH` exported in the environment produces the same assertion outcomes as one started with it unset.
- [ ] `lib/test/run.sh`'s preamble retains its own `DEVFLOW_GH` clearing and its `DEVFLOW_AC13_PROBE` probe mode unchanged.
- [ ] `lib/test/coverage_map_guard.py` gains a single label-derivation implementation, used for both `lib/test/run.sh` and `lib/test/modules/*.sh`, that anchors on assertion-name position so a label appearing only in a comment is not derived.
- [ ] That derivation covers assertions made through the namespaced harness API (`devflow_module_pin_unique`, `devflow_module_pin_red_under`, `devflow_module_pin_present`, `devflow_module_pin_count`) and through module-private assertion wrappers, so that for every module carrying labelled assertions the derived label set is non-empty and contains each label that module asserts. Measured at `93e5cd13`, the current `generated_by` head list fails this criterion by deriving 16 module labels where the wider scan derives 20.
- [ ] `main()` performs the `lib/test/run.sh` and `lib/test/modules/*.sh` reads and injects the derived label sets and any scan-read failures into `evaluate` as keyword arguments, mirroring the existing `map_read_error` and `registry_read_error` convention. `evaluate` performs no file access, still never raises, and its existing positional call contract is unchanged, so all 35 existing `evaluate(` call sites in `lib/test/test_coverage_map_guard.py` continue to pass without edit.
- [ ] The arm records a violation for every label derived from `lib/test/run.sh` that is absent from `run_sh_blocks`.
- [ ] The arm records a violation for every **fully-extracted** label — one derived from a module and absent from the `lib/test/run.sh` derivation — whose `run_sh_blocks` entry is absent, and for every such label whose entry names `unmodularized` rather than a module that carries it.
- [ ] The arm records no attribution violation for a **partially-extracted** label — one derived from both a module and `lib/test/run.sh`. Because `#487` and `#491` both retain assertions in `run.sh` after the extraction, both remain `unmodularized`.
- [ ] The arm exempts the synthetic `unlabeled` aggregate key from both checks, since no mechanical derivation produces it.
- [ ] The arm records no attribution violation when `_registry_module_ids` yields no id set, leaving the registry failure to arm 8, matching `_valid_owner`'s existing stand-down.
- [ ] The arm records a violation naming the unreadable path when `lib/test/run.sh` or a module file cannot be read, rather than treating an unreadable source as an empty label set.
- [ ] `lib/test/coverage_map_guard.py` gains a `--fix` mode that rewrites `lib/test/modules/coverage-map.json` so the new arm reports no violation, adding a missing `run.sh` label as `unmodularized` and setting a fully-extracted label's owner to a module that carries it. Running `--fix` twice in a row leaves the file byte-unchanged on the second run.
- [ ] Adding `--fix` preserves the guard's existing CLI contract: `main` still accepts the repo root as a positional argument, and `lib/test/run.sh`'s existing invocation of the guard is not edited.
- [ ] `lib/test/modules/regenerate-artifacts.sh`'s `coverage-map-ratchet` row remains classified `by-hand`, and its `#619 A3 write scope: the coverage map is byte-unchanged` assertion still passes, so the batched generated-artifact pass never writes the map mechanically and `--fix` stays a hand-invoked repair.
- [ ] `lib/test/modules/coverage-map.json` is brought into compliance with the new arm by running `--fix` rather than by hand-transcribing a label list. At `93e5cd13` that repair covered at minimum the 36 unmapped `run.sh` labels and the 11 fully-extracted labels 462, 464, 465, 467, 522, 559, 561, 593, 613, 614 and 624; the criterion is satisfied against the tree at implementation time.
- [ ] `lib/test/modules/coverage-map.json` records `installer-wiring` as the owner of every label the extraction fully removes from `run.sh`, and as the `files` owner of `scripts/install-gh-wrapper.sh`.
- [ ] `lib/test/modules/coverage-map.json`'s `generated_by` value names the guard as the derivation's owner instead of restating a scanner the guard never executes.
- [ ] `lib/test/run.sh` passes with zero failures on the resulting tree, and the existing `#591 coverage-map guard: shipped tree + map is clean` assertion still reports rc 0.
- [ ] `CONTRIBUTING.md` item 6 and the `docs/DEVFLOW_SYSTEM_OVERVIEW.md` coverage-map paragraph no longer state that the `run_sh_blocks` half is hand-maintained or that mechanical `run.sh` block scanning is a non-goal, and instead state that the guard derives and enforces it, including the partial-extraction rule.
- [ ] `CONTRIBUTING.md` documents the new per-PR obligation and the `--fix` remedy in a section an author adding an assertion reaches, not only inside "Authoring a new focused module".
- [ ] The PR carries a `.changeset/*.md` file with `bump: patch` and Keep-a-Changelog prose. The change edits `scripts/workflow-flight-recorder-registry.json` and `.github/workflows/ci.yml`, which are engine surface **by path**, and the Phase 3 review gate FAILs an engine-surface change carrying no changeset — so a changeset is required even though the change is functionally test infrastructure.

## Implementation Notes

- **Approach** — Coupled edits landing in one PR.

  First, promote `mint_blk`, `probe_tmp` and `probe_assert` from `run.sh` into `lib/test/module-harness.sh`, and delete their `run.sh` definitions so the monolith obtains them by sourcing. Promotion rather than copying is deliberate: 15 of the 28 `mint_blk` uses stay in `run.sh` alongside the `#487` assertions that are not extracted, so a copy would create an uncoupled mirror of an awk block whose job-boundary logic is load-bearing — the coupled-mirror defect class `CLAUDE.md` names as dominant. Do this before the extraction, so the module can be written against the harness API from the start.

  Second, extract the syntactic unit — the `for _wf487` loop header through the `rm -rf "$D533"` teardown — into `lib/test/modules/installer-wiring.sh` whole, without splitting it by label. Its assertions share fixtures and `run.sh`'s own comment records that the `#487` install step "delegates its whole body to the checked-in seven-output installer (issue #533)", so the labels are semantically coupled and a label-wise split would leave dangling locals. Replace the removed unit with the `devflow_run_full_suite_module` boundary call, and convert monolith pin helpers to the namespaced API.

  Third, move the inherited-`DEVFLOW_GH` clearing into `lib/test/module-harness.sh` so it runs before a module body is sourced, on both the focused-runner path and the full-suite boundary path.

  Fourth, add the guard arm and the `--fix` mode. Implement one label-derivation function inside `lib/test/coverage_map_guard.py`, call it from `main()` for both the monolith and the module glob, and inject its results into `evaluate` alongside the existing read-error keywords so the pure-function boundary and the 35 existing call sites survive intact. Keep the coverage map a `by-hand` judgment artifact in the batched pass; `--fix` is a hand-invoked repair, which is what lets the `#619 A3` write-scope assertion stay green. Point `generated_by` at the guard so one implementation owns the derivation.

  Note the ordering constraint: converting the unit's `assert_pin_unique` and `assert_pin_red_under` calls to `devflow_module_pin_*` is exactly the transformation that makes the current scanner blind, so the derivation must cover the namespaced API before `--fix` regenerates the map.

- **Code Patterns** — mirror `lib/test/modules/capability-profiles.sh` and its inventory for module shape; mirror the existing arm structure in `coverage_map_guard.py` (a bracketed `[armN]` prefix, a remedy clause, an appended `violations` entry); mirror the `map_read_error`/`registry_read_error` keyword-injection convention for the new scan inputs; mirror `lib/test/test_coverage_map_guard.py`'s synthetic-fixture style for the new arm's cases.

- **Testing Strategy**

  *Move 1 — boundary and level.* Automated coverage applies on every half. The guard arm has a pure-function boundary: `evaluate(...)` returns violation strings for injected values, which `lib/test/test_coverage_map_guard.py` already drives directly — unit level. The label derivation is a second pure function over file contents, unit-tested against temporary fixture files. The `--fix` mode has a CLI/file-effect boundary, tested by fixing a synthetic broken map and re-evaluating. The module extraction and the helper promotion have a CLI-contract boundary: `lib/test/run-module.sh installer-wiring` exit code and executed-assertion count, plus the full-suite boundary call — integration level. All are exercised; none substitutes for another.

  *Move 2 — coverage dimensions.*
  - **Happy path** — the arm returns no violation for a compliant map; `--fix` makes a non-compliant map compliant; the focused module exits 0 at or above its floor.
  - **Boundary and degenerate inputs** — a source deriving zero labels; an empty `lib/test/modules/` glob; a `run_sh_blocks` containing only the synthetic `unlabeled` key; a label carried by two modules at once; a label carried by a module while remaining in `run.sh`, asserted to produce no attribution violation; a label whose every assertion leaves `run.sh` in the same change, asserted to require attribution.
  - **Error and failure paths** — an unreadable `run.sh`; an unreadable module file; a registry that yields no id set, which must produce no attribution violation; a map rejected by arm 4, after which the new arm must not run at all. Each asserts the specific violation text or its specific absence, not merely that something failed.
  - **Adversarial and malformed input** — the arm reads the map, so it inherits the repo's config-JSON adversarial matrix, exercised for the `run_sh_blocks` half specifically: object, array, scalar, valid-falsy, missing, and wrong-type. An entry whose `owner` is a non-string, and an owner naming a module that does not carry the label, are both included. `--fix` is additionally driven against each of those shapes and must refuse to write rather than corrupt a malformed map. On the text-scanning side the fixtures include an assertion name carrying two issue labels, a label token inside a comment rather than an assertion name, a label token inside a quoted string that is not an assertion name, and an assertion made through a module-private wrapper.
  - **State, concurrency and idempotency** — running the guard twice over an unchanged tree yields byte-identical violation lists; running `--fix` twice leaves the file byte-unchanged on the second run.
  - **Scale and performance** — not applicable; the scan is a single pass over files the suite already reads, and no criterion implies a bound.
  - **Security and authorization** — not applicable; the change crosses no access boundary and reads only tracked repository files.

  *Move 2a — governing conventions.* `governing conventions consulted:` `CLAUDE.md` (the six-shape adversarial matrix for config-JSON consumers — object, array, scalar, valid-falsy, missing, wrong-type — enumerated in full above for the `run_sh_blocks` half), `CONTRIBUTING.md` ("Authoring a new focused module", whose seven items are each covered by a criterion above), `docs/DEVFLOW_SYSTEM_OVERVIEW.md` (the coverage-map, module-harness, and batched-artifact-pass paragraphs).

  *Move 3 — named assertions.*
  - The guard-arm assertions live in `lib/test/test_coverage_map_guard.py`, one per case named in Move 2, each asserting the presence or absence of the arm's specific violation string. Every criterion naming arm, derivation, or `--fix` behavior maps to at least one of them, and each maps back to exactly one criterion.
  - Each guard-arm assertion is test-first and must fail first because the arm does not exist yet, so `evaluate` returns no such violation.
  - A derivation assertion covers the head-coverage criterion directly: a fixture module asserting through `devflow_module_pin_unique` and through a private wrapper must yield both labels, which fails first under the current head list.
  - A helper-promotion assertion pins that `mint_blk`, `probe_tmp` and `probe_assert` are defined exactly once in the tree and that the definition lives in `lib/test/module-harness.sh`, which fails first because each is defined in `run.sh` today.
  - **Planted-defect positive control, discharging the detector-coverage claim.** The guard is a deterministic backstop for a step an author can skip, so the test must prove it fires on the skipped-step path rather than only on the cooperative one. Planted defects are required, one per violation the arm can record: remove one label's entry from `run_sh_blocks` while it remains in `run.sh` and assert the arm reports it; set a fully-extracted label's owner back to `unmodularized` and assert the arm reports it; delete a fully-extracted label's entry entirely and assert the arm reports it. Each is planted against a scratch copy, never the tracked map.
  - **Negative control for the partial-extraction rule** — plant a partially-extracted label marked `unmodularized` and assert the arm reports nothing, so the rule that keeps `#487` and `#491` unowned is pinned rather than assumed.
  - **Write-scope regression control** — assert that a full suite run leaves `lib/test/modules/coverage-map.json` byte-unchanged, so the `--fix` mode cannot drift into the batched pass and falsify the `#619 A3` assertion.
  - The floor-related assertion is an obligation on the implementer, not a prediction: the over-floor probe must be run and the observed executed count recorded. This issue states no expected floor value.
  - Fixtures are synthetic maps, registries, and temporary scanned files built in the test. The units under test — `evaluate`, the new arm, the label derivation, and `--fix` — are never mocked.

- **Documentation Needed**
  - `CONTRIBUTING.md`
  - `docs/DEVFLOW_SYSTEM_OVERVIEW.md`

- **Potential Gotchas**
  - The extraction changes which labels `run.sh` derives, so the new arm's own inputs shift within the same PR. The map repair, the helper promotion, and the extraction must land together; landing one without the others leaves the suite RED.
  - The coverage map is a `by-hand` judgment artifact in the batched generated-artifact pass, which asserts it is byte-unchanged. A `--fix` mode wired into that pass would flip that assertion RED.
  - `mint_blk` is defined roughly 28,000 lines away from the unit that uses it and has 15 uses that stay behind; treat it as an extraction blocker to resolve first, not an incidental helper.
  - The lint direction is the opposite of the usual worry: `run.sh` is never shellchecked, so extracting code into a module and promoting helpers into the harness makes it **newly lint-gated** at `--severity=warning` in a CI job separate from `lib + python tests`. A green suite does not surface those warnings.
  - A module may not self-skip — `run-module.sh` overrides `skip` to a fatal — so any host-conditional behavior in the extracted unit must fail or pass, never skip.
  - The focused runner confines `TMPDIR` to an owned scratch root and supervises the module in a process group; extracted fixtures that assume the monolith's temp-directory conventions need re-anchoring to the boundary root.
  - Focused-module logs accumulate unpruned under `.devflow/tmp/test-module-logs/`.
  - Two of the five existing modules carry zero issue-labelled assertions, so the attribution arm cannot reach them; this is a stated limit of label-derived ownership, not a defect to design around.

The01Geek:
## Summary
- Extracts the installer / workflow-wiring coverage out of the 51,000-line `lib/test/run.sh` monolith into a registered focused module, so iterating on `scripts/install-gh-wrapper.sh` or the workflow install steps costs one `run-module.sh` invocation instead of a full-suite run.
- Makes the coverage map's `run_sh_blocks` `owner` field mean something: it was the constant `unmodularized` on every entry, so the artifact documented as "a ranked to-do list for future extractions" could not answer the question it exists to answer.
- Replaces the map's hand-maintained half with a derivation the guard owns and enforces, plus a hand-invoked `--fix` that keeps the new per-PR obligation cheap.

## Changes

**Focused module**: `lib/test/modules/installer-wiring.sh` carries the `#487`/`#491`/`#533`/`#544`/`#599`/`#690` coverage, extracted as the **syntactic** unit (the `for _wf487` loop header through the `rm -rf "$D533"` teardown) rather than a slice cut at the first and last labelled assertion, which would have carried an orphan `done` and an unassigned loop variable. It re-derives `WF` from the harness-provided `LIB` — both runners execute a module body under `set -u`, so a verbatim extraction would have aborted on its first statement. Registered at `minimum_assertions: 118` (the observed executed count from the over-floor probe), inventoried, and added to the explicit `ci.yml` shellcheck list.

**Harness promotion**: `mint_blk`, `probe_tmp` and `probe_assert` move from `lib/test/run.sh` into `lib/test/module-harness.sh`. Promoted, not copied — uses of all three remain in the monolith, so a copy would have created an uncoupled mirror of load-bearing logic. A test pins exactly one definition tree-wide.

**Fixture isolation**: `module-harness.sh` now clears an inherited `DEVFLOW_GH` before any module body is sourced, extending to focused runs the isolation the monolith's preamble already guaranteed. A leaked override outranks every fixture-local PATH stub *silently*, so the new test drives the AC's own observable rather than the implementation.

**Coverage-map guard**: `lib/test/coverage_map_guard.py` gains one shared label derivation used for both `run.sh` and every module, anchored on assertion-name position so a `#NNN` in a comment derives nothing. It covers the namespaced harness API and module-private assertion wrappers — including the local-variable hop `local name="$1"; assert_eq "$name" …` that `capability-profiles.sh` uses. Arm 9 fails RED on a `run.sh` label with no map entry and on a fully-extracted label whose entry is absent or still says `unmodularized`; a *partially* extracted label correctly keeps `unmodularized`, because one `owner` string cannot describe split coverage. `evaluate` stays pure and never-raising: `main()` performs the reads and injects them, mirroring the existing read-error keyword convention, so all 35 pre-existing call sites pass unedited.

**`--fix` mode**: a hand-invoked repair that makes the map satisfy the arm. Deliberately absent from the batched generated-artifact pass's `argv`, so the `coverage-map-ratchet` row stays `by-hand` and its `#619 A3` byte-unchanged assertion stays green.

**Docs**: `CONTRIBUTING.md` item 6 and the `docs/DEVFLOW_SYSTEM_OVERVIEW.md` coverage-map paragraph no longer claim the `run_sh_blocks` half is hand-maintained or that mechanical scanning is a non-goal. A new *Coverage-map block ownership* subsection under "Running the tests" puts the new per-PR obligation where an assertion author reaches it, not only inside the module-authoring checklist.

## Resolves
Resolves #695

## Test Plan
- [ ] `lib/test/run-module.sh installer-wiring` exits 0 and reports 118 passed, 0 failed.
- [ ] `lib/test/run.sh` passes with zero failures, and the `#591 coverage-map guard: shipped tree + map is clean` assertion reports rc 0.
- [ ] `python3 lib/test/test_coverage_map_guard.py` and `python3 lib/test/test_module_runner.py` both pass.
- [ ] `shellcheck --severity=warning -e SC1091 lib/test/modules/installer-wiring.sh lib/test/module-harness.sh` is clean.
- [ ] `python3 lib/test/coverage_map_guard.py .` exits 0, and running `--fix` twice leaves `lib/test/modules/coverage-map.json` byte-unchanged the second time.
- [ ] `python3 lib/test/regenerate-artifacts.py` exits 0 with `coverage-map-ratchet` clean and the map byte-unchanged.

## Visual Changes
N/A — repo-internal test infrastructure.

## Breaking Changes
None. No `lib/test/` path ships to a consumer repository through any channel (`install.sh`'s workflow copy loop names no `lib/test/` file), so no adopter is affected and the `install.sh`-versus-`devflow_version` two-artifact skew cannot arise.


The01Geek:
**Deferral record (receiving-code-review, PR #697)**

**Suggestion #2 — deferred (not fixed this pass):** promoted helpers `mint_blk` / `probe_tmp` / `probe_assert` have no behavioral unit tests of their own failure logic.

- **WHY:** As the review itself notes, this is a **pre-existing** gap — these helpers were untested in `run.sh` before this PR merely *relocated* their definitions into `lib/test/module-harness.sh`. It is not a regression this PR introduces, and its severity (Suggestion, `test_gap`) is below the correctness-defect bar. Adding behavioral failure-logic tests for three bash harness helpers is out-of-scope new coverage, not a fix to the extraction this PR performs.
- **REVISIT:** when `module-harness.sh` next changes the failure/return-code logic of any of these three helpers, or a defect in them is actually observed — at which point their behavioral coverage should be added in that change.

**Addressed this pass** (Important #1 + Suggestions #1, #3, #4, #5): line-continuation and `"$@"`-forwarding label derivation now unit-tested; `_forwarding_aliases` tightened to a balanced `${1}` so default-expansion forms (`${1:-default}`) no longer bind as forwarding aliases (+ test); `derive_labels` and `_write_map` docstrings corrected. All new guard tests mutation-checked RED for the behavior they pin.

PR #734, opened 2026-07-23
Make lib/test/run.sh lintable, lint it in CI, and guard the lib/test lint carve-out

### LINKED ISSUE #717: Make lib/test/run.sh lintable, lint it in CI, and guard the lib/test lint carve-out
## Problem Statement

`lib/test/run.sh` is the repo's primary merge gate and its largest shell file (50,984 lines), and it has never been analyzed by shellcheck. Two shellcheck **parse** errors abort the analysis near the top of the file, and CI never invokes shellcheck on it at all. So every contributor and every autonomous `/devflow:implement` run that adds shell logic to `run.sh` ships that logic unlinted, and a reviewer reading a green CI check cannot tell. A recent proposal to add background-job and status-file logic to `run.sh` was rescoped partly because of this.

The same defect exists one level up: `.github/workflows/ci.yml` excludes `lib/test/` from the shellcheck glob and then re-adds a **hand-maintained** explicit file list. Nothing guards that list, so any new `lib/test/*.sh` that is not manually added silently ships unlinted. `run.sh` is that gap's largest instance, not its only one.

## Current Behavior

Verified: `shellcheck --severity=warning -e SC1091 lib/test/run.sh` exits 1 and reports exactly one location before stopping:

```
In lib/test/run.sh line 2554:
  if ! [ "$count" "$op" "$bound" ] 2>/dev/null; then
       ^-- SC1073 (error): Couldn't parse this test expression.
                   ^-- SC1072 (error): Expected comparison operator (don't wrap commands in []/[[]]).
```

SC1072/SC1073 are parse errors, so shellcheck aborts and analyzes nothing beyond that point. A sibling construct at line 2654 (`[ "$mut_count" "$op" "$bound" ] 2>/dev/null; breach_rc=$?`) is the second instance and is never reached.

Both constructs are correct bash and intentional. They live in `assert_count_red_under()`, whose signature is `name start end pattern op bound mutation [file]` — the integer comparison operator is a caller-supplied argument, which shellcheck's parser cannot express. `# shellcheck disable=` directives do **not** suppress parse errors, so a structural rewrite of both sites is required before the file can be analyzed at all.

Verified: `.github/workflows/ci.yml`'s `lint (shellcheck + actionlint + ruff)` job runs three shellcheck invocations — a `git ls-files '*.sh' | grep -v '^lib/test/'` glob, an explicit list naming `module-harness.sh`, `run-module.sh`, `summary.sh` and the six `lib/test/modules/*.sh` files, and a third for `describe-budget-delta.sh`. `lib/test/run.sh` appears in none of them; in `ci.yml` it appears only in a comment and in the `bash lib/test/run.sh` test step.

## Desired Behavior

`lib/test/run.sh` parses cleanly under shellcheck, reports zero findings at `--severity=warning`, and is linted by a dedicated CI job on every pull request. A new `lib/test/**/*.sh` file that is neither linted by CI nor recorded in a declared exempt set turns the test suite RED at the desk.

## User Impact

Contributors and autonomous implement runs get the same linting feedback on `run.sh` that every other shell file in the repo already gets, and a green CI check becomes an honest signal about it. The carve-out guard means the next `lib/test/` file added cannot repeat this silently.

## Technical Context

- **Relevant Classes/Files**
  - `lib/test/run.sh` — `assert_count_red_under()` and its two parameterized-operator test expressions (step 3's real-file bound check and step 8's mutated-file bound check), plus the existing self-test probes that drive the helper's contract arms.
  - `.github/workflows/ci.yml` — the `lint (shellcheck + actionlint + ruff)` job's `ShellCheck (lib + scripts)` step, and the new job this issue adds.
  - `lib/test/pin-corpus-lint.py` — documents `assert_count_red_under` in prose as deliberately excluded from its `HELPERS` set while included in its mutation-taking set.
  - `CLAUDE.md`, `docs/DEVFLOW_SYSTEM_OVERVIEW.md`, `CONTRIBUTING.md` — carry statements about the lint scope that this change falsifies (see Documentation Needed).

- **Architecture Alignment** — the guard is an ordinary `lib/test/run.sh` assertion, matching the repo's existing pattern of asserting workflow properties from the suite. The operator rewrite preserves the helper's existing validation structure: `assert_count_red_under` already validates `op` against the closed set `-eq -le -lt -ge -gt` in a `case` statement, and validates `bound` as a non-negative integer, before either comparison runs.

- **Dependencies** — `shellcheck`. The `Bash(shellcheck:*)` grant in `devflow_implement.allowed_tools` covers the in-env verification commands below; it does not govern the CI workflow step. Pinning the CI analyzer adds one new CI-side dependency: a version-pinned marketplace action that installs the named `shellcheck` release, referenced at a pinned major tag, matching how `ci.yml` already references `raven-actions/actionlint@v2` and `actions/checkout@v6`. When that install step fails, the job fails; it does not fall back to the runner image's build, so a red job never means an unknown analyzer ran.

- **Data/Schema Considerations** — none; no persisted state changes.

- **Cross-layer Impact** — repo-internal only. Nothing under `skills/`, no workflow shipped by `install.sh`'s copy loop, and no vendored artifact changes, so there is no `install.sh`-versus-`devflow_version` install-channel skew. The read-only `review` capability profile and the `devflow-implement` profile are both untouched: no skill or phase fence invokes a new helper.

**Verified measurements** (this repo, on `main` at commit `607ec800`, using a scratch copy of `run.sh` with only the two operator splices literalized so the file parses):

- Verified: `wc -l lib/test/run.sh` reports 50,984 lines.
- Verified: `grep -n '"$op"' lib/test/run.sh` returns exactly seven occurrences, every one inside `assert_count_red_under` — the `case` validator, its FAIL `printf`, the two comparison splices, two diagnostic `printf` calls, and the PASS `printf`. No other function reads `op`.
- Verified: `shellcheck --severity=warning -e SC1091 -f gcc` over the patched full-length copy completed in **609 seconds** of wall clock, exit 1, with **62 findings**, of which three are severity `error`. By code: SC2034 nineteen, SC1007 ten, SC1010 eight, SC2319 four, SC1090 four, SC1083 four, SC2010 three, SC1087 three, SC2215 two, SC2069 two, SC2155 one, SC2043 one, SC1112 one.
- Verified: shellcheck's runtime is superlinear in file length on this file — the same invocation over `head -n` prefixes took 1 s at 3,000 lines, 2 s at 6,000, 6 s at 12,000, and 73 s at 24,000. This is why the lint gets its own job rather than being appended to an existing one.
- Verified: `git ls-files 'lib/test/*.sh' 'lib/test/**/*.sh'` enumerates the tracked population, and partitions cleanly. Every tracked `lib/test/**/*.sh` file outside `lib/test/fixtures/` is named in `ci.yml`'s explicit list except `lib/test/run.sh` itself; everything under `lib/test/fixtures/` is unlinted, including deliberately-malformed corpora (`adversarial-*.sh`, `violation-*.sh`, `planted.sh`) that must stay that way. The exempt set is therefore the `lib/test/fixtures/` prefix, and its complement is exactly the set the guard requires CI to cover. The explicit list grows over time — a module was added to it while this issue was being drafted — which is precisely why the guard is needed and why no figure here states its size.
- Verified: a naive membership test that greps `ci.yml` for a path reports `lib/test/run.sh` as covered, because the file is named there in a comment and in the `bash lib/test/run.sh` test step. The guard's derivation must read the shellcheck invocations specifically, which is why the comment-only fixture below is a required case rather than a nicety.

**Mechanism dependencies this change relies on:**

- `shellcheck` exits non-zero when it emits any finding at the configured severity — the property the new CI job's pass/fail rests on. Verified: the invocations above exited 1 whenever findings were present.
- `# shellcheck disable=<code>` directives suppress the non-parse codes this change applies them to. Verified: a minimal fixture carrying a curly apostrophe reports SC1112 without a directive and exits 0 with `# shellcheck disable=SC1112` above it; a fixture invoking `git rev-parse refs/heads/main^{tree}` exits 0 with `# shellcheck disable=SC1083`. This is the property the disposition rule rests on, and it is distinct from the parse-error case: SC1072/SC1073 abort before any directive is consulted, which is why those two sites need a structural rewrite instead.
- The CI runner's shellcheck version is not pinned by this repo today. Verified: `.github/workflows/ci.yml` carries the comment `shellcheck ships on ubuntu-latest runners — no extra action needed`, so the analyzing build rolls forward with the runner image.

## Acceptance Criteria

- [ ] `assert_count_red_under()`'s two parameterized-operator test expressions are replaced by calls to a single new integer-comparison helper in `lib/test/run.sh` that dispatches on `op` through a `case` statement with one arm per member of the closed set `-eq -le -lt -ge -gt` — exactly these five, complete by construction — each arm running a literal-operator `[ ]` test and returning its exit status.
- [ ] That helper's default `case` arm returns status 2, so an operator outside the closed set produces an errored comparison status rather than a satisfied one.
- [ ] Step 3's real-file bound check preserves its current fail-closed direction: a comparison status of 2 or greater routes to the `BOUND-VIOLATED-ON-REAL-FILE` arm.
- [ ] Step 8's mutated-file bound check still captures the comparison's own exit status and treats every value other than exactly 1 as `BOUND-NOT-BREACHED`, so neither a satisfied bound nor an errored comparison reaches the PASS arm.
- [ ] Every contract FAIL arm of `assert_count_red_under`, including its `INVALID-OP` and `INVALID-BOUND` arms, still emits its existing verdict-and-token pair unchanged, proven by the helper's existing self-test probes passing.
- [ ] `shellcheck --severity=warning -e SC1091 lib/test/run.sh` parses the whole file — it reports no SC1072 and no SC1073.
- [ ] `shellcheck --severity=warning -e SC1091 lib/test/run.sh` exits 0. Every finding is resolved at its site: a finding whose flagged construct is intentional carries a per-line `# shellcheck disable=<code>` directive with a one-line reason comment immediately above it, and every other finding is fixed in the code.
- [ ] The workflow declares the `shellcheck` version once as a workflow-level value, and installs that pinned release through a version-pinned marketplace action rather than using the runner image's build.
- [ ] Both the new job and the existing `lint (shellcheck + actionlint + ruff)` job's `ShellCheck (lib + scripts)` step run that same pinned release, so a single analyzer version covers every directly-invoked shellcheck in CI.
- [ ] The workflow fails the job when the pinned install step fails, rather than continuing against the runner image's build.
- [ ] `.github/workflows/ci.yml` defines a job, separate from the `lint (shellcheck + actionlint + ruff)` job and with no `needs:` dependency on it, whose only work is `shellcheck --severity=warning -e SC1091 lib/test/run.sh`.
- [ ] `lib/test/run.sh` contains an assertion that reads `.github/workflows/ci.yml`, derives the set of `lib/test/**/*.sh` files CI lints, and fails naming the offending path when any tracked `lib/test/**/*.sh` file is neither in that set nor under the declared exempt prefix `lib/test/fixtures/`.
- [ ] The assertion's derivation accounts for the glob invocation's `lib/test/` exclusion expression: when that expression is not the exact form the assertion recognizes, the assertion fails closed rather than deriving coverage from the explicit list alone.
- [ ] That assertion fails closed with a distinct diagnostic naming `.github/workflows/ci.yml` when it cannot locate the shellcheck invocations in the workflow file, rather than reporting full coverage.
- [ ] The assertion's declared exempt set is the single prefix `lib/test/fixtures/` and no other path, and the assertion's diagnostic states which side of the partition the offending file landed on.
- [ ] `lib/test/run.sh` itself satisfies the guard by being present in the CI-linted set.
- [ ] `CLAUDE.md`, `docs/DEVFLOW_SYSTEM_OVERVIEW.md` and `CONTRIBUTING.md` no longer state that `lib/test/` is excluded from CI shellcheck, and no sentence added or edited by this change states a count of linted files.

## Implementation Notes

**Approach.** Add a small integer-comparison helper to `lib/test/run.sh` beside `assert_count_red_under`. It takes a left operand, an operator, and a right operand, and dispatches through `case "$op" in` with five literal-operator arms plus a default arm returning 2. Step 3 becomes an `if !` around a call to it; step 8 calls it and captures `$?` into `breach_rc` exactly as today. The load-bearing comment block above step 8 — which reasons explicitly about why the status is captured rather than inverted, and why an errored status must fail closed — is updated in the same edit to describe the helper call rather than the inline `[ ]`, and its reasoning is preserved.

With the file parsing, work the remaining findings under the single decided rule in the acceptance criteria: fix the construct at its site, and where the flagged construct is intentional, add a per-line `# shellcheck disable=<code>` directive carrying a one-line reason. Verified spot-checks of the finding sites show both dispositions are genuinely needed, and that a blanket fix-everything-in-code policy would introduce defects — see Potential Gotchas for the sites that establish this.

Then add the CI job and the suite guard. The guard is a best-effort reader of `.github/workflows/ci.yml`, so it fails closed on every shape it cannot interpret.

**Code Patterns.** The `case`-dispatch shape mirrors the two validators already at the top of `assert_count_red_under`. The guard mirrors the suite's existing workflow-property assertions and `CLAUDE.md`'s best-effort-parser discipline. Per `CLAUDE.md`'s ordinal-count gotcha, neither the guard's comments nor the updated docs state a count of linted files.

**Testing Strategy.**

The change has automated test boundaries at two levels: the helper's exit-status contract (driven by `assert_count_red_under`'s existing self-test probes inside `lib/test/run.sh`), and the new guard's own behavior (driven by synthetic `ci.yml` fixtures through the suite's existing scratch-file probe machinery).

`governing conventions consulted: CLAUDE.md (best-effort-parser adversarial input-shape matrix; ordinal-count rot; behavioral-fix-pin rule), CONTRIBUTING.md (the lib/test module lint-list step)`

*Operator-helper assertions.* Each fails first for the right reason — the helper does not exist yet — and each maps to the operator acceptance criteria above:

- One assertion per member of the closed operator set, asserting the helper returns 0 when the comparison holds and 1 when it does not.
- An assertion that an operator outside the set returns 2.
- An assertion that `assert_count_red_under` still reaches `BOUND-VIOLATED-ON-REAL-FILE` when the real count violates the bound, and `BOUND-NOT-BREACHED` when the mutation leaves the bound satisfied — the two arms that consume the two rewritten comparisons.
- The existing self-test probes for the helper's remaining contract arms run unchanged and stay green; they are the regression evidence that the rewrite preserved the arms it did not target.

*Guard assertions.* The guard reads a human-maintained YAML file, so its case matrix follows `CLAUDE.md`'s best-effort-parser discipline adapted to that input type. Each of these is a synthetic `ci.yml` fixture:

- A well-formed workflow whose lint step covers every non-fixture `lib/test/**/*.sh` — the guard passes.
- A well-formed workflow missing one such file — the guard fails naming that exact path. This is the planted-defect positive control for the guard's coverage claim: the defect the guard exists to catch is planted, and the guard is proven to fire on it.
- A workflow containing a new `lib/test/fixtures/` file not named anywhere — the guard passes, proving the exempt prefix works.
- A workflow whose shellcheck invocations are absent entirely — the guard fails closed naming `.github/workflows/ci.yml`, rather than reporting coverage.
- An empty file, and a file that is not valid YAML — the guard fails closed with the same diagnostic.
- A workflow that names a `lib/test/` path inside a comment only — the guard does not count it as covered.
- A workflow whose glob invocation has had its `lib/test/` exclusion removed or narrowed — the guard fails closed rather than reporting the explicit-list-only coverage as complete.

*Verification commands.* Each is a leading-token invocation of a command already granted in `devflow_implement.allowed_tools`:

- `shellcheck --severity=warning -e SC1091 lib/test/run.sh` — must exit 0. The finding-count measurement command is `shellcheck --severity=warning -e SC1091 -f gcc lib/test/run.sh` piped to a line count.
- `lib/test/run.sh` — the full suite, run as the leading token, must report zero failures and disclose any skipped checks.

**Documentation Needed.**

- `CLAUDE.md` — correct the Commands block's shellcheck line, which shows only the first of the workflow's shellcheck invocations and so does not reproduce the CI gate; correct the `#248` bullet's claim that `lib/test/` is excluded from CI shellcheck; correct the stale singular framing that `lib/test/summary.sh` was the one explicit carve-out.
- `docs/DEVFLOW_SYSTEM_OVERVIEW.md` — correct the same stale singular carve-out framing and record the new guard.
- `CONTRIBUTING.md` — extend the module-authoring step so it describes the guard that now enforces the lint list.

**Potential Gotchas.**

- The new CI job gates merge only once a human adds it to the branch-protection required-check set. The existing required check is `lib + python tests`; adding a second required check is a repository-settings action outside any pull request, and until it happens the new job reports without blocking.
- A ten-minute job is a long pole. The measured 609 seconds is on this hardware; GitHub-hosted runners are slower, so budget accordingly.
- `assert_count_red_under` is load-bearing for `CLAUDE.md`'s behavioral-fix-pin rule and for issue #666's mutation-routing gate. `lib/test/pin-corpus-lint.py` names it in prose in both its deliberately-excluded-from-`HELPERS` note and its mutation-taking set; check whether the rewrite changes anything that file asserts.
- Every comment in `run.sh` that describes the two rewritten comparisons is a coupled site and is edited in the same change, per `CLAUDE.md`'s coupled-invariant rule.
- **The single SC1112 unicode-quote finding must not be retyped to ASCII.** Verified: it sits at a pin literal in `run.sh` that quotes skill prose containing a real curly apostrophe. Retyping it would silently break the pin against its target. This site takes a disable directive.
- **The SC2069 findings are deliberate.** Verified: both sites are `2>&1 1>/dev/null` — the standard idiom for capturing stderr while discarding stdout, which shellcheck flags as a probable ordering mistake. These take disable directives, not reordering.
- **The SC1083 findings are not `gh api` placeholders.** Verified: they are `git rev-parse refs/heads/<branch>^{tree}` invocations, where the literal braces are git revision syntax. These take disable directives.
- **The SC1087 findings are real ambiguity.** Verified: each is a `$var` immediately followed by `[` inside a `grep -E` pattern, which bash may read as an array subscript. These are fixed in the code by brace-wrapping the expansion.
- **The SC1010 findings come from a test fixture whose slug is the literal string `do`.** Verified: a block of `python3 "$IAS" <verb> do` calls uses `do` as an audit-state slug argument, and shellcheck reads the bare `do` as a misplaced loop keyword. Quoting the argument resolves it without renaming the fixture.
- The exempt set is closed by today's file census, not by construction: nothing structurally prevents a future deliberately-unlintable file from landing outside `lib/test/fixtures/`. When one does, the guard fires (fail-closed, the safe direction). The intended remedy is to move that file under `lib/test/fixtures/`, and amending the declared exempt set is a considered change made in the same commit as the file that needs it.
- `actionlint` bundles its own shellcheck build for analyzing workflow `run:` blocks, and this change does not pin that one. It analyzes a different surface (workflow YAML, not the repo's `.sh` files), so the two never disagree about the same construct; the residual is accepted and named here rather than left implicit.
- The runner image's shellcheck rolls forward on GitHub's schedule. Without the version pin the acceptance criteria require, an image bump that adds a check or promotes one to `warning` severity would turn the new job red across every open pull request at once, with no diff that caused it and no local reproduction — on the repo's largest shell file. The pin makes a linter upgrade a deliberate, reviewable commit. Because both directly-invoked shellcheck steps read the same declared version, that upgrade is a one-line change.
- Adding disable directives changes line numbers throughout `run.sh`. Any assertion or doc referring to a `run.sh` line number will drift; the repo already forbids `path:line` references in docs.

The01Geek:
## Summary
- Make `lib/test/run.sh` pass `shellcheck --severity=warning -e SC1091` cleanly, so the suite's own harness is held to the same lint bar as the code it checks.
- Lint `run.sh` in CI via a dedicated job (it is superlinearly slow on that one large file), using a workflow-pinned ShellCheck release shared with the existing lint job.
- Add a fail-closed guard (`lib/test/lint-carveout-guard.py`) that reconciles the `lib/test/` shellcheck carve-out: every tracked `lib/test/**/*.sh` file must be CI-linted or under the single exempt prefix `lib/test/fixtures/`, else the suite goes RED naming the offending path.

## Changes
**`lib/test/run.sh` (lintability)**: Resolved every ShellCheck finding to exit 0 — fixed in code where the construct was accidental, and added per-line `# shellcheck disable=<code>` directives with one-line reasons where intentional. The two `[ "$n" "$op" "$bound" ]` splices whose variable comparator triggers unsuppressable SC1072/SC1073 parse errors are replaced by a structural `_int_cmp` helper that dispatches the operator through a `case` (one arm per member of the closed set `-eq -le -lt -ge -gt`, default arm returns 2), preserving the fail-closed comparison direction at both call sites.

**CI (`.github/workflows/ci.yml`)**: A workflow-level `SHELLCHECK_VERSION` pin, installed via a pinned release download (with `set -euo pipefail` + `curl -f` so a bad install fails the job) in both the existing `lint` job and a new dedicated `shellcheck (lib/test/run.sh)` job (separate, no `needs:`), so a single analyzer version covers every directly-invoked ShellCheck in CI.

**Carve-out guard (`lib/test/lint-carveout-guard.py`, new)**: Reads `ci.yml`, derives the set of `lib/test/**/*.sh` files CI lints, and fails naming the offending path when a tracked `lib/test` file is neither CI-linted nor under `lib/test/fixtures/`. It is a best-effort YAML reader that fails CLOSED on every shape it cannot interpret (unreadable/empty/non-YAML `ci.yml`; no `jobs:` mapping; no locatable shellcheck invocation; a `git ls-files '*.sh'` glob whose `lib/test/` exclusion differs from the exact recognized form — whether written backslash-joined or across POSIX trailing-pipe continuations). Driven from `run.sh` over the real tree plus a matrix of synthetic `ci.yml` fixtures.

**Docs**: `CLAUDE.md`, `CONTRIBUTING.md`, and `docs/DEVFLOW_SYSTEM_OVERVIEW.md` reframed the carve-out as singular/guarded (shipped `lib/test/` scripts are linted via an explicit list + the dedicated `run.sh` job + the guard; only `lib/test/fixtures/` is exempt); no added/edited sentence states a linted-file count.

## Resolves
Resolves #717

## Test Plan
- [ ] `shellcheck --severity=warning -e SC1091 lib/test/run.sh` exits 0 (verified locally on ShellCheck v0.11.0).
- [ ] `bash lib/test/run.sh` passes with no failures and no skips (verified: 12953 passed, 0 failed).
- [ ] `ruff check lib/test/lint-carveout-guard.py` passes.
- [ ] The guard fails closed on each malformed/narrowed `ci.yml` shape (trailing-pipe / grep-removed / no-jobs / non-YAML / narrowed-glob / comment-only) and passes over the real tree — exercised by the `#717 guard:` cases in `run.sh`.

## Visual Changes
N/A

## Breaking Changes
None — internal-only change (test infrastructure + this repo's own CI + dev-only docs); nothing ships to consumer repos, so no changeset or release note is warranted.


The01Geek:
**Parked (deliberate hold) — not stalled.**

This PR is complete and review-ready: `lib/test/run.sh` is shellcheck-clean at the desk (exit 0), the full suite passes (12953 passed / 0 failed), the carve-out guard fails closed correctly (a fail-open bug found in review was fixed), the branch is current with `main`, and the required **`lib + python tests`** check is green.

It is intentionally **not being merged yet** because of a single infrastructure blocker, tracked in **#739**: the dedicated `shellcheck (lib/test/run.sh)` CI job deterministically fails at ~1m53s (runner shutdown — almost certainly ShellCheck OOM-ing on the ~47k-line file; it completes fine at the desk with more RAM). Because DevFlow's auto-review waits for *all* other CI to be green (`require_ci_green`), that one un-completable, **non-required** job holds back the required `Devflow Review`, which blocks merge.

No code fix is pending here — `run.sh` genuinely passes shellcheck. Resolving #739 (split the file, use a larger runner, or decouple the run.sh lint from the review's CI-green gate) is what unblocks this. Leaving the PR open until then.

The01Geek:
Superseded by #739. The root-cause fix is to split `lib/test/run.sh` into shellcheck-able modules, which removes the single-giant-file job this PR adds (the one that OOMs the CI runner). #739 now owns the "lint run.sh in CI" goal and carries forward this PR's two keepers — the `_int_cmp` fix and `lib/test/lint-carveout-guard.py`. Recommend closing this PR once #739 lands; leaving it open for now as the reference for those two pieces.

The01Geek:
Closing as superseded. Thank you for the evidence — the failure this PR surfaced is what made the correct fix findable.

**What this PR proved.** The dedicated `shellcheck (lib/test/run.sh)` job died with `The runner has received a shutdown signal` in **6 of 6 runs across 5 distinct head SHAs** (1m29s–1m54s). That is a deterministic, commit-independent memory eviction, not a flake — and it is the measurement that ruled out the whole one-job-per-big-file approach.

**The actual fix.** ShellCheck's blowup is entirely its dataflow pass. `--extended-analysis=false` lints the whole unsplit file in **9.9 s at 1.15 GB**, costing 4 of 75 findings. No split, no dedicated job, no larger runner. Caveat that would have silently defeated it: `ubuntu-latest` ships ShellCheck 0.9.0, which has neither the flag nor the directive (both landed in 0.10.0) — so the binary must be pinned, which is exactly the `SHELLCHECK_VERSION` pinning mechanism this PR already built.

**Both useful pieces are carried forward:**
- `_int_cmp` — landed in #744, extracted rather than cherry-picked (this branch is 30 commits behind `main` and predates #736's python3 `_nonws_count` rewrite, so a whole-file take would have reverted it).
- `lint-carveout-guard.py` — carried into #745, where it now passes with no exemption beyond `lib/test/fixtures/`, because `run.sh` becomes genuinely linted. Its trailing-pipe fail-open fix and its `git ls-files` population survive unchanged.

Follow-ons: **#745** (lint `run.sh` with dataflow analysis disabled) and **#746** (modularization on its own merits). #739 is closed as superseded.

Issue #857, opened 2026-07-27
Route the review live-comment seed through a token-line helper so the #384 screens execute in cloud


The01Geek:
## Problem Statement

The review engine's live-progress-comment seed fence in `skills/review/SKILL.md` is a `case` + `if`/`elif` compound implementing three screens — S1 numeric guard, S2 readability precheck, S3 rc-2-with-empty-stderr discriminator — added as the issue #384 fix. The cloud review matcher refuses that compound outright, so the screens never run there.

The everyday consequence is a silent degradation: the seed fails, `$WP` is never set, and Phase 4.4 posts the full report inline in `gh pr review` with no live progress comment at all. Reviewers lose the running narrative the engine advertises, and nothing reports why.

The smaller and more dangerous tail is that the engine improvises a replacement shape. Sometimes that improvisation preserves the stderr evidence S3 needs; sometimes it does not, and the run reaches the `create` arm unable to tell a clean absence from an interpreter-level exit 2 — the duplicate-workpad-comment failure #384 exists to prevent.

Maintainers reading `skills/review/SKILL.md` cannot see any of this. The fence is present, its pins are green, and the desk lint passes it; the refusal is visible only in cloud transcripts.

## Current Behavior

**Verified** — swept the 299 most recent completed `Devflow Review (auto-trigger)` runs; 25 had a non-skipped engine job (`review / run`), spanning PRs 832, 833, 835, 836, 837, 840, 841, 844, 845, 846, 847, 848, 849, 851, 852. All 25 `claude-execution-transcript-*` artifacts were live and were downloaded and parsed.

- The shipped fence executed as written in **0 of 25 runs**. Screen S2 appears in no transcript in the sample.
- It was attempted in 8 of those runs and refused in 8 of 8, each time with the byte-identical harness string `Contains shell syntax (string) that cannot be statically analyzed`. Example: run `30222310785` (PR #847, head `e11436bcfd203728251ab8478678e6eb280d80ef`), transcript entry idx 113, `is_error: true`.
- In the remaining runs the engine did not attempt the fence at all and emitted an ad-hoc shape instead.
- 21 of 25 runs reconstructed an S3-equivalent by hand via the permitted `… 2>rv-id.err; echo "rc=$?"; cat rv-id.err` shape.
[...]

PR #880, opened 2026-07-28
create-issue: re-check Verified: premises at implement time

### LINKED ISSUE #868: create-issue: a Verified: premise is never re-checked at implement time, so an issue can prescribe an unimplementable mechanism (worked case: #857)
## Problem Statement

An issue drafted by `/devflow:create-issue` records its evidence as `Verified:` bullets quoting source sentences. Those bullets are true **when the issue is drafted**. Nothing re-checks them when the issue is later implemented, and nothing detects when a premise has since become false.

Issue #857 is the worked case. It was drafted against a tree in which its premises held; by the time PR #864 implemented it, **three of them were false**, and two of its acceptance criteria were literally unimplementable — a prescribed mechanism that the suite now rejects by construction.

## Current Behavior

**Verified** — the three stale premises in issue #857, each checked against `main` at `c32fec4a`:

1. #857 prescribes flipping two `disposition` values in place in `.devflow/logs/residual-prose-retirement-manifest.tsv`, citing commit `b430c9b8` as precedent. `lib/test/pin-corpus-lint.py`'s `load_retired_wording_literal_keys` now treats every path in `_RETIREMENT_MANIFEST_SPECS` as byte-frozen against the merge base:

   ```python
   if head_bytes != base_bytes:
       raise InfrastructureError(
           f"historical retirement manifest changed since merge base: {path}"
       )
   ```

   There is **no authorization escape** on that branch (unlike the adjudication table, which has one). Observed empirically on PR #864's head `858520f`: the gate exited **2** with exactly that message. The `b430c9b8` precedent predates the freeze.

2. #857 states that deletion from `lib/test/pin-corpus-adjudications.tsv` is barred because `test_adjudications_keep_the_base_table_as_an_exact_prefix` requires a frozen prefix. **That test does not exist** — `grep -rn "exact_prefix" lib/test/*.py` matches nothing.

3. #857 describes tombstone rows as "the shape already present in `lib/test/pin-corpus-adjudications.tsv`". **`grep -c '^tombstone:'` returns `0`** — there are none.

The consequence in PR #864: the prescribed manifest edit failed the `#810` gate, and its `InfrastructureError` raised *first*, masking a second failure (an unauthorized adjudication delta) until the manifest was reverted. Two review rounds and roughly an hour of wall-clock went to rediscovering that the spec was unimplementable.

**Verified** — no re-verification obligation exists today. `.devflow/prompt-extensions/create-issue.md` defines the `Verified:` bullet discipline at drafting time (its "Every entry is a **`Verified:` bullet quoting the sentence from the file**" rule and the consumers-axis evidence floor). A search of that file for re-check / re-verify / stale-premise guidance returns **no matches**. Neither `skills/implement/SKILL.md` nor its phase files obligate an implementing run to re-check a `Verified:` bullet before building on it.

## Why this matters beyond one issue

A `Verified:` bullet is the single most load-bearing line in a DevFlow issue: it is what licenses an implementing run to skip its own investigation. A stale one is strictly worse than no bullet at all — it converts "go and check" into "this was already checked", and the run builds confidently on a false premise. The failure is silent and only surfaces as a red gate late in the PR, or not at all.

This is a **drafting-and-implementation-path** defect, not a defect in issue #857.

## Proposed Direction

Deliberately left open — the right mechanism is a design question, and the options differ a lot in cost:

- an implement-phase obligation to re-run each `Verified:` bullet's own check before relying on it, failing closed to ordinary investigation when a quote no longer resolves;
- a drafting-time requirement that each bullet record *how* to re-derive it (the command / file+symbol), so re-checking is mechanical rather than a re-investigation;
- a staleness signal keyed on the drafting revision, surfaced when the issue's cited files have moved since.

## Notes / constraints

- The gates in the worked example (`#810`, the pin-corpus lint) are DevFlow-internal, but the `Verified:` bullet discipline **ships to consumer repos** via `skills/create-issue/`. Any fix must not assume DevFlow-internal gates exist in a consumer tree.
- A prose-only fix in `skills/create-issue/**` or `.devflow/prompt-extensions/create-issue.md` requires **no new tool grant**, so it is not subject to the trigger-time grant-timing bootstrap and is live in its own implementing run.
- Per issue #810, do **not** discharge this with a documentation-presence pin. If a mechanism is added, it needs an executable behavioral test; if the outcome is prose-only, that absence of coverage is a decision to record explicitly (the #843 pattern).

## Evidence trail

PR #864 discussion, in particular https://github.com/The01Geek/devflow-autopilot/pull/864#issuecomment-5096223056 (the full stale-premise analysis with code citations).

The01Geek:
## Summary
- A `Verified:` bullet is what licenses an implementing run to skip its own investigation, and nothing re-checked one after the issue was filed — so a premise that had since become false converted "go and check" into "this was already checked". Issue #857 is the worked case: three premises were false by the time #864 implemented it, and two acceptance criteria were unimplementable as prescribed.
- Both limbs of the fix ship together, served by one new helper: the drafting side must give each bullet a self-contained re-derivation handle, and `/devflow:implement` re-checks every recognized bullet before building on it.
- Refuting is deliberately the hardest verdict the helper can reach, because a refutation makes the run *discard* the premise and file issue-accuracy feedback against the issue. Measured against issue #857's live body: 5 refutations before the review loop's fixes, all of them false; 0 after.

## Changes

**New helper (`scripts/check-verified-premises.py`)**: Enumerates every `Verified:` bullet an issue body carries, classifies its re-derivation handle (`path-quote` / `path` / `quote` / `command` / `none`), and adjudicates it against the tree as `holds` / `refuted` / `unestablished`. Exit 0 = nothing refuted, 2 = at least one refuted, 3 = the measurement could not be established. Every terminating path prints a `VERIFIED_PREMISES` line, so its absence — not the exit code — is what a caller reads as "this did not run".

**Security boundary**: The issue body is third-party text, so the helper performs read-only file reads and nothing else. A command handle is *reported* for the caller to re-run under its own judgment and never executed; the module imports no execution or network module and calls no dynamic-execution builtin; a cited path that is absolute or escapes the repository root is refused rather than opened.

**Implement side (`skills/implement/phases/phase-1-setup.md`)**: Phase 1.6 gains Pass 6, which runs the helper over the §1.1 issue-body cache (no re-fetch) and routes four outcomes. A refuted premise is discarded in favour of investigation and recorded as `issue-accuracy` feedback — it does not Block the run. An unestablished measurement fails closed: every bullet is treated as unverified.

**Drafting side (`skills/create-issue/references/`)**: `issue-template.md` requires each `Verified:` bullet to carry the repository path plus the sentence quoted verbatim, or the exact command, and adds the matching drafting-checklist row. `step-3-5-steelman.md` states the obligation and its precedence against `check-claim-staleness`; `step-3-6-audit.md`'s pre-dispatch canonical write is where it executes, because that is the first anchor at which a canonical draft exists.

**Capability profiles (`lib/capability-profiles.json`)**: The helper is granted on the `implement` and `command` profiles; all generated workflow literals were regenerated from the manifest and the `review` lock is byte-identical.

**Tests (`lib/test/test_python_scripts.py`)**: Ordinary executable tests driving `main()` at its CLI boundary over real trees — no wording, presence, or documentation pins. Every guard was mutation-checked on a copy of the helper.

## Resolves
Resolves #868

## Test Plan
- [ ] `lib/test/run.sh` is green with a zero failure tally and no new skips
- [ ] `lib/test/test_python_scripts.py` passes, including the `#868` block
- [ ] `ruff check scripts/check-verified-premises.py` is clean
- [ ] `python3 lib/generate-capability-profiles.py --check` exits 0 and `lib/review-profile.tokens` is unchanged
- [ ] `python3 lib/test/regenerate-artifacts.py` reports all artifacts reconciled
- [ ] Running the helper over issue #857's real body reports 0 refutations (it reported 5, all false, before the review loop's fixes)

## Visual Changes
N/A

## Breaking Changes
None. The helper is new, and both consumers treat an unavailable invocation as non-blocking — the drafting-side check never blocks issue creation, and Pass 6's unestablished arm falls back to ordinary investigation.


The01Geek:
Addressed the REJECT driver in 916acb5f:

- **F1 (Important, REJECT driver)** — `--repo-root` `--help` now reads `default: the enclosing git working tree; unestablished (exit 3) if none is found`, matching `_default_root()`/`_run`'s fail-closed behavior.
- **F2 (Suggestion)** — the test comment now attributes the no-`--repo-root` invocation shape to create-issue Step 3.5 rather than implement Pass 6.

Not taken: S2 (aligning Step 3.5 to pass `--repo-root` — that would change the production shape the test above documents), S3/S4 (coverage + type-design improvements; no defect).

**Repo-policy gate correction:** the report states the `Writing-skills evidence:` marker "was not found on the linked issue #868 workpad (no workpad present)". The workpad comment does exist on issue #868 and carries the marker (`00:45:41 — Writing-skills evidence: trigger file skills/implement/phases/phase-1-setup.md; mode=subagent …`), plus a second routing record at `00:58:47`. The gate is discharged.

Verification: `ruff check` clean; focused `lib/test/test_python_scripts.py` → 2846 passed, 0 failed.

The01Geek:
Addressed the union of findings still open across the review iterations, in `b085daa3`. Threshold here is `suggestion`, so the Suggestion-level notes were in scope.

**Fixed**

- **F2 — `_ArgParser` exit-2 surfaces.** Confirmed against HEAD: the remap lived only in `error()`, so a direct `parser.exit(2)` — reachable from an argument action, a `type=` callable, or a mutually-exclusive group — emitted this module's REFUTED code for an invocation that adjudicated nothing. The remap now lives in `exit()`. Status 0 (`--help`) is deliberately left alone, and that narrowness is pinned too.
- **F3 / prior S5 — the path triple.** `(strength, bare, suffix)` is now a `CitedPath` NamedTuple, destructured by name at all six sites. Two same-typed `str` fields unpacked positionally meant a reorder would have swapped the refutation gate's comparand for a path with no test able to see it; it is an `AttributeError` now. The seven bare `2`/`3` exit literals are named constants.
- **F4/F5/F6 + prior S4/S3 — the test gaps.** All closed: `_default_root`'s nearest-of-two-`.git` selection, a cited target holding invalid UTF-8 (`errors='replace'`), `_resolves_inside`'s true branch on an in-tree `..`, quote-beats-command handle precedence, a locator-suffixed strong path whose quotation *misses* still refuting, typographic-dash folding in the holds direction, and non-overlapping in-order elided-fragment matching in both directions.
- **Prior S3 (changeset).** Reworded: Step 3.5 *requires* the helper to run over the assembled draft; it executes at the Step 3.6 pre-dispatch canonical write.

**F1 — continuation-line absorption: disclosed, not re-bounded.** The behavior is real and reproduces. The suggested alternative — bounding the span at a hard newline — would sever every soft-wrapped bullet from its own citations and refute *true* premises against files they did cite, a miss in exactly the direction this module trades away everywhere else, and markdown itself reads a lazy continuation as part of the same list item. So the correct disposition is the other one the finding offered: `parse_bullets` now carries a DISCLOSED RESIDUAL note stating the trade, why the bound is not `\n`, and the converse cost. Unpinned by construction — which of the two shapes an author meant is not decidable from the text. *Revisit if a real issue body is observed where a following-line citation drove a false `refuted`.*

**Deferred, with reasons**

- **Prior S1 — quote resolution searches every readable co-cited file.** This is the module's deliberate asymmetry direction (favour not-refuting), already stated at the `readable.values()` loop. *Revisit only if evidence contradicts that comment.*
- **Prior S2 — `normalize()` strips emphasis symmetrically.** Documented in `normalize`'s own docstring as one of three folds that each close a way a **true** premise would read as refuted. Tightening it moves the module toward false refutation. *Revisit if a false `holds` is observed that turns on emphasis alone.*
- **Prior S2b — Step 3.5's fence omitting `--repo-root`.** Previously declined; that shape is the production invocation `_default_root` exists for, and it is what the no-`--repo-root` tests document. Unchanged.
- **`typing.Literal`/`enum` for the `handle`/`state`/`strength` vocabularies.** Declined on a corrected premise: the review noted mypy is "already in the toolchain", but `git grep mypy` finds it only as a *granted tool token* in three workflow allowlists — there is no mypy config and no lint job runs it (the Python gate is `ruff`). `Literal` annotations would therefore be unenforced churn. I took the half that pays without a checker (the NamedTuple and the exit constants) and left the rest.
- **`manifest_version` 10→12.** Already assessed harmless by two agents; final state is consistent and `--check` passes.

**Verification:** 8 mutation checks, one per new guard, each run against a mutated *copy* (the working tree was never mutated) — all 8 killed, each for the reason its assertion pins, with the baseline confirmed green first. `ruff check` clean; focused `lib/test/test_python_scripts.py` → **2859 passed, 0 failed** (2846 before). Full `lib/test/run.sh` running now; result to follow.

Issue #827, opened 2026-07-26
Raise the Step 3.6 automatic re-audit ceiling from 1 to 2


The01Geek:
## Problem Statement

`/devflow:create-issue`'s Step 3.6 audit lifecycle funds exactly **one** automatic re-audit round. `scripts/issue-audit-state.py` sets `_MAX_AUTOMATIC_REAUDITS = 1`, so a run whose first audit returns `VERDICT: REVISE` gets one automatic re-audit and then falls through to `revise-then-evaluate-offer` — the point at which the skill stops re-auditing on its own.

The consequence is that **a draft not converging within two audit rounds cannot make further progress unattended.** That is a low bar for a non-trivial draft: the first round surfaces the substantive defects, the revision answering them is itself unaudited, and the second round is the first opportunity to check that revision. When the second round finds anything — including a defect the revision introduced — the automatic budget is already spent, before the loop has completed a single verified pass over a revised draft.

Raising the ceiling to 2 buys one more unattended round: enough to audit a revision and, if that round is clean, converge without a user decision.

**This is a budget change, not a mechanism change.** No verdict, no arm, no answer token, and no state field changes. The only behavioural difference is when the automatic budget runs out.

## Current Behavior

`_MAX_AUTOMATIC_REAUDITS = 1`. Two expressions read it: `next_action`'s `revise-and-reaudit` / `revise-then-evaluate-offer` selection, and `cmd_record_dispatch`'s derived automatic spend. A run with no accepted offers can open **2** rounds (the initial round plus one automatic re-audit); with the user cap, 5; with honoured final-byte passes, 8; a run whose passes degrade can inflate the funding sum to 11.

## Desired Behavior

`_MAX_AUTOMATIC_REAUDITS = 2`. A run with no accepted offers opens **3** rounds; the other totals become 6, 9, and up to 12. `query-next-action` answers `revise-and-reaudit` for a REVISE round while fewer than two automatic re-audits have been spent, and `revise-then-evaluate-offer` after — the answer keys off the spend counter, not off round ordinality, so a run whose middle round was a final-byte pass (which the `not final_byte_pass` guard excludes from the spend) still has automatic budget left. The answer set, the funding refusal's message, the user-round cap and the final-byte slot are unchanged.

## User Impact

**The population this helps** is runs that reach a second REVISE round on findings that trigger a user-round offer: today they stop and ask, and after this change they get one more unattended round first.

[...]

Issue #792, opened 2026-07-24
create-issue Step 3.6: make final-byte audit coverage a reported, non-substitutable lifecycle fact


The01Geek:
## Problem Statement

`/devflow:create-issue`'s Step 3.6 audit lifecycle never binds an audit round to the bytes that actually get filed, and never says so. Three explicitly sanctioned paths file bytes no auditor has ever seen:

- a `user-decline` override recorded at `step4-approval-after-exhausted-offer` — the documented path by which "the audit informs without blocking a determined filer";
- a run that converged `basis=resolution` (or `resolution-stale`) on the drafter's own self-verified fixes rather than an auditor's clean verdict;
- a `record-creation-epoch --draft-file` attestation whose comparand is deliberately the digest of the bytes actually posted, so a legitimate override filing attests `match` against bytes the audited round never covered.

The attestation proves **identity**; it has never proved **audit coverage**. Nothing on `query-summary`, `query-convergence` or `query-eligibility` answers the one question a maintainer needs before approving: *does the draft about to be filed carry a `VERDICT: FILE` from a round that saw these exact bytes?* A run can today report `attestation=match` beside bytes no auditor read, and the summary line the user reads before approving is silent on the difference.

## Current Behavior

**Verdict invalidation on edit exists — on one ground only.** `evaluate_eligibility` answers `eligible` on exactly two grounds: (a) a completed `VERDICT: FILE` round whose identity still holds — on a file-arm round, its recorded dispatch digest equals the current canonical-file digest — and (b) an explicitly recorded override that is still current. Ground (a) therefore already invalidates on any byte edit and answers `unaudited-revision`. Ground (b) bypasses the comparison entirely, and on embed/inline arms ground (a)'s identity test degrades to revision ordering, which the docstring discloses as "weaker than byte identity".

**`cap-reached` grants nothing.** `_OVERRIDE_KINDS = ('user-decline', 'cap-reached')`. `cmd_record_override` refuses a `cap-reached` record while `user_rounds_used < _USER_ROUND_CAP` ("a premature cap record would silently burn the remaining user rounds") and then pins the counter to the ceiling. It records that the ceiling was reached; it is not a grant of extra rounds. The only budget-restoring lever is `init --force`, which destroys the run's entire lifecycle record and is refused outright once a creation attestation is frozen.

[...]

Issue #705, opened 2026-07-22
Make canonical draft updates durable: stage the intended bytes, replace once, verify the replace


The01Geek:
## Problem Statement

`/devflow:create-issue` mandates **what** the canonical draft file must contain and **when** it must be written, but never states **by what transport**. The write mandate in `skills/create-issue/references/step-3-6-audit.md` names only the root-resolution recipe — `resolve-main-root.sh`, `mkdir -p`, title as a top `# ` heading — and no clause names a write tool, a patch format, an edit granularity, nor a durability requirement. A repo-wide grep of `skills/create-issue/` for `Edit tool`, `MultiEdit`, and `Write tool` returns zero hits.

That unconstrained space is the defect, and two analyzed sessions landed at its opposite ends. In one, a single large generated shell command carrying a full-draft rewrite was the only copy of the pending mutation; an interruption destroyed it and it had to be reconstructed from the conversation and replayed. In the other, an audit-heavy draft was mutated through 44 separate edit calls, so a multi-finding revision wave had 44 interruption points and 44 chances to leave the canonical file in a state that is neither the audited bytes nor the intended bytes.

The time cost is minutes. The correctness cost is larger and is currently undetectable: `record-revision` is recorded **after** the bytes are mutated and records nothing about them, so a wave interrupted partway can still be followed by a `record-revision` and a `record-resolution` naming findings that were never actually fixed. Nothing in the state owner can distinguish a fully-applied wave from a partially-applied one, because no durable record of the *intended* bytes exists to compare the file against.

The drafting orchestrator of this skill is who hits this, and the user whose issue is filed is who bears the result.

## Current Behavior

The canonical draft file `<bound-root>/.devflow/tmp/issue-draft-<slug>.md` is mutated at three sites — the Step 3.6 pre-dispatch write (including every re-dispatch), the Step 4 sub-step 2 presentation write, and the Step 4 sub-step 4 iterate-on-feedback overwrite. Each site states the delete-first, write-fresh, confirm-landed discipline and the root binding, and none states how the bytes get there.

"Confirm the write landed" is discharged today by observing that the write command succeeded and the path is readable. That establishes *a* file exists; it does not establish that the file holds the bytes the run intended. The observation is then reported to `query-arm` as `--write-landed yes|no`.

The state owner already ships the state half of the durability closure landed by issue #562, and the skill invokes none of it:

[...]

Issue #709, opened 2026-07-22
Make the create-issue Step 3.6 audit dispatch instructions observable and verifiably canonical (narrowing parent steering)


The01Geek:
## Problem Statement

The `/devflow:create-issue` Step 3.6 fresh-context audit exists so an auditor that **did not draft the issue** can judge it without the drafter's anchoring (`skills/create-issue/references/step-3-6-audit.md`: "So after Step 3.5 passes and before Step 4 presents anything, dispatch one fresh-context audit subagent whose entire value is that it did not draft the issue"). Issue #600 hardened this by moving the ~2,000-word audit-instruction block out of the freehand orchestrator surface into the deterministic, observable renderer `scripts/render-audit-prompt.py`.

But the renderer output is only *part* of what the auditor reads. The orchestrator still composes a **compact preamble** freehand that wraps the renderer invocation. Per `step-3-6-audit.md` (Consumption categories (i)), that preamble carries "the draft title (in prose), the absolute draft path (file arm) or the sentinel-bracketed draft body (embed arm), the exact renderer invocation to run first, the absolute template-file path, the instruction to treat the renderer's stdout as the complete audit instructions only when its two markers stand in position, and the fallback ladder." The renderer deliberately does **not** own this surface — the draft title "never crosses a command line at all — it travels in the orchestrator's dispatch preamble prose" (`scripts/render-audit-prompt.py` docstring; `skills/create-issue/references/audit-prompt-template.md`: "The draft title never appears here").

Nothing observes or constrains the preamble beyond intent prose. It can add focus, prioritization, reassurance, or scoping language — subtle *positive* steering, not only leaked prior findings — and a topic-primed audit that returns `VERDICT: FILE` is then **indistinguishable in recorded lifecycle state and approval eligibility** from an unsteered fresh-context audit. The state owner records what the auditor **read** (a full-content draft digest, cross-checked against the auditor's quoted `git hash-object` return) but records **no evidence of what the auditor was told**, and the approve-eligibility gate never reads any steering-related operand. The audit's entire independence guarantee therefore rests on unenforced discipline, exactly the gap that issue #600 closed for the instruction block and left open for the preamble.

[...]

PR #884, opened 2026-07-28
create-issue Step 3.6: tool-owned round kinds and a durable byte history

### LINKED ISSUE #793: create-issue Step 3.6: tool-owned round kinds and a durable byte history so later rounds stop re-deriving the whole draft
## Dependencies

- Unblocked: #792 landed (PR #800, merged 2026-07-25). The final-byte coverage axis, its dedicated round slot and its offer producer are in the tree, so this issue's round kinds decide how a claim-scoped round interacts with an axis that now exists.
- Unblocked, and **this issue lands second**: #795 landed (PR #859, merged 2026-07-27). It reshaped the same `record-dispatch` caller contract and argparse surface this issue adds a required `--kind` to, so the reconciliation obligation the earlier draft left to whichever landed second is now wholly this issue's. Its shipped surface — re-derive it from the tree rather than from this list — includes: the `next_call=` answer-line contract, under which every subcommand outside a named exclusion set prints as its **final** stdout line one of exactly three shapes (an invocation line, `next_call=none`, `next_call=unestablished reason=<token>`), with every state-derivable operand filled, every caller-supplied operand rendered as a bare flag name, and a `needs=` field naming those flags; the dispatch-routing renderers that render a `record-dispatch … --round` invocation bare in `needs=`; and the closed-by-construction enumerations (the query-class listing, the `next_call=` exclusion set, `_NEXT_ACTIONS`) that this issue's new round-kind vocabulary, its kind query and its confirming-round next-action token must be admitted by. A required `--kind` that does not reach the rendered `record-dispatch` invocation and its `needs=` list reintroduces exactly the forgotten-flag failure #795 removed.

## Problem Statement

DevFlow maintainers drafting complex engine issues with `/devflow:create-issue` spend four or five Step 3.6 audit rounds and surface 15–22 substantive findings per run. Every one of those rounds is the *same* round: a cold, whole-draft, whole-repository derivation. The state owner records an `arm` per round and nothing else about what that round is *for*, and `scripts/render-audit-prompt.py` forwards no prior-round state into any arm's instructions — so round 4 re-establishes from scratch the repository facts round 1 established, over sections no revision has touched, while the findings it should be re-checking sit durably in the ledger untouched by any targeted treatment.

Deleting rounds is not the fix: later rounds keep finding real defects. The maintainer-reported cost is tens of minutes and roughly 100k tokens per round, with a supported saving of roughly 22–38 minutes across the quantified cases.

## Current Behavior

**Every round is undifferentiated, and the tool has no vocabulary for anything else.**

- The round record `cmd_record_dispatch` appends is `{'round', 'attempts', 'no_parseable_retry_used', 'unreadable_retry_used', 'outcome', 'findings_count', 'consumer_dimensions_appended', 'embed_markers', 'degraded', 'adjudicated_verdict', 'must_revise_count', 'advisory_count', 'invalid_count', 'unresolved_must_revise', 'final_byte_pass', 'final_byte_pass_digest'}`. There is no kind, scope, or delta field, and no producer for one anywhere in the file.
- `_EVENTS`, `TRANSITIONS`, and `_NEXT_ACTIONS` carry no kind dimension. `arm` (`file` / `embed` / `inline`) is the only per-round dispatch dimension the tool owns.
- Blinding is already free and already total: the renderer forwards no prior-round findings, verdicts, or revision history on any arm. A differently-scoped later round is therefore a **round-scheduling** change, not a transport change.

**The instruction file is regenerated and digest-compared over a closed recorded tuple, and a divergence is sticky.** `cmd_record_dispatch` records `attempt['instructions'] = {'digest': …, 'instructions_path': …, 'draft_path': …, 'template_path': …}` and the regeneration re-invokes `render_instructions(template_path, slug, draft_path, instructions_path, draft_text)`, documented pure. A divergence sets `rnd['any_dispatch_diverged'] = True`, which `_steering_established` treats as terminal for the round.

**The staging artifact retains only the latest staged bytes.** `stage` lands them at the single nonce-keyed path `issue-draft-<slug>.<nonce>.staged.md`, so a second revision overwrites the first's bytes and, after `apply`, the canonical file and the staging artifact hold the same content. A tool-derived cumulative delta has no operand today.

**The automatic re-audit slot is spent only after a `REVISE` predecessor.** `cmd_record_dispatch` spends it under `if (not final_byte_pass and prev is not None and prev.get('outcome') == 'REVISE' and doc.get('automatic_reaudits_used', 0) < _MAX_AUTOMATIC_REAUDITS)`, and `_MAX_AUTOMATIC_REAUDITS = 1`.

**Thirteen derivations key on the run's latest completed round — eleven directly and two through the `_last_discovery_round` selector #792 introduced —**, spanning the read and mutation surfaces, and none of them is kind-aware because no kind exists.

**The measurement instrument has no round awareness.** `scripts/create-issue-context-eval.py` parses `message.usage.*` and `Read` tool-use blocks from a single transcript directory; it has no notion of a round, a round boundary, or a round kind.

## Desired Behavior

**The state owner owns the round kind exactly as it already owns the arm.** A closed vocabulary of exactly two kinds — `discovery` and `targeted`, complete by construction — is derived by the tool from recorded facts, answered by a query, and enforced at `record-dispatch`, which refuses any other kind. The orchestrator never chooses a kind, because the context that would choose it is the anchored context the audit exists to remove.

**The run keeps a durable, self-verifying, nameable byte history, because a tool-derived delta has no operand without one.** `stage` lands the bytes at a path carrying both this run's nonce and the staged bytes' own digest, reports that resolved path, and the state owner records it — so each distinct draft state survives as its own artifact, and the recovery arm can still name the artifact it must re-apply after an interruption, which a path known only to the turn that computed it could not. The tool reconstructs the "before" side of a span by locating the artifact whose digest equals the digest recorded at the relevant round's dispatch; a missing artifact is a missing operand rather than a silently wrong one.

**Kind selection fails toward the expensive kind, never away from it.** `targeted` is selected only when a recorded revision postdates the last completed round, the epoch is on the `file` arm, the round's dispatch bytes are recoverable from the byte history with a matching digest, and the enumerated claim set is non-empty. Every other input selects `discovery` and names the failing condition.

**A `targeted` auditor learns what to check, never what was concluded.** It receives the draft file as sole source, the enumerated already-raised findings as id plus one-line summary, and the tool-derived changed-section set. Both payloads travel in one **dispatch-scope file** whose path and content digest are recorded on the dispatch attempt and frozen there — nothing rides beside it — so the instruction file the auditor hashes still regenerates byte-identically from the round's recorded inputs, and a scoped round establishes steering on the same terms a cold one does rather than being pushed onto the override path by its own optimisation. It receives no status, no severity, no disposition, no prior verdict, no rationale and no evidence — the `fix_decision`-carrying shape this repository already withholds from every independence-bearing pass. It verdicts each claim `addressed` or `not-addressed` with a quote from the draft, inspects the changed sections for defects the revision introduced, and reports anything noticed outside them as out-of-scope observations that do not extend the round. "Attempted" is not addressed.

**A `targeted` round cannot pass vacuously.** The renderer refuses an empty claim set, any claim not positively returned `addressed` counts as not-addressed, and a degraded or unusable return falls through to `discovery` rather than to a clean verdict.

**A `targeted` round is never mistaken for whole-draft evidence, and a clean one is confirmed rather than trusted.** It never grounds the clean scan; when it returns every claim `addressed` the tool selects a confirming whole-draft round, answered through a new closed-set next-action token and funded from a **counter of its own** — separate from the automatic re-audit pool, because that pool is shared and a run with two revision cycles exhausts it before the confirming round is ever reached. The shared `_MAX_AUTOMATIC_REAUDITS` constant is left alone: it is also read by the branch that decides whether the skill runs another round itself or asks the user first, so raising it would move a round out of the user's decision on every run, including the embed-arm, inline-arm and empty-delta runs that can never take a scoped round and would pay a full cold round for no saving.

**The efficiency claim is measured on the channel that already carries the cost it reduces.** The ~100k tokens a round costs are spent by the **auditor subagent**, whose records the existing instrument *drops* — one line excludes `isSidechain` — while the harness emits them, with full `usage` payloads, into the very transcripts the instrument reads. So the instrument stops discarding them: it attributes sidechain usage per round — boundaries derived from the transcript's own `record-dispatch` records, since the state document carries no clock, and the state file supplying only each round's kind — and main-thread context stays a secondary figure. Nothing is sourced from a cost the orchestrator volunteers, alongside three named escaped-defect proxies with the unmechanizable one declared out of scope rather than implied away.

## User Impact

A maintainer drafting a complex engine issue stops paying for round 4 to re-read the sections round 3 already cleared: later rounds check the claims they exist to check, in a fraction of the context, and the confirming round lands without a user prompt. Convergence improves rather than degrades, because a `targeted` round reports per-claim `addressed`/`not-addressed` instead of surfacing five fresh orthogonal findings on a draft that keeps growing.

## Technical Context

- **Relevant Classes/Files** — `scripts/issue-audit-state.py` (`_MAX_AUTOMATIC_REAUDITS`, `_EVENTS`, `TRANSITIONS`, `_NEXT_ACTIONS`, `_checked_action`, `cmd_record_dispatch`, `cmd_record_revision`, every `last_completed` reader, `_steering_established`, `_validate`, the argparse wiring); `scripts/render-audit-prompt.py`; `scripts/stage-draft-write.py`; `scripts/create-issue-context-eval.py`; `skills/create-issue/references/{step-3-6-audit,audit-prompt-template}.md`; `lib/test/modules/create-issue-contract.sh`; `lib/test/test_python_scripts.py`; `lib/test/test_render_audit_prompt.py`; `lib/test/test_create_issue_context_eval.py`; `lib/test/run.sh`; `docs/DEVFLOW_SYSTEM_OVERVIEW.md`; `docs/create-issue-context.md`.

- **Verified — the constraint the dispatch-scope file must satisfy:** the file-arm instruction file is regenerated and digest-compared over a closed recorded input tuple, and a divergence is sticky. `cmd_record_dispatch` records `attempt['instructions'] = {'digest': …, 'instructions_path': …, 'draft_path': …, 'template_path': …}` — no payload operand — and the regeneration re-invokes `render_instructions(template_path, slug, draft_path, instructions_path, draft_text)`, documented pure over `(template bytes, draft title, paths, installation)`. On a divergence it sets `rnd['any_dispatch_diverged'] = True`, and `_steering_established` opens `if rnd.get('any_dispatch_diverged'): return False` under the comment `letting a retry erase it would restore the laundering path the dispatch-time refusal was removed for`; `evaluate_eligibility` then withholds the clean ground with `steering-unestablished`. Any payload carried as an unrecorded render argument — or read from live run state, where post-close status mutations would give the return-time regeneration a different ledger than dispatch saw — would make **every** scoped round diverge, so both payloads are widened **into** the recorded tuple.

- **Verified:** the automatic budget is one shared pool, and a two-revision-cycle run exhausts it before the confirming round. `cmd_record_dispatch` spends under `if (not final_byte_pass and prev is not None and prev.get('outcome') == 'REVISE' and doc.get('automatic_reaudits_used', 0) < _MAX_AUTOMATIC_REAUDITS)` and then refuses any round for which `len(doc['rounds']) >= _funded_rounds(doc)`, where `_funded_rounds` is `1 + sum(doc.get(k, 0) for k in _ROUND_BUDGETS)` and `_ROUND_BUDGETS = ('automatic_reaudits_used', 'user_rounds_used', 'final_byte_passes_used')`; `record-return`'s `accept-file` branch sets `rnd['outcome'] = 'FILE'`, so a clean `targeted` round spends nothing. Walk it with the shipped ceiling of one: round 2 is funded off round 1's `REVISE` and drives the counter to its ceiling, so **round 3 is already refused** unless an accepted user offer funds it. A confirming round after a clean `targeted` round therefore has no automatic funding available at any position — not merely at the second revision cycle. This is why the confirming round gets its own counter rather than a widened predicate on the shared one.

- **Verified:** the shared constant has a second reader. `next_action` compares `state.get('automatic_reaudits_used', 0) < _MAX_AUTOMATIC_REAUDITS` to choose between `revise-and-reaudit` (run another round now) and `revise-then-evaluate-offer` (ask the user first), and `docs/DEVFLOW_SYSTEM_OVERVIEW.md` states the principle that paragraph encodes: `the user, not the skill, spends the tokens`. Raising the constant therefore reassigns a token-spending decision on every run, which is why it stays where it is.

- **Verified:** the confirming round needs a next-action token the closed vocabulary cannot express. `_NEXT_ACTIONS` holds eight tokens, `_checked_action` raises on anything outside it, and `next_action` returns `'proceed'` the moment a round's outcome is `FILE`. That vocabulary is a three-way coupled site: the tuple in the tool, the verbatim eight-token obey list in `skills/create-issue/references/step-3-6-audit.md`, and the `devflow_module_pin_unique` pin on that sentence in `lib/test/modules/create-issue-contract.sh`, whose own comment calls it a `COUPLED PAIR with the tool`.

- **Verified:** the thirteen readers of the latest completed round, by enclosing function. An operand-agnostic `grep -c "last_completed(" scripts/issue-audit-state.py` returns 12 lines — the definition plus eleven call sites — resolving by enclosing function to `_effective_unresolved`, `_convergence_basis`, `evaluate_triggers`, `evaluate_convergence`, `_valid_override`, `evaluate_eligibility`, `summary_fields`, `cmd_record_adjudication`, `_ledgered_round`, `cmd_record_revision`, `cmd_record_override`. Two further readers — `_coverage_round` and `_calibration_round` — reach the same round through `_last_discovery_round`, the #792 selector that excludes a final-byte pass, so the sweep must cover **both** operands. An operand-pinned sweep on `last_completed(state)` misses the last four, including the guard that binds a filing override's digest to an epoch.

- **Verified:** the clean scan breaks on the first `REVISE`. `evaluate_eligibility` walks completed rounds newest-first with `if rnd.get('outcome') == 'FILE': clean = rnd; break` followed by `if rnd.get('outcome') == 'REVISE': break`, terminating at `return _no('unaudited-revision')`. A rule that merely *skips* a clean `targeted` round lands the scan on the preceding `REVISE`, breaks, and refuses — which is why a confirming whole-draft round is scheduled rather than the scan taught to skip.

- **Verified — and this is why the change includes the staging helper:** today's staging artifact retains only the *latest* staged bytes. `skills/create-issue/references/step-3-6-audit.md` specifies the single path `<bound-root>/.devflow/tmp/issue-draft-<slug>.<nonce>.staged.md` and states `The staging path carries this run's **nonce**, so no prior run's artifact is reachable and **no delete step exists at all**`; `scripts/stage-draft-write.py`'s docstring states `apply` `NEVER renames the staging artifact itself, so S survives a successful replace and a failed one alike`. Both guarantees are about *that one path*, which every `stage` replaces. An executed `grep -n -E "unlink|os\.remove|rmtree|missing_ok|delete" scripts/stage-draft-write.py` returns exactly one hit, `os.unlink(tmp)` inside `_atomic_write`'s temp-sibling cleanup — no artifact is deleted, and equally none is preserved per revision.

- **Verified:** the recovery arm names the artifact from a path that is a pure function of durable values today. `step-3-6-audit.md` directs a write-failure recovery to `re-run `apply` **exactly once** from the still-present staging artifact`, and states that after a non-revision canonical write `the comparand is that write's own **in-turn** `--expect-digest`` — so the two non-revision write sites persist no digest at all. A purely digest-derived path would be known only to the turn that computed it, which is why the nonce stays in the leaf and each resolved path is recorded.

- **Verified:** the digest-keyed lookup rests on an already-guaranteed cross-tool agreement. `scripts/stage-draft-write.py`'s docstring states `Digests come from ``git hash-object --stdin --no-filters`` — the one filter-free mode ``issue-audit-state.py`` uses at every compare site, so all digests agree byte-for-byte on every host regardless of ``core.autocrlf``.`

- **Verified:** retaining more *staged* artifacts leaves the enumerations byte-unchanged, but the *dispatch-scope* artifact is a new name they do not cover. The file-arm enumeration names `any staged canonical-draft artifact `.devflow/tmp/issue-draft-<slug>.*.staged.md`` — a glob a digest-bearing staged name matches exactly as the nonce-only name does. It is a **closed list of exactly six named paths** (seven on the embed arm), not a `.devflow/tmp/**` blanket, so a new persistent artifact holding this run's enumerated findings falls outside it and a later round's auditor — which has repository read access — could read it unimpeded. The artifact must persist, because its digest is recompared at `record-return`. The artifact directory is gitignored: `git check-ignore -v .devflow/tmp/probe.md` reports `.devflow/.gitignore:4:/tmp/`.

- **Verified:** a `targeted` dispatch needs no new consumption category. `step-3-6-audit.md`'s *Consumption categories (complete by construction)* block opens `(i) **Every state-owner-routed file-arm audit dispatch** — the initial round, same-round retries, boundary-offer rounds, revise-and-reaudit rounds, and Step 4 sub-step 4 re-audits — takes the **generated-instructions transport**`, so a new round kind on that transport is a member of category (i)'s inner round list rather than a sixth category, and the count-locked `(vi)` absence pin stays green.

- **Verified:** the measurement instrument has no round awareness. An executed `grep -c -E "record-dispatch|issue-audit-state|\bround\b" scripts/create-issue-context-eval.py` returns **0**. `python3 scripts/create-issue-context-eval.py --help` reports a positional `transcript_dir`, `--format {text,json}` and `--large-block-chars`, and `docs/create-issue-context.md` states it measures runtime main-thread context per turn as `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` and is `a maintainer/CI-adjacent instrument, never invoked by the skill's runtime path`. An executed `git ls-files lib/test/fixtures/create-issue-eval/` returns 7 tracked fixtures including a `before/` and `after/` session pair, so the A/B shape already exists.

- **Verified:** the per-finding evidence `locator` is a repository coordinate, not a draft one. `_EVIDENCE_REQUIRED = ('locator', 'command', 'observed', 'baseline_revision')`, the conflict index keys on `(entry['locator'], entry.get('command'))` as a probe identity, and the audit prompt defines the field as the `path:line` or `path:region` the check reads — which is why the scope-escape proxy is keyed on the draft line the per-finding bar already requires each finding to quote.

- **Verified:** a mechanical escaped-information number is not deliverable for this skill. Open issue #775, deferred from #767, records that the create-issue corpus shows zero compactions and that detecting semantic re-derivation needs semantic diffing rather than byte comparison. It is a related follow-up, not a prerequisite.

- **Verified:** the coupled sites the automatic-budget change touches. An executed line-based sweep over the git-index population for `at most once` and `automatic re-audit` returns, outside `CHANGELOG.md` and `.devflow/learnings/`: `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s prose mirror `re-audits **at most once** automatically`; `lib/test/run.sh`'s behavioral block commented `The AUTOMATIC BUDGET: one initial round plus AT MOST ONE automatic re-audit.` plus `round 2 is the automatic re-audit (funded); round 3 unfunded refuses`; and `lib/test/modules/create-issue-contract.sh`'s comment quoting the retired pin phrase `automatic budget stays one audit plus **at most one** automatic re-audit`. `run.sh`'s cap test imports `m._USER_ROUND_CAP` rather than transcribing it, under the comment `a literal 3 here would keep passing while the tool's cap drifted underneath it`.

- **Verified:** the pin blast radius, enumerated with a whitespace-normalized sweep whose result set was byte-identical to the line-based leg for all fourteen swept literals. `lib/test/test_python_scripts.py` asserts whole-line `query-triggers` comparands and holds a hand-transcribed `_TRANSITION_ROWS` locked against the module table by length and in-order content. `lib/test/modules/create-issue-contract.sh` holds count-locked absence pins including `devflow_module_pin_count '(vi)'` expected `0` and `'within the existing automatic audit budget'` expected `0`. The three out-of-bounds enumerations are count-locked in their own literals (file arm 6 paths, embed arm 7 files) with a fourth comparand in `lib/test/test_render_audit_prompt.py`'s `EMBED_ARM_OOB`.

- **Verified:** the execution tier and the discharge command. An executed `grep -rn create-issue .github/workflows/` matches only `ci.yml`'s test-module list, so this skill runs on the local and interactive tier only and no new command head is invoked. An executed read of `.devflow/config.json` returns a `devflow_implement.allowed_tools` list that grants the suite and lint heads in their direct leading-token form (`Bash(lib/test/run.sh:*)`, `Bash(lib/test/run-module.sh:*)`, `Bash(lib/preflight.sh:*)`, `Bash(shellcheck:*)`) alongside a large set of `scripts/*` helper grants — re-derive the live list rather than trusting a transcription. `python3 scripts/create-issue-context-eval.py` is **not** among them, so every measurement criterion whose discharge is *executable in the implementing PR* runs through `lib/test/run.sh` in the **direct leading-token form** the grant matches, never the `bash lib/test/run.sh` wrapper this repository keeps deny-floored. The real-corpus figures are deliberately **not** such a criterion: they are a post-merge maintainer obligation, so the PR ships the record's placeholder shape rather than invented numbers. Adding a grant would not help: that key resolves at trigger time from the default branch, so a grant this change ships is inert for its own implementing run.

- **Verified:** the state owner reads no config. Its module docstring states `and reads no config file. Its only subprocess is native git`. This change adds no `.devflow/config.json` key, so the schema is untouched and the trigger-time-resolved in-PR-inert hazard cannot arise.

- **Verified:** this repository already ships every shape this mechanism reuses. `skills/review/phases/phase-0-3-6-blocker-recheck.md` dispatches a blinded verifier receiving **only** the enumerated blockers and the delta — "never the fixer's reasoning, the prior run's other findings, or any fix-decision rationale" — treats anything not positively reported fixed as still-unfixed, refuses an empty enumerated set as vacuously true, falls through to the full pipeline when anything outside the enumerated sites changed, and fails closed on a verifier failure. `skills/review-and-fix/references/fix-delta-gate.md` blinds a verifier over a mechanically computed delta; `loop-exit.md` gates on "nothing changed after the reviewer ran" and refuses an empty or errored delta as clean; `skills/review/phases/phase-3-agents.md`'s `fix_decision`-carrying block is the anchoring shape every independence-bearing pass withholds.

- **Consumer-repo, host-OS and degraded-environment variance.** A consumer receives this change through the `devflow_version` vendor fetch as skill and `scripts/` content only — no new config key, no new tool grant, no workflow edit, so `install.sh`'s copy loop is uninvolved and the two-artifact install skew does not arise; the eval instrument stays a maintainer instrument of this repository and is never on a consumer's runtime path. The added path component is a git object id, so it is bounded lowercase hex and portable on Windows, macOS/BSD and Linux. The only subprocess stays native `git`, and `git hash-object` reads no repository history, so a shallow clone is immaterial. No value that selects a round kind or a trigger passes through `tr`, `sed`, `wc`, `cut` or `head`.

- **Architecture Alignment** — the round kind becomes the second tool-owned per-round dispatch dimension beside the arm, resolved by the same query-then-obey-then-cross-check shape `query-arm` → `record-dispatch` → `--write-path` already uses.

- **Dependencies** — python3 standard library only. No new binary, no new config key, no new cloud tool grant.

- **Data/Schema Considerations** — the round record gains `kind` and, on a `targeted` round, the derived scope it was dispatched under; the run record gains the recorded staging paths. Every added field is additive and read with a default, and `SCHEMA_VERSION` is held: `_validate` treats a version mismatch as a hard refusal with no migration path, so a bump would strand every run in flight at upgrade time. `_validate` does not reject unknown extra keys, so holding the version also lets a state file written by the new build load unchanged under the old one.

## Acceptance Criteria

- [ ] `scripts/issue-audit-state.py` defines a closed round-kind vocabulary whose members are exactly these two — `discovery`, `targeted`, complete by construction — and an off-vocabulary kind reaching any consuming function raises rather than silently taking a permissive path, matching the existing eligibility mode-vocabulary guard.
- [ ] A read-only query answers the kind the next round must take, together with the reason token that selected it, the delta state and the enumerated claim ids, and it always exits 0 once its arguments parse.
- [ ] `record-dispatch` requires the kind and refuses any kind other than the one the query selects at that moment, with a named breadcrumb, mirroring the existing `write-path-mismatch` cross-check.
- [ ] The scope file records the **basis digest** of the canonical bytes the changed-section set was computed from, and `record-dispatch` refuses a `targeted` dispatch whose `--draft-file` digest differs from it, with a named breadcrumb mirroring the existing `write-path-mismatch` cross-check — because the skill re-runs the Step 3 gate between selection and dispatch, so a byte edit landing in that window would point the auditor at superseded regions while carriage, regeneration and steering all still pass.
- [ ] A scope file that is **absent or unreadable** at `record-return` records its own named reason, distinct from the tampered-file arm, rather than falling through an unnamed regeneration failure.
- [ ] The tool selects `targeted` only when all of these hold — exactly these five conditions, complete by construction: a recorded revision postdates the last completed round; the latest attempt of that round was on the `file` arm; the round's dispatch bytes are recoverable from the byte history and their recomputed digest equals the recorded dispatch digest; the enumerated claim set is non-empty; and the scope file's recorded basis digest equals the canonical file's digest at dispatch.
- [ ] Every input that fails any `targeted` condition selects `discovery` and names the failing condition in the reason token.
- [ ] An empty computed changed-section set selects `discovery`, and a changed-section computation that errors selects `discovery` — neither is read as "nothing changed".
- [ ] `scripts/stage-draft-write.py`'s `stage` mode takes `--path` as a **base** and completes it with the staged bytes' digest, reporting the resolved path — a caller cannot compose that leaf, because the digest is computed from stdin inside `stage` and each shell fence is a fresh process. On a run with a bound nonce the resolved path carries both the nonce and the digest; a second `stage` of different bytes leaves the first artifact readable at its own path.
- [ ] The `state-owner unavailable` arm keeps its existing nonce-free name and records nothing durably, and that reduced durability is disclosed on that arm's existing marker rather than asserted away — the staging criteria above are scoped to the arms where a state owner is available.
- [ ] Re-staging byte-identical content resolves to the same path and leaves exactly one artifact for that byte state.
- [ ] Every `stage` **on a state-owner-available arm** records its resolved staging path durably, so an interrupted or compacted turn recovers the artifact's name from recorded state rather than from that turn's stdout — asserted at all three write sites, including the revision site, whose `stage` → separate fence → `emit` → `record-revision` → `apply` window is the longest and whose revision record is `targeted` selection condition one.
- [ ] A `stage` → new fence → `emit` → `apply` round-trip resolves the same artifact from recorded state with the staging turn's stdout discarded.
- [ ] The write-failure recovery arm names the staging artifact to re-apply from the durably recorded path, and a run holding several staged artifacts re-applies the one that write recorded rather than the newest on disk.
- [ ] The tool reconstructs a round's dispatch bytes by locating the staged artifact whose digest equals that round's recorded dispatch digest, and accepts the artifact only when its recomputed digest equals that recorded digest.
- [ ] The whole `targeted` payload — the enumerated claims **and** the tool-derived changed-section set — reaches the auditor as one **dispatch-scope file whose path and content digest are both recorded on the dispatch attempt**, frozen there, so the instruction-file regeneration reproduces byte-identically from the recorded tuple alone and no payload rides outside it.
- [ ] A `targeted` round carrying a non-empty changed-section set records a **verified** dispatch regeneration and establishes steering, asserted directly rather than inferred.
- [ ] A post-dispatch ledger mutation — a resolution, reopen or invalidation recorded after the round opened — does not change the round's recorded dispatch-scope digest and does not turn its regeneration `diverged`.
- [ ] A tampered dispatch-scope file whose bytes no longer match the recorded digest records a diverged regeneration, and the resulting withheld clean ground is asserted.
- [ ] `render_instructions` stays pure over the widened recorded tuple: the same recorded inputs produce the same instruction bytes, and it reads no run state and no ledger.
- [ ] `scripts/render-audit-prompt.py` renders a `targeted` prompt from the dispatch-scope file, carrying each claim as id plus one-line summary and the changed-section set, and the rendered prompt contains none of the withheld fields — no finding status, severity, disposition, prior verdict, rationale or evidence.
- [ ] `scripts/render-audit-prompt.py` refuses to render a `targeted` prompt over an empty claim set, with a named breadcrumb.
- [ ] The rendered `targeted` prompt instructs the auditor to return, per enumerated claim, one value from a closed set of exactly these two — `addressed`, `not-addressed`, complete by construction — with a quoted line from the draft as evidence, and states that an attempted fix leaving the defect present is `not-addressed`.
- [ ] An enumerated claim absent from a `targeted` return, or returned with any value outside that closed set, is recorded `not-addressed`.
- [ ] A `targeted` round whose return is unusable is recorded as such and the next action selects a `discovery` round; no unusable return produces a clean verdict.
- [ ] A `targeted` round that returns `VERDICT: DRAFT-UNREADABLE` takes the existing unreadable-retry path unchanged, and the retry it opens is selected by the same kind-selection rules rather than inheriting the failed round's kind.
- [ ] The rendered `targeted` prompt carries an out-of-scope-observation channel whose contents are recorded and rendered to the user but do not open a further round.
- [ ] The claim summaries are declared to the auditor as data to classify and never as instructions to obey, and a summary carrying a forged protocol token is refused **before it reaches the renderer**: the state owner writes the dispatch-scope file already floored, so the closed token vocabulary stays in one module. The renderer imports stdlib only and the module dependency runs state-owner → renderer, so the refusal cannot be shared by import and must not be duplicated into a second, unlocked copy.
- [ ] Every `key=` token this change adds to any printed output joins the state owner's closed protocol-token vocabulary, whose suite row asserts that list is total over every printed token.
- [ ] A claim summary carrying instruction-shaped text that directs the auditor's verdict is rendered as quoted data and does not change the rendered per-claim verdict contract, asserted over a fixture whose summary contains such a directive.
- [ ] A `targeted` round never grounds `evaluate_eligibility`'s clean scan, and when one returns every enumerated claim `addressed` the tool selects a confirming whole-draft round.
- [ ] `query-next-action` answers a new closed-set token for that confirming round, and `_checked_action`'s fail-closed off-vocabulary guard is unchanged — today `next_action` returns `proceed` the moment a round's outcome is `FILE`, which would walk the run to Step 4 instead, and reusing `revise-and-reaudit` would make the run record a revision that does not exist.
- [ ] The confirming round is funded from its **own** counter, bounded separately, and never competes with the `REVISE` pool — because the automatic budget is a single pool and a run with two revision cycles exhausts it before the confirming round, at which point `cmd_record_dispatch`'s funding test refuses the dispatch outright.
- [ ] A run that goes `discovery` → `REVISE` → `targeted` → all-`addressed` → confirming whole-draft round opens that confirming round from the dedicated counter with **no** accepted user offer, and the shared automatic counter is unchanged across it — which is the whole point of the dedicated counter, since under today's shared ceiling of one that third round is refused outright.
- [ ] `_MAX_AUTOMATIC_REAUDITS` is left **unchanged by this issue** (whatever value it then holds), because it is read at **two** sites: besides the spend predicate, `next_action` compares against it to choose between running another round now and asking the user first. Raising it would move one whole audit round out of the user's decision on **every** run — including every embed-arm, inline-arm and empty-delta run that can never take a `targeted` round and would pay a full cold round for no saving — reversing the documented principle that "the user, not the skill, spends the tokens". This was adjudicated on 2026-07-26: #827, which proposed raising the constant from 1 to 2, was **closed as not planned** in favour of the position stated here. The constant stays at its current value and no open issue proposes changing it.
- [ ] Every reader of the run's latest completed round decides its treatment of a `targeted` round explicitly, naming each by function — complete by construction (`_coverage_round` and `_calibration_round` reach the round through `_last_discovery_round`; the rest read `last_completed` directly): `_coverage_round`, `_calibration_round`, `_convergence_basis` and `evaluate_eligibility`'s clean scan treat a `targeted` round as **not** whole-draft evidence, so a successful scoped round neither demotes an established coverage backing, nor wipes a recorded calibration signal, nor reports `basis=adjudicated`, nor grounds the clean scan; `summary_fields` renders the Step 4 audit-summary verdict and class counts from the latest **whole-draft** round and names the scoped round separately; `_effective_unresolved`, `evaluate_triggers` and `cmd_record_adjudication` take a decided treatment too: a `targeted` round's per-claim return **updates the existing ledger entry each claim names** — `not-addressed` leaves it unresolved or reopens it, `addressed` changes nothing, since resolution stays the drafter's own recorded verification — and the round records **no ledger of its own**. Without that seam the shipped reconciliation discipline lists every re-checked defect on both rounds' ledgers and the aggregate counts it per listing, so a mechanism built to reduce rounds would inflate the very count that fires the offer for more of them, compounding with each scoped round. And `evaluate_convergence`, `_ledgered_round`, `cmd_record_revision`, `cmd_record_override` and `_valid_override` remain kind-blind.
- [ ] A `discovery` → `REVISE` → `targeted` run in which every claim returns `not-addressed` leaves the run-wide effective unresolved count equal to the discovery round's, not doubled.
- [ ] `_valid_override` and `cmd_record_override` resolve their epoch from the **same** round and stay a matched pair, because the read-side gate's own docstring states its preconditions exist "mirroring the guards `record-override` applies at the write boundary" — an override overrides *auditing*, not *whole-draft evidence*, so both stay kind-blind rather than one half being taught to skip a `targeted` epoch.
- [ ] A run whose latest completed round is `targeted` records a `user-decline` override and then answers `eligible` on the override ground, so the decline path files rather than dead-ending at `stale-override` with `emit-body` gated shut.
- [ ] The final-byte coverage axis this issue's prerequisite introduces reports `covered` only for a whole-draft round, so a clean `targeted` round never sets it. The axis exists in the tree (`final_byte_pass` on the round record, `_last_discovery_round`, `_FINAL_BYTE_PASS_CAP`), so this criterion is live and no deferral arm applies.
- [ ] The generated instruction file `.devflow/tmp/issue-audit-dispatch-<slug>.md` is named in the **embed-arm** out-of-bounds enumeration only. It persists after the run and the scope file's content is spliced into it, so a `targeted` round that returns `VERDICT: DRAFT-UNREADABLE` and retries on the embed arm would otherwise hand its auditor a list naming neither that path nor the scope file while both still hold the round's enumerated findings. The **file-arm** and dispatch-instructions enumerations deliberately do **not** name it: on that arm it is the artifact the auditor is directed to read and hash, so naming it would order the auditor to void its own return, and the path is slug-keyed and regenerated on every file-arm dispatch, so it can never hold a prior round's payload there.
- [ ] The dispatch-scope artifact is named in **both** out-of-bounds enumerations as a glob, so a later round's fresh-context auditor cannot read the prior round's enumerated findings off disk — the artifact must persist, because its digest is recompared at `record-return` for the regeneration check, so it cannot be deleted after dispatch, and the six-path and seven-file enumerations are closed lists rather than a `.devflow/tmp/**` blanket.
- [ ] Adding that glob is a same-change edit across every carrier, named per site rather than by count: `skills/create-issue/references/step-3-6-audit.md`'s skill-prose path list and its count word; `skills/create-issue/references/fallback-audit-dispatch-arms.md`'s embed-arm list and its count word; **all three** enumerations inside `skills/create-issue/references/audit-prompt-template.md` — the dispatch-instructions block, the file-arm block and the embed-arm block; the `lib/test/modules/create-issue-contract.sh` pins on the two single-line literals; and `lib/test/test_render_audit_prompt.py`'s `EMBED_ARM_OOB` comparand.
- [ ] The dispatch-instructions block — the declaration a file-arm auditor actually obeys, which is line-wrapped and carries **no** comparand today — gains one, asserted against the **rendered** `dispatch-instructions` output rather than a source grep, because a phrase wrapped across lines lives on no single line.
- [ ] The dispatch-scope glob is **total** — a round's own scope file is out of bounds to that round's auditor too, because the payload reaches it only through the rendered splice and never by reading the file — asserted over a run whose confirming whole-draft round follows a `targeted` one. The enumeration therefore stays byte-stable across rounds, which its count-locked comparands require.
- [ ] The state file remains out of bounds on both arms, and the retained staged artifacts stay covered by the existing staged-artifact glob.
- [ ] The Step 3.6 consumption-category enumeration keeps exactly its current five members and its count-locked absence pin stays green, with the `targeted` dispatch carried as a member of category (i)'s inner round list.
- [ ] `scripts/create-issue-context-eval.py` attributes the **sidechain** `usage` records already present in the transcripts it reads — the auditor's own cost — to rounds. The instrument today *chooses* to drop those records with a single line; it is not blind to them, and the committed realistic fixture carries a sidechain record with a full `usage` payload.
- [ ] Round **boundaries** are derived from the transcript's own recorded state-owner invocations — the `record-dispatch --round N` tool-use records — and the state file supplies only the round-to-kind labelling and the round count. The state document carries **no** time or ordering coordinate a transcript record could be keyed to: the module imports no time facility at all and a round record's key set is closed and time-free, so an implementation that tries to join on the state file alone has no join to make.
- [ ] The state owner records only what the transcript cannot supply, and no cost figure is sourced from a value the orchestrator volunteers — the agent-volunteered-operand shape `docs/efficiency-trace.md` records as having steered three issues' work while the harness was emitting the same data deterministically.
- [ ] Wall-clock is **not** claimed as a measured axis on this tier: it is reported `unestablished`, citing that document's local row, rather than asserted as something the orchestrator observes.
- [ ] The instrument's main-thread context figures are reported as a **secondary** axis and are never the sole basis of the reduction claim.
- [ ] `scripts/create-issue-context-eval.py` reports, per run, a per-round breakdown carrying each round's recorded kind and its attributed sidechain cost, and its aggregate summary gains the per-kind medians — asserted from committed fixtures by `lib/test/test_create_issue_context_eval.py`, which `lib/test/run.sh` drives.
- [ ] `scripts/create-issue-context-eval.py` accepts a before and an after run as separate operands and reports the paired deltas for attributed auditor cost, per-run context, round count and finding count — asserted from committed fixtures by the same test. Latency is **not** in that list, because the sibling criterion reports wall-clock `unestablished` on this tier and a paired delta would present a number the tier never measured.
- [ ] The instrument's state-file reader is best-effort: a state file that is absent, unreadable, empty, malformed, carries a wrong-typed rounds container, or carries a round with no recorded kind yields a report whose per-kind figures read `unestablished`, never a number and never a crash.
- [ ] The escaped-defect axis reports exactly these three named proxies, complete by construction: the count of `record-reopen` events per run; the count of must-revise findings raised by a later round whose **quoted draft line** falls inside an earlier `targeted` round's recorded scope; and a declared post-filing class the instrument reports as `unestablished` and never as a number.
- [ ] The ledger records, per finding, the draft line the auditor quoted as the line it attacks, so the scope-escape proxy compares two coordinates in the draft's own space rather than a repository `path:line` against a set of draft sections.
- [ ] The scope-escape proxy reports its own denominator beside its count — the number of later-round must-revise findings carrying no recorded draft line and therefore unattributable — and an unattributable finding is never counted as attributable and never counted as zero.
- [ ] `lib/test/fixtures/create-issue-eval/` carries a committed synthetic before/after pair exercising the per-kind breakdown and the three proxies, and `lib/test/test_create_issue_context_eval.py` asserts the reduction live from those fixtures with a strict inequality rather than from a transcribed figure.
- [ ] `docs/create-issue-context.md` gains a measurement-obligation record for this change under its existing *maintainer measurement obligation* tier, carrying the same **unfilled placeholder shape** that tier already uses, stating that no CI check re-derives the figures and that a measurement producing no decrease means the change is reverted or deferred rather than shipped as a reduction. The figures are filled **post-merge by the maintainer**, not in the implementing PR — capturing an after-corpus requires running the instrument over real transcripts produced by the merged change, which no granted command can do during implementation.
- [ ] `CLAUDE.md`'s create-issue bullet states that `scripts/issue-audit-state.py` owns the audit lifecycle and points to the canonical `docs/DEVFLOW_SYSTEM_OVERVIEW.md` section, and its `#614` sentence names the two files the on-disk reconciliation exempts.
- [ ] `docs/DEVFLOW_SYSTEM_OVERVIEW.md` states one current value for the state `schema_version`, and the superseded figure is marked as the historical value it carried at that issue's time rather than left reading as current.
- [ ] `SCHEMA_VERSION` is held at its current value, every added field is additive and read with a default, no added key joins the required-top-level key set, and a state file written by the post-change build loads under the prior build and behaves as it does today.
- [ ] The required `--kind` operand joins the `next_call=` answer-line contract #795 shipped rather than sitting beside it: every rendered `record-dispatch` invocation carries `--kind` filled from the tool's own selection when it is state-derivable, and the round-kind query and the confirming-round next-action token are admitted by the closed query-class listing, the `next_call=` exclusion predicate and `_NEXT_ACTIONS` — asserted against the **rendered** answer line, so a new required flag the renderer omits fails rather than reintroducing the forgotten-flag failure that contract removed.
- [ ] Every comparand and prose mirror this change perturbs is re-derived in the same change, in the groups below. **`next_call=` group:** the dispatch-routing invocation renderers and their `needs=` composition in `scripts/issue-audit-state.py`, the query-class listing and the exclusion set, and the `next_call=` contract paragraph in `skills/create-issue/references/step-3-6-audit.md` — re-derived from the tree as #795 left them, never from a transcription in this issue. **Protocol-vocabulary group:** `scripts/issue-audit-state.py`'s closed protocol-token list and `lib/test/test_python_scripts.py`'s printed-token coverage row. **Renderer-contract group:** `scripts/render-audit-prompt.py`'s module contract header, which today states it "reads no run *state*" over a closed list of reads and must be amended to admit the round's frozen dispatch-scope file named on its command line; `lib/test/modules/harness-python-guards.sh`'s statelessness guards; and `lib/test/test_render_audit_prompt.py`'s statelessness row. **Budget group:** `lib/test/run.sh`'s automatic-budget behavioral rows, which drive the funding test the new counter joins. The `at most once` prose mirrors and the shared budget comment are deliberately **not** in this group, because `_MAX_AUTOMATIC_REAUDITS` is unchanged. **Staging-path group:** `scripts/stage-draft-write.py`'s `--path` help text **and its module docstring's `stage`, `emit` and `apply` bullets**, which today state the contract "land them at the staging path P"; `skills/create-issue/references/step-3-6-audit.md`'s staging-path sentence, its `stage` fence, its step-2 `emit` fence, its step-3 `apply` fence and its `the staging artifact is nonce-keyed` clause; `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s `nonce-keyed staging artifact` sentence; `lib/test/test_python_scripts.py`'s `#705` rows that construct a nonce-only staged path and read it back through `emit`/`apply`; and the nonce-free `state-owner unavailable` fallback name, which keeps its existing shape. **Lifecycle-vocabulary group:** `lib/test/test_python_scripts.py`'s hand-transcribed `_TRANSITION_ROWS`, locked against the module table by length and in-order content; `_NEXT_ACTIONS` in the tool; the verbatim eight-token obey list in `skills/create-issue/references/step-3-6-audit.md`; and the `devflow_module_pin_unique` pin on that sentence in `lib/test/modules/create-issue-contract.sh`. **Out-of-bounds group:** the file-arm `exactly these 6 paths` and embed-arm `exactly these 7 files` literals with their count words, their `create-issue-contract.sh` pins, the `audit-prompt-template.md` enumeration, and `lib/test/test_render_audit_prompt.py`'s `EMBED_ARM_OOB` comparand.

## Implementation Notes

- **Approach.** Add the round kind as the second tool-owned per-round dispatch dimension beside the arm, resolved by the same query-then-obey-then-cross-check shape the arm already uses. In `scripts/issue-audit-state.py`: add the closed kind vocabulary and a `_checked_kind` guard modelled on `_checked_action`; add the selection function deriving the kind from recorded facts alone and the query that answers it; extend `cmd_record_dispatch` to require `--kind`, cross-check it against the selection, persist it with the derived scope, and **widen the recorded instruction tuple with the dispatch-scope file's path and content digest**; add a **separate** confirming-round counter with its own bounded spend predicate, leaving `_MAX_AUTOMATIC_REAUDITS` and its `REVISE`-keyed predicate untouched — the constant has a second reader in `next_action`, so raising it would reassign a token-spending decision on every run; give every reader of the latest completed round its decided kind treatment. In `scripts/render-audit-prompt.py`: add the `targeted` render block, its dispatch-scope splice with the identity-data floor, and the empty-claim-set refusal. In `audit-prompt-template.md`: add that block's per-claim verdict contract and out-of-scope channel. In `scripts/stage-draft-write.py`: key `stage`'s path on the nonce *and* the staged digest and report the resolved path. In `step-3-6-audit.md`: update the staging fence and state which call to make and when — never the selection logic. Compute the changed-section set inside the tool from the reconstructed dispatch bytes and the current canonical bytes, each digest-verified, falling through to `discovery` on any failure.

- **Code Patterns.** `_checked_action`'s fail-closed off-vocabulary guard; `query-arm` → `record-dispatch` → `--write-path` cross-check for a tool-owned dispatch dimension; `record-adjudication`'s ledger ingestion for the identity-data floor; `phase-0-3-6-blocker-recheck.md`'s enumerated-claim shape, fail-closed default, empty-set refusal and fall-through arm; `loop-exit.md`'s stored-identity gate; `fix-delta-gate.md`'s blinded verifier over a mechanically computed delta.

- **Testing Strategy.** The change is fully covered by test automation: the state owner and the renderer are CLI and function boundaries, and the eval instrument is a pure function of committed fixtures. Test levels are unit assertions in `lib/test/test_python_scripts.py`, renderer assertions in `lib/test/test_render_audit_prompt.py`, CLI round-trip and behavioral rows in `lib/test/run.sh`, contract pins in `lib/test/modules/create-issue-contract.sh`, and fixture-driven assertions in `lib/test/test_create_issue_context_eval.py`. Each assertion is written first and must fail because the behavior does not exist yet, except the budget rows, where the right failure is that today's code opens a third round only after a recorded user offer.

  Named assertions, each tied to an acceptance criterion. **Staging:** two `stage` calls with different bytes leave both artifacts readable at their own paths; identical bytes resolve to one path; each resolved path is recoverable from recorded state after the staging turn's output is discarded; a write-failure recovery on a run holding several artifacts re-applies the recorded one. **Kind selection:** vocabulary closure and an off-vocabulary raise; one row per `targeted` condition driven to failure asserting `discovery` with its reason token; empty-delta and errored-delta rows asserting `discovery`; a `record-dispatch` kind-mismatch refusal. **Regeneration:** an untampered dispatch-scope file carrying a non-empty section set records a verified regeneration and establishes steering; a tampered file records `diverged` and a withheld clean ground; post-dispatch resolution, reopen and invalidation leave the regeneration verified. **Budget:** the confirming round draws on its own counter while `automatic_reaudits_used` is unchanged, and `next_action`'s `revise-and-reaudit` / `revise-then-evaluate-offer` split is byte-unchanged, with ceilings imported from the module rather than transcribed; `discovery`→`REVISE`→`targeted`→all-addressed→confirming round opens with no accepted user offer, paired with a row asserting today's `REVISE`-only predicate refuses it. **Renderer:** the `targeted` prompt carries ids, summaries and the section set, driven from a ledger fixture in which every withheld field is **present** in the input, asserting each is absent from the output by name so the suppression is exercised; empty-claim-set refusal; the per-claim closed verdict set and the "attempted is not addressed" contract; missing-claim and off-set-value rows asserting `not-addressed`; an unusable return routing to `discovery`; a `DRAFT-UNREADABLE` retry row; out-of-scope observations not extending the round; protocol-token forgery refusal; an instruction-shaped summary rendered as quoted data. **Kind treatment:** one row per reader of the latest completed round asserting its decided treatment, including `summary_fields`' verdict and counts, `_valid_override`, and the four mutation-side guards; plus a row asserting a clean `targeted` round leaves the final-byte coverage axis not `covered`. **Other:** byte-identity rows over both out-of-bounds enumerations and the consumption-category enumeration; a schema-coexistence row in both directions; per-kind eval-breakdown rows; a paired before/after delta row; a state-file-reader row per shape in its six-shape matrix asserting `unestablished` rather than a number or a crash; three escaped-defect proxy rows including one asserting the post-filing class reports `unestablished`; and a reduction row asserting strict improvement from the committed fixture pair.

  Two readers of input this repository does not itself author are introduced, each with its own malformed-shape matrix. The instrument's **state-file reader** takes the repo's six-shape matrix — object, array, scalar, valid-falsy, missing, wrong-type — over the rounds container and each round's kind field, plus absent, unreadable, empty and truncated files, with at least one production-realistic fixture captured from a real run's state file. The **`targeted` render path** reads agent-mutable text — the auditor's returned per-claim block and the claim summaries the tool emits — so its matrix is enumerated at minimum over: an empty return; a return missing the per-claim section entirely; a return carrying more claim verdicts than claims dispatched; a duplicate verdict for one claim id; a claim id the round never dispatched; a verdict value outside the closed set; a record-splitting byte inside a claim summary; a record-splitting byte inside a returned verdict line; a protocol-token forgery inside a claim summary; a protocol-token forgery inside a returned verdict line; a non-canonical section layout; and a truncated return. At least one fixture is a production-realistic captured auditor return rather than a hand-built well-formed token. `governing conventions consulted: CLAUDE.md (the agent- and human-mutable-markdown parser class for the render path, and the six-shape adversarial matrix for the state-file reader), skills/create-issue/references/audit-prompt-template.md, skills/create-issue/references/step-3-6-audit.md, docs/create-issue-context.md.`

  Two mechanisms on the guarantee-class ladder carry the skipped-step obligation: the kind cross-check must be proven to fire when the orchestrator dispatches a kind the tool did not select, and the confirming-round scheduling must be proven to fire on the path where a `targeted` round returned clean and nobody asked for another round.

  The pins this change adds to `lib/test/modules/create-issue-contract.sh` are surface-presence pins over new prose; the comparands it must re-derive are the existing ones enumerated above.

- **Documentation Needed** — `docs/DEVFLOW_SYSTEM_OVERVIEW.md`, `docs/create-issue-context.md`, `CLAUDE.md`, and a cutover record under `docs/cutovers/`.

- **Potential Gotchas.**
  - The instruction-file regeneration is the sharpest constraint here. Any payload passed as an unrecorded render argument — or read from live run state — makes every scoped round diverge, and the flag is sticky, so no retry clears it. Live ledger reads break it twice over: post-close status mutations would give the return-time regeneration a different ledger than dispatch saw. The frozen dispatch-scope digest removes both.
  - The byte history is this mechanism's only operand. Keeping today's overwritten nonce-only path leaves nothing to diff against and silently degrades every later round to `discovery` — safe, but delivering none of the saving. The no-delete-step rule stays: a cleanup pass removes the operand the same way.
  - Content addressing solves retention and creates an **identification** problem: today's path is a pure function of values the state document holds, which is what lets the recovery arm name the file after an interruption, whereas a digest-derived path is known only to the turn that computed it — and the two non-revision write sites persist no digest. Keeping the nonce in the leaf and recording each resolved path durably is what preserves that arm.
  - The clean scan breaks on the first `REVISE`, so teaching it to *skip* a clean `targeted` round makes it refuse rather than fall through. Scheduling a confirming whole-draft round is what reaches the clean ground, and that round has to be funded — today's spend predicate keys on a `REVISE` predecessor and would refuse it.
  - A sweep for readers of the latest completed round must be operand-agnostic: pinning the parameter name to the read-side spelling hides the mutation-side guards, including the one binding a filing override's digest to an epoch. Leaving any reader undecided is the fail-open shape this repository's guard-operand rule exists to catch. The sweep must now cover `_last_discovery_round(` as well as `last_completed(`, since #792 moved the coverage and calibration readers behind it.
  - The word "blinded" appears nowhere on this surface today; the property is carried by the information-removal paragraph and the out-of-bounds enumerations. The renderer's reads-no-run-state contract is **not** an intact anchor after this change — the renderer now reads the round's frozen dispatch-scope file — so that sentence is amended rather than cited.
  - The number 3 exists only as `_USER_ROUND_CAP`; restating any ceiling as a prose literal reintroduces the drift the earlier reconciliation removed.
  - `record-revision --stdin-digest` is required only when the latest round's last attempt is on the file arm, so an embed or inline run has no revision digest to compare — which is why the arm is one of the `targeted` selection conditions rather than an assumption.
  - The auditor's steering self-report remains a self-report by the very context steering would have primed; a `targeted` round is never described as provably steering-free.

The01Geek:
## Summary
- Every `/devflow:create-issue` Step 3.6 audit round was the same round — a cold, whole-draft, whole-repository derivation. The state owner recorded an `arm` per round and nothing about what the round was *for*, so a later round re-established from scratch the repository facts an earlier one had established, over sections no revision had touched, while the findings it should have been re-checking sat untouched in the ledger.
- Rounds now carry a tool-owned **kind** — `discovery` (cold, whole-draft) or `targeted` (a claim-scoped re-check of what a revision was supposed to fix) — derived by `scripts/issue-audit-state.py` from recorded facts alone. The orchestrator never chooses a kind, because the context that would choose it is the anchored context the audit exists to remove.
- A `targeted` round needs a byte delta, which needs a byte history, so `stage --path` became a base the helper completes with the staged bytes' own digest — a durable per-byte-state history instead of a single latest-bytes slot.

## Changes

**Round kinds (`scripts/issue-audit-state.py`)**: A closed two-member kind vocabulary with a fail-closed `_checked_kind` guard, `select_round_kind`, and the read-only `query-round-kind` (kind + the reason token that selected it + the delta state + enumerated claim ids). `record-dispatch` now **requires** `--kind` and refuses any kind other than the one the tool selects at that moment (`kind-mismatch`), mirroring the existing `write-path-mismatch` cross-check. Selection fails toward the expensive kind: `targeted` requires all five of a revision postdating the last completed round, a `file`-arm latest attempt, dispatch bytes recoverable from the byte history with a matching digest, a non-empty enumerated claim set, and a scope-basis digest equal to the canonical file's digest at dispatch — every other input, including every unestablished one, selects `discovery` and names the failing condition. An empty changed-section set and one whose computation errored both select `discovery`; neither is read as "nothing changed".

**Durable byte history (`scripts/stage-draft-write.py`)**: `stage --path` is a **base** completed with the staged bytes' digest, reporting the resolved path — a caller cannot compose that leaf, since the digest is computed from stdin inside `stage` and each shell fence is a fresh process. A second stage of different bytes lands beside the first; re-staging identical bytes resolves to the same path and leaves exactly one artifact. The new `record-staged-write` / `query-staged-write` pair records and resolves that path durably, so an interrupted or compacted turn recovers the artifact's name from recorded state rather than that turn's stdout, and the write-failure recovery arm re-applies the artifact **that write recorded** rather than the newest on disk.

**Dispatch-scope file**: A `targeted` round's whole payload — the enumerated claims as id plus one-line summary, and the tool-derived changed-section set — travels in one file whose path **and** content digest both join the round's closed recorded regeneration tuple, frozen at dispatch. The identity-data floor (`_forged_protocol_token`) sits at the single state-owner write site, before any write, so the closed token vocabulary stays in one module; the renderer imports stdlib only and refuses an empty claim set independently, so a scoped round cannot pass vacuously. Absent/unreadable and tampered scope files take **distinct** named steering reasons, because they send a reader to opposite remedies.

**Renderer (`scripts/render-audit-prompt.py`)**: A `targeted` render block carrying each claim as id plus summary and the changed-section set, with a per-claim closed-set verdict contract (`addressed` / `not-addressed`), an out-of-scope-observation channel, and an explicit withheld-field list — no finding status, severity, disposition, prior verdict, rationale or evidence reaches the auditor. Claim summaries are declared as data to classify, never instructions to obey.

**Per-reader kind treatment**: `_last_discovery_round` was widened rather than duplicated, exactly as its own docstring said a second non-discovery kind should. A clean `targeted` round never grounds `evaluate_eligibility`'s clean scan, never backs coverage, never wipes the calibration signal and never reports `basis=adjudicated`; when every enumerated claim returns `addressed` the tool selects a **confirming whole-draft round** through a new `confirm-whole-draft` next action funded from a counter of its own (the shared automatic pool is exhausted by two revision cycles, so the confirming round would otherwise be refused outright). A scoped round records **no ledger of its own** — its per-claim return updates the existing entry each claim names — so a mechanism built to reduce rounds cannot inflate the count that fires the offer for more of them. `_MAX_AUTOMATIC_REAUDITS` is left unchanged (issue #827, which proposed raising it, was closed as not planned).

**Audit summary**: `summary_fields` grounds its verdict and class counts on the latest **whole-draft** round rather than `last_completed`, and names the scoped round beside them under a new `scoped_round=` token. The selector differs from `_last_discovery_round` by exactly one clause — a final-byte exact-byte pass *is* whole-draft evidence here, whose verdict issue #792 renders on purpose.

**Out-of-bounds enumerations**: The dispatch-scope artifact is added as a glob to **both** arms' closed enumerations (it must persist, because its digest is recompared at `record-return`), and is total over a round's own scope file. The generated instruction file joins the **embed-arm** list only — on the file arm it is the artifact the auditor is directed to read and hash, so naming it would order the auditor to void its own return. The dispatch-instructions block, which is line-wrapped and carried no comparand, gains one asserted against the **rendered** output rather than a source grep.

**Docs**: `docs/DEVFLOW_SYSTEM_OVERVIEW.md` gains the canonical round-kind sections and reconciles `schema_version` (current value 3; the superseded 2 is marked as the historical value it carried when #705 landed). `CLAUDE.md`'s create-issue bullet states the lifecycle is owned by `scripts/issue-audit-state.py` and points at the overview as canonical. A `docs/cutovers/` record covers the deleted superseded prose.

## Resolves
Resolves #793

## Test Plan
- [ ] `lib/test/run.sh` is green end-to-end with zero failures and zero skips on the committed tree.
- [ ] Round-kind selection is driven row-by-row over each of the five `targeted` conditions failing independently, plus the no-completed-round, empty-delta and delta-errored inputs, asserting both the selected kind and the reason token naming the failing condition (`lib/test/test_python_scripts.py`).
- [ ] The `record-dispatch` kind cross-check is exercised in both directions — a matching kind is accepted, a mismatching one is refused with a named `kind-mismatch` breadcrumb — and a retry is validated against the round's **recorded** kind rather than a fresh selection.
- [ ] The staging round-trip is driven end to end: `stage` → new process → `emit` → `apply` resolves the same artifact from recorded state with the staging turn's stdout discarded, and a run holding several staged artifacts re-applies the one that write recorded.
- [ ] The scope-file arms are asserted separately — absent/unreadable records its own named reason, a tampered file routes to `diverged`, and a post-dispatch ledger mutation leaves the regeneration `verified`.
- [ ] `render_instructions` purity is asserted over the widened recorded tuple, and pre-#793 recorded tuples regenerate byte-identically (`lib/test/test_render_audit_prompt.py`, 96 tests).
- [ ] The `targeted` prompt is asserted to contain none of the withheld fields, to refuse an empty claim set, and to render an instruction-shaped claim summary as quoted data without changing the per-claim verdict contract.
- [ ] The audit summary is driven over a discovery→targeted run, a targeted-only run, a final-byte-pass run and a no-targeted-round negative control, asserting the verdict and every class count come from the whole-draft round and that `scoped_round` names the scoped one.
- [ ] Both out-of-bounds enumerations are pinned at their count-locked comparands, and the dispatch-instructions declaration is asserted against the **rendered** output.
- [ ] Schema coexistence is asserted in both directions: a state file written by the post-change build loads under the prior build and behaves as it does today.

## Post-Merge Verification
- [ ] `docs/create-issue-context.md` gains a measurement-obligation record for this change under its existing *maintainer measurement obligation* tier, carrying the same **unfilled placeholder shape** that tier already uses, stating that no CI check re-derives the figures and that a measurement producing no decrease means the change is reverted or deferred rather than shipped as a reduction. The figures are filled **post-merge by the maintainer**, not in the implementing PR — capturing an after-corpus requires running the instrument over real transcripts produced by the merged change, which no granted command can do during implementation.

## Visual Changes
N/A

## Breaking Changes
None. `SCHEMA_VERSION` is held at its current value, every added field is additive and read with a default, and no added key joins the required top-level key set, so a state file written by either build loads under the other. The one new **required** operand is `record-dispatch --kind`, which is filled from the tool's own selection through the `next_call=` answer line rather than by the caller.


The01Geek:
## Review findings addressed — all four Important items

Pushed as `868c270c`.

### Important 1 — reproduced, then fixed

The divergence **does** reproduce. Verified at HEAD `2c0b4e61` on all four links by direct source read, then reproduced executably before changing any code: a test driving `discovery → REVISE → targeted(VERDICT: FILE, no --claim-verdicts) → record-dispatch round 3` went RED with `rc=1, confirming_rounds_used=None` — `query-next-action` answered `confirm-whole-draft` and `record-dispatch` then refused the round it scheduled as *not funded*.

Fix is the prescribed one: the funding `elif` now gates on `_round_kind(prev) == 'targeted' and prev.get('outcome') == 'FILE'` — the same predicate `next_action` schedules under.

Three same-commit consequences:

- `_targeted_all_addressed` lost its only production caller and is **removed**, along with the two test rows asserting it as the funding gate (their wording — *"never schedules the confirming round"* — became false) and the `_validate` comment naming it the **SOLE** gate on that budget.
- Suggestion 1's co-located false comment is corrected — it is now true of the shipped predicate and records why the narrower sweep was wrong.
- The changeset entry and `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s round-kind section both said the confirming round fires *"when every enumerated claim returns addressed"*; both now state the shipped rule.

### Importants 2–4 — coverage added

Each new test is driven at its executable surface and paired with a **companion row** proving the assertion is attributable to the guard rather than to some unrelated precondition.

- **AC32 limb one** — `evaluate_eligibility`'s clean-scan `targeted` skip, in both directions: a clean scoped round never becomes the clean ground, and a trailing scoped `REVISE` never revokes an earlier whole-draft clean verdict.
- **`_convergence_basis`** — a trailing FILE-adjudicated `targeted` round never answers `basis=adjudicated`; one trailing a whole-draft round is skipped rather than superseding it.
- **AC18 positive arm** — an untampered scope file regenerates to the recorded identity, so the round records steering as `established`, asserted **directly** on the `record-return` answer line *and* on the persisted round record.

### Mutation table

Each mutation applied alone to `scripts/issue-audit-state.py`; full `lib/test/test_python_scripts.py` run each time. Unmutated tree: **2965 passed, 0 failed**.

| # | Mutation | Verdict |
|---|---|---|
| M1 | funding predicate re-narrowed with `and prev.get('claim_verdicts')` (= pre-fix HEAD behavior) | **RED** — the end-to-end unusable→confirming dispatch row (`rc=1`, counter `None`). This is the reproduction. |
| M2 | `continue` → `pass` in `evaluate_eligibility`'s targeted skip | **RED** — both new AC32 rows |
| M3 | `continue` → `pass` in `_last_discovery_round`'s targeted skip | **RED** — both new `_convergence_basis` rows (plus 3 pre-existing rows) |
| M4 | drop the frozen scope payload from `regenerate_instructions_digest` (`scope_text` never assigned) | **RED** — both new AC18 rows **only**; the #709 non-scoped positive control stays GREEN, so the mutation is precise to the scoped-round regeneration these rows were added for |

### Verification

- Full suite, foreground, clean tree: **`13184 passed, 0 failed`** — zero skips.
- `ruff check` clean on both changed `.py` files. No `.sh` touched, so no shellcheck delta.

### Not changed (flagged, not guessed)

- **Pre-existing truncated comment** in `lib/test/modules/create-issue-contract.sh`, ending *"the seams carry an SC2034 disable only while nothing in"*. Present on this branch before my change; I did not guess its intent and left it as-is.
- `record-return`'s UNUSABLE warning still says *"the next action selects a discovery round"*. Ambiguous rather than clearly false — the confirming round **is** dispatched with `kind=discovery`, while the `next_action` **token** is `confirm-whole-draft`. Left alone deliberately.
- Suggestions 2–5 are out of scope for this pass.

Issue #810, opened 2026-07-25
Prevent wording-only pin tests from regrowing after the corpus cleanup


The01Geek:
## Dependencies

Blocked by #809 — the final cleanup establishes the post-retirement corpus and removes the issue-248 static pins whose commentary this issue reconciles.

## Problem Statement

The pin-retirement work removes wording-only and prose-presence assertions that made harmless edits expensive and encouraged tests of implementation text instead of behavior. The current authoring and CI controls do not prevent that class from returning. An implement run can add the same kind of assertion, attach a plausible `# structural-pin-ok:` comment, and pass the required test job.

Maintainers need the cleaned corpus to remain stable without relying on implementer memory and recurring manual censuses.

## Current Behavior

- Behavioral-fix pins are routed through mutation-taking helpers.
- Added non-mutation pins require a `# structural-pin-ok:` marker.
- The marker parser checks for the marker substring in a real comment but does not validate a category plus a non-empty rationale.
- The production census classifier is deliberately maintainer-run and absent from `lib/test/run.sh`.
- Mutation-routing setup failures become skips, while the suite exit status depends only on the failure tally.
- Implement guidance still directs authors to pin mirror-fact comments and to cover stub blind spots with static assertions without limiting those assertions to executable boundaries.
- Review guidance does not explicitly reject newly added wording-only presence pins.
- After #809, the issue-248 commentary in `CLAUDE.md` will describe removed static pins as present.

## Desired Behavior

New wording-only presence tests are rejected by the required CI test job. Static presence pins remain available only for an explicitly closed set of executable-boundary categories with a machine-validated category and non-empty rationale. Behavioral prompt and prose guards use mutation evidence tied to the guarded regression.

The authoring and review instructions state the same classification rule, and the required gate fails on an unestablished comparison population.

## User Impact

Maintainers can revise prompt wording, comments, and documentation without rebuilding a prose snapshot suite. Implement runs receive a clear authoring rule, reviewers receive a deterministic backstop, and legitimate schema, sentinel, routing, security, state, generated-artifact, helper-contract, and cross-phase boundaries remain protected.

## Technical Context

> **Scope note:** The files and details below are the known starting points, not the full
> list. Before implementing, trace the change through the codebase to find every affected
> call site, consumer, and layer — this issue maps the work, it does not bound it.

[...]

Issue #885, opened 2026-07-28
Retire the prose-pin stock: re-adjudicate the corpus and sweep (deferred from #876)


The01Geek:
Deferred from #876 (PR #882), which delivered only the **unlock + record** subset: AC1 (generalize the `CLAUDE.md` #843 bullet into class-level policy), AC2 (a per-pin adjudication record current with the tree), AC4 (`CONTRIBUTING.md`'s #798 rule restated so a prose-pin retirement is an accepted pin-only removal), and AC10.

That scope split was deliberate and user-confirmed. The sweep proper is this issue.

## Why this is a separate issue

Two things blocked the sweep, and #882 removed both:

1. `CONTRIBUTING.md`'s #798 rule prohibited pin-only removal outright. It now routes on an ordered, first-match-wins gate whose **arm 2** authorizes a pin-only removal for a site with `counted_occurrences < 2` whose `bucket_final` is a prose bucket.
2. The checked-in adjudication table classifies **every** site `boundary`. Nothing was retirable under the record.

So arm 2 is authorized but currently selects nothing. **This issue is the re-adjudication pass that makes it selectable, plus the sweep it enables.**

## Measured starting state (at PR #882's head)

| Quantity | Value |
|---|---|
| Census sites (`.devflow/logs/pin-corpus-inventory.tsv`) | 897 |
| Sites with `bucket_final: boundary` | 897 (all) |
| `bucket_mechanical: prose-sole-copy` | 510 |
| `bucket_mechanical: prose-multi-copy` | 146 |
| `bucket_mechanical: unclear` | 128 |
| `bucket_mechanical: required-copy` | 110 |
| `bucket_mechanical: suite-internal` | 3 |
| Adjudication rows (`lib/test/pin-corpus-adjudications.tsv`) | 851 |
| — of which carry a hand-written retention rationale | ~655 over prose-bucketed sites |
| `assert_pin_unique` call sites in `lib/test/run.sh` | 876 |
| Governance Python under `lib/test/` | 11,858 lines across 8 files |

Note the issue-#876 body's census was measured at `94203da5` and is now stale: `run.sh` is 44,741 lines (not 44,617), `assert_pin_unique` is 876 sites (not 853), and `create-issue-contract`'s `minimum_assertions` is **245**, not the 248 AC8 states.

## Deferred acceptance criteria (verbatim from #876)

[...]

PR #886, opened 2026-07-28
Cut the pin-corpus-lint block by memoizing the per-source parses

### LINKED ISSUE #881: Cut the pin-corpus-lint test fixture cost: 47 per-test git-repo rebuilds dominate the required CI check
## Problem Statement

`lib/test/test_pin_corpus_lint.py` is the single largest block in the required `lib + python tests` CI check, and essentially all of its cost is a test **fixture**, not the assertions it makes.

The fixture rebuilds a throwaway git repository from scratch for each test that needs one: it `shutil.copy2`s a set of real repository sources into a temp dir, then runs `git init`, `git add .`, and `git commit`. The dominant term is that the copied set includes `lib/test/run.sh`, which is **3,323,582 bytes**.

Measured on the current tree:

| Class | Tests | `_repo` calls | Copies the full audited set? |
|---|---|---|---|
| `StaticPinWorktreeCompositionTests` | 21 | 20 | yes |
| `RetiredPinRevivalTests` | 9 | 12 | no |
| `AdjudicationChangeScanTests` | 8 | 10 | no |
| `RetiredMutationHelperBanTests` | 5 | 5 | yes |
| `PinCorpusLint810Tests` | 61 | 0 | — |
| `AdjudicationStateTests` | 9 | 0 | — |
| `Issue687OutputRoutingTests` | 3 | 0 | — |

So **47 fixture builds per run**, of which **25 copy-and-commit the full `AUDITED_PIN_SOURCES` set**. That set alone is roughly 3.9 MB per build before the remaining audited modules and manifests are counted, so a single run copies and git-commits on the order of a hundred megabytes to assert properties of a linter.

This cost is paid **twice per CI run**, because the `harness-python-guards` module executes once on the suite's main shell and again concurrently inside the pooled `test_module_runner.py` real-runner meta-test.

## Why this is being filed now — a corrected premise

Issue #870 attacked this block by **sharding** it across concurrent workers, on the stated premise that "the runner's remaining vCPUs sit idle". PR #872 implemented that correctly and it is green, but the measurement does not support the premise.

Measuring the block directly as the contiguous log gap after `#810 pin-corpus wording-only authoring gate` (a figure insensitive to whole-job runner variance):

| Tree | Block duration |
|---|---|
| `main` @ `a7fb1119` | 517.0s |
| `main` @ `5afceaeb` | 506.5s |
| `main` @ `94203da5` | 332.2s |
| PR #872, 4 concurrent workers | **478.9s** |

478.9s falls **inside** the unmodified range of `main`. Four times the workers produced no speedup separable from noise. Whole-job totals agree: 834s for PR #872 against 609/863/882s for recent `main`.

The likely reason is that the vCPUs were never idle. The module already runs twice concurrently, and each execution spawns `git` subprocesses per test, so roughly four processes were already competing for the runner's four vCPUs before any sharding was added. There was no spare capacity to claim.

That redirects the work to the term that actually dominates: **the fixture itself**. Issue #870 listed this under *explicitly rejected alternatives* ("potentially worthwhile, but orthogonal and additive"). The measurement above promotes it from orthogonal to primary.

**Past-time snapshot.** Every figure above describes specific observed runs and is a frozen record, not a live measurement. Sources are listed under Evidence.

## Current Behavior

`StaticPinWorktreeCompositionTests._repo` (and the structurally similar helper in `RetiredMutationHelperBanTests`) performs, per invocation:

1. `shutil.copy2` of every `AUDITED_PIN_SOURCES` entry plus `lib/test/module-harness.sh`, `lib/test/pin-corpus-lint.py`, `lib/test/test_pin_corpus_lint.py`, `lib/test/pin-corpus-adjudications.tsv`, four `.devflow/logs/*.tsv` manifests, and `scripts/workflow-flight-recorder-registry.json`;
2. a synthetic fixture file write;
3. `git init -q`, two `git config` calls, `git add .`, `git commit -qm base`;
4. `git update-ref refs/remotes/origin/main HEAD`.

Nothing is shared between invocations. Two tests in the same class that need an identical starting repository each build one from zero.

## Desired Behavior

The same tests assert the same properties against the same corpus, but the repeated construction of an identical base repository is done once and reused, so the file's contribution to the required check drops substantially.

The **assertions and the audited corpus must not change** — this is a fixture-cost issue, not a coverage-reduction issue. A test that today audits the real `lib/test/run.sh` must still audit the real `lib/test/run.sh`.

## Candidate approaches

Not prescribed; the implementing run should measure rather than assume. Plausible directions, cheapest first:

- **Build the committed base repository once and copy it per test.** A single `_repo`-equivalent build cached at class or module scope, then `shutil.copytree` (including `.git`) per test. Trades N git-commits for N directory copies, which may or may not win given the same bytes move.
- **Build once and branch instead of copying.** Tests that only need a distinct working state may be able to share one repository and isolate via `git worktree add` or a fresh branch, removing both the copy and the commit.
- **Shrink what is copied.** Many audited entries may be irrelevant to a given test's assertion. If the linter can be pointed at a smaller corpus without weakening what is checked, the 3.3 MB `run.sh` copy is the first candidate — but only where the test does not actually depend on its content.
- **Cheapen the commit.** `git add` of a 3.3 MB file hashes and zlib-compresses it every time. A pre-built object store, or `git commit-tree` against a cached tree object, may avoid re-hashing unchanged content.

## Constraints

1. **Cross-test isolation is currently guaranteed by construction and must stay guaranteed.** Every filesystem-touching test allocates its own `tempfile.TemporaryDirectory()` and passes an explicit `cwd=`; there is no `os.chdir` in the file. Any shared or cached fixture must not let one test observe another's mutations — several tests deliberately `git checkout`, create branches, and delete files inside their repo.
2. **The coupled assertion-count triple.** `harness-python-guards` carries `minimum_assertions: 39` in `scripts/workflow-flight-recorder-registry.json`, the same `39` as the third operand of its `devflow_run_full_suite_module` call in `lib/test/run.sh`, and its emitted tally. These are an equality contract with zero slack; if the count moves, all three move in the same change.
3. **`lib/test/run.sh` may not mention `test_pin_corpus_lint.py`** — not even in a comment. The module asserts `grep -cF 'test_pin_corpus_lint.py' lib/test/run.sh` equals `0` (issue #865).
4. **The file is executed twice concurrently**, so any per-process cache is built twice and any on-disk shared cache must be safe under concurrent access by independent processes.
5. If PR #872 has merged, the file now runs as one process **per test**, so a cache scoped to the process or the class is rebuilt per test and buys nothing. The caching scope must be chosen against whichever driver is actually in place, and the implementing run should confirm which that is rather than assuming.

## Out of scope

- Reducing the audited corpus, retiring pins, or deleting tests to hit a number.
- Re-litigating the sharding driver from #870 / PR #872. If sharding turns out to be unnecessary once the fixture is cheap, that is a separate cleanup decision made against a measurement, not here.
- The broader CI job-matrix split (#877), which is independent and additive.

## Acceptance Criteria

- [ ] The number of tests executed by `lib/test/test_pin_corpus_lint.py` is unchanged, and the set of sources each test audits is unchanged.
- [ ] The `harness-python-guards` module remains green, and RED when any single test in the file fails.
- [ ] The coupled triple (registry `minimum_assertions`, the `run.sh` boundary operand, the emitted tally) is mutually consistent after the change.
- [ ] `grep -cF 'test_pin_corpus_lint.py' lib/test/run.sh` still returns `0`.
- [ ] Cross-test isolation is preserved: a test that mutates its repository cannot affect another test's repository. This is demonstrated by an executed test, not asserted in prose.
- [ ] The block's duration is reported from a real CI run and compared against the `main` baselines recorded under Evidence. The figure is reported as a snapshot, not pinned as a constant.
- [ ] A full `lib/test/run.sh` run reports zero failed, with no skip beyond the known `#671 claude plugin validate --strict [blocking-gate] — claude CLI not on PATH` skip.

## Evidence

Past-time snapshots, each traceable to the run that produced it:

- Block durations on `main`: 517.0s (`a7fb1119`, run `30315000855`), 506.5s (`5afceaeb`, run `30314976912`), 332.2s (`94203da5`, run `30312537348`).
- PR #872 (4 concurrent workers): block 478.9s, suite step 834s, job `90150136272`, head `6c638542`. Its own log line reports `test_pin_corpus_lint.py: executed 116 test(s) across 4 concurrent worker(s) (116 enumerated)`.
- Whole-job suite-step totals on `main`: 609s / 863s / 882s across the three runs above — a spread wider than the effect #870 sought, which is why the block gap is used as the measurement rather than the job total.
- Fixture counts and byte sizes were derived from the tree at the time of filing and will drift as the file changes.

## Related

- #870 / PR #872 — sharding the same file. Complementary, and the source of the corrected premise above.
- #877 — splitting the required check into a concurrent job matrix. Independent of this.

The01Geek:
## Summary
- Cuts the cost of `lib/test/test_pin_corpus_lint.py`, the largest block in the required `lib + python tests` check, by memoizing the pure per-source parses in `pin-corpus-lint.py` and `mutation-pin-census.py` and removing duplicate parse passes.
- **Issue #881's cost attribution is refuted by measurement and the change targets the real term instead.** Instrumenting the file gave fixture construction 10.5s of a 250s run (78 builds, mean 0.134s — 4%) against 220.0s in the linter scan (68 calls, mean 3.236s — 88%). The issue's four candidate approaches all target the 4% term, so the git fixture is left untouched and its isolation-by-construction intact.
- Measured on the merged base with the post-#872 sharded driver: `harness-python-guards` 2:30.66 → 1:33.20 wall and **263.6s → 150.4s user CPU (−43%)**, with peak RSS on the heaviest worker below the pre-change baseline. User CPU is the figure that translates on CI, since #881's own finding is that the runner's 4 vCPUs were already saturated.

## Changes

**`lib/test/pin-corpus-lint.py`**: Derive function-definition line numbers from one newline-offset index via `bisect` instead of re-counting the text prefix per definition. Memoize the two pure, repo-root-independent per-source parses behind defensive copies. Tokenize each logical line once and share it between the wrapper-invocation and site passes, retaining tokens only for lines that can carry a site. Take the assignment-map snapshot only where an assignment could change it, sharing one `MappingProxyType` view across the run between two — this is what removes the change's original peak-memory regression. Gate the second, span-aware tokenization behind a literal `grep` in the line. Load the census module once per process. Drop `helper_specs_for_source`'s `include_families`/`include_origins` flags, whose sole caller passed both.

**`lib/test/mutation-pin-census.py`**: Split the per-file derivations out of `_definition_counts` and `_extract_source` so each is a pure function of a source's name and text, then memoize those plus the logical-line decomposition. The cache bound is sized against the set one sweep actually visits — sizing it against the audited subset was measured to miss, because the sweep is ordered and evicts an audited source before the row extraction asks for it again.

**`lib/test/test_pin_corpus_lint.py`**: Add `MemoizedParseContractTests` (defensive copies, the read-only views, snapshot advancement, and the census's within-build reuse) and four `StaticPinWorktreeCompositionTests` tests — three cross-repository memo-leak probes plus the filesystem-isolation test.

**`lib/test/modules/harness-python-guards.sh`**, **`docs/DEVFLOW_SYSTEM_OVERVIEW.md`**: Reconcile the sharding-safety rationale with the newly process-global memos, and record the contract in the overview as a pointer to the coupled pair rather than a third copy.

## Resolves
Resolves #881

## Test Plan
- [x] Output equivalence against the pre-change modules: `extract_guard_sites` identical across 42 sources, `extract_pins` identical across 26 sources under two override shapes, `variable_maps_by_line` value-identical across all 38,394 logical lines of `run.sh`, and the full census identical.
- [x] Each memo-leak probe verified RED under a simulated mis-keying of the memo it names — `_function_definitions_cached` and `_helper_specs_for_source_cached` constant-keyed, census `_definition_scan` path-keyed, census `_extract_rows` source-keyed — with the unmutated control green.
- [x] The census-reuse pin verified RED at a shrunk cache bound, which also closes the gap that deleting every `lru_cache` decorator previously left the suite green.
- [x] `bash lib/test/run-module.sh harness-python-guards` → `38 passed, 0 failed`, and the coupled triple (registry `minimum_assertions` 38, the `run.sh` boundary operand 38, the emitted tally) is mutually consistent.
- [x] `grep -cF 'test_pin_corpus_lint.py' lib/test/run.sh` returns `0`.
- [x] Full `lib/test/run.sh`: **12744 passed, 0 failed, 0 skipped**, plus all three `CLAUDE.md` lint gates clean (shellcheck over tracked `*.sh`, the `--extended-analysis=false` pass over `lib/test/run.sh`, and `ruff` over tracked `*.py`).
- [x] Block duration reported from CI — see *Measured effect* below.

## Measured effect

**CI block, this PR** (run `30331269295`, the required `lib + python tests` job): **332.0s**, measured as the contiguous log gap after `#810 pin-corpus wording-only authoring gate`, executing **125 tests** (116 pre-existing + 9 added). Past-time snapshot, not a pinned constant.

Against `main` in the **same contended single-job topology** — the topology this branch's base still has, since PR #883's shard matrix has not merged:

| Tree | Block | Tests |
|---|---:|---:|
| `main` @ `a7fb1119` | 517.0s | 116 |
| `main` @ `5afceaeb` | 506.5s | 116 |
| `main` @ `94203da5` | 332.2s | 116 |
| PR #872 @ `6c638542` | 478.9s | 116 |
| PR #872 @ `f74d7a7f` | 405.6s | 116 |
| **this PR** @ `30331269295` | **332.0s** | **125** |

**Read this honestly: one CI sample does not separate the effect from runner noise.** `main`'s own spread is ~185s, and 332.0s sits at the bottom of but inside it — the same standard issue #881 applied to PR #872's two samples applies here. What it does show is that the block ran 9 extra tests without moving out of the bottom of that range.

The controlled evidence is a **local like-for-like A/B** on one machine, back to back, same command, the three changed files checked out of `origin/main` and back:

| | wall | user CPU |
|---|---:|---:|
| before | 2:30.66 | 263.6s |
| after | 1:33.20 | **150.4s (−43%)** |

User CPU is the figure that translates to CI, because issue #881's own finding is that the runner's 4 vCPUs were already saturated — on a saturated runner it is CPU, not wall clock, that is the binding resource.

The **291–306s isolated-shard figures** from PR #883 are a *different* topology and are not comparable to the rows above; this branch's base does not carry that matrix.

## Visual Changes
N/A

## Breaking Changes
None. The change is confined to `lib/test/`, which `install.sh` does not vendor, so no consumer observes it — hence no changeset, per the repo's internal-only rule.


The01Geek:
**Sequencing note (issue #881 correction 2).** This PR is marked ready rather than held for #885's residual census, by explicit decision.

Rationale: the change touches neither `AUDITED_PIN_SOURCES` nor the git fixture — it makes the linter's *scan* cheaper, so every `_repo` build that survives #885's sweep gets cheaper regardless of how many survive. The sweep therefore cannot invalidate it. The one real interaction is a likely merge conflict in `lib/test/pin-corpus-lint.py`, which #885 also edits; whichever lands second resolves it.

Note for that resolution: this PR's edits to that file are structural (extracted `_function_definitions_cached` / `_helper_specs_for_source_cached`, a `bisect` line index, one shared per-line tokenization, copy-on-assignment snapshots in `variable_maps_by_line`, and a `grep` gate on the span tokenization). None of them depends on which pins remain in the corpus.

PR #892, opened 2026-07-28
Materialize the review tier's prompt extensions from the trusted base ref

### LINKED ISSUE #874: Materialize the review tier's prompt extensions from the trusted base ref
## Problem Statement

The merge-gating cloud reviewer extends its own prompt with bytes read from the PR-head checkout. `scripts/load-prompt-extension.sh` resolves `.devflow/prompt-extensions/<skill>.md` against the git repo root, `.github/workflows/devflow-runner.yml`'s review job checks out `ref: ${{ inputs.ref }}` (the PR head SHA), and `skills/review/SKILL.md` instructs the agent to treat whatever that helper prints as instructions appended to its own prompt for the run.

The PR-head-supplied inputs DevFlow has already moved behind a trusted-source boundary — the deny-floor helper source (issue #404), the git-env helper, the plugin-compose helpers (issue #505), and the Stop-hook target closure (issue #458) — are each materialized from the base ref into `$RUNNER_TEMP` before the agent starts. The reviewer's *appended prompt extension* is one such input still read from the pull request, and it is a strictly more powerful lever than the tool floor those hardenings protect. It is not the only one: the plugin tree itself remains PR-selectable through `.claude-plugin/marketplace.json` and the baked `./` marketplace entry, which is a distinct and larger defect filed separately — this issue closes the extension channel and states that neighbouring channel as a residual rather than claiming to close it.

The people this hurts are maintainers of any repository where `Devflow Review` is a required status check and the merge decision rests on the bot's verdict. The population that can land bytes in that prompt is same-repo pull-request authors (write-access collaborators) and allowlisted bots — precisely the population issue #404 hardened the deny floor against.

## Current Behavior

On the review tier, `scripts/load-prompt-extension.sh` composes its path from the checked-out working tree and prints the file verbatim:

```sh
_devflow_root="$(git rev-parse --show-toplevel 2>/dev/null)" || _devflow_root=""
if [ -z "$_devflow_root" ]; then
    _devflow_root="$(pwd)"
fi

ext_file="${_devflow_root}/.devflow/prompt-extensions/${skill}.md"
```

The helper's own header states the assumption that no longer holds on this tier: *"The argument is the only attacker-influenceable input …; the file's bytes are trusted."* On a PR-head checkout the bytes are attacker-influenceable too.

Two reads on this tier resolve that way, not one: `skills/review/SKILL.md`'s own preflight loads `review`, and `skills/review/phases/phase-3-agents.md` composes the dispatched final-pass reviewer's prompt with `.devflow/vendor/devflow/scripts/load-prompt-extension.sh requesting-code-review`, whose findings feed the verdict.

The review profile's `TOOLS=` literal grants the helper directory-agnostically, and the workflow comment justifying that grant reasons only about tree mutation: *"load-prompt-extension.sh only `cat`s a consumer-owned .devflow/prompt-extensions/<skill>.md and prints it"*. Tree non-mutation is a different property from prompt authorship.

`.devflow/prompt-extensions/review.md` is an existing tracked file, so a hostile edit is an ordinary diff to a normal file rather than a conspicuous new addition.

The one existing provenance control lives in `skills/review-and-fix/references/shadow-review.md`, and its own text scopes it away from this problem: it runs on the `/devflow:review-and-fix` shadow pass, it is advisory (the extension *"is still loaded"* when the check fails), and it is explicitly *"checked without a base-ref read"*.

## Desired Behavior

On the review tier, the bytes **appended to the reviewing agent's prompt** come from the trusted base ref, the plugin version that supplies the reader is chosen by the base ref rather than by the pull request, and the part of the mechanism that *suppresses* PR-head bytes depends on nothing that can fail.

`.github/workflows/devflow-runner.yml` resolves `devflow_version` for the `vendor-plugin` fetch from the **trusted base ref**, joining `provision_env`, the `allowed_tools` list, and the `setup` block, which that job already reads only from the base branch. A pull request therefore no longer selects which plugin commit — and so which `load-prompt-extension.sh` — reviews it. That read is performed by a step placed **above** the `vendor` step, not by `baseprovision`, which the job declares after `vendor`: a reference to a later step's output resolves to the empty string, and `vendor-slice.sh` then refuses to run with an empty ref.

The same workflow then performs four **unconditional** actions on every review run, outside any fetch-success branch and with no dependency on a helper resolving: it creates `$RUNNER_TEMP/devflow-trusted-prompt-ext/`, it creates the workspace directory `.devflow/prompt-extensions/`, it truncates to empty the workspace copy of each extension in the protected set — creating an empty file when the pull request carries none — and it publishes `DEVFLOW_PROMPT_EXTENSION_ROOT` pointing at the closure through `$GITHUB_ENV`, the job-scoped channel whose reach into the Claude Code CLI subprocess this workflow's git-env pins step already records as measured rather than assumed. Inside the existing base-ref fetch-success branch it then *populates* that closure from the base ref. `scripts/load-prompt-extension.sh` honors `DEVFLOW_PROMPT_EXTENSION_ROOT` when it is set and non-empty, and falls back to today's repo-root resolution when that variable is unset, and likewise when it is set to the empty string.

The directory the reviewer reads from is therefore the trusted closure on every arm, and every failure degrades to that closure being empty rather than to the repo-root path being taken.

Two outcomes leave the closure without a given file, and they are distinguished:

- **The base ref carries no such extension.** This is the ordinary shape for a consumer that never committed one. Nothing is materialized for that name, the run proceeds with no extension text, and **no warning is emitted** — there is nothing wrong.
- **The base ref was read and the file is there, but the closure was not populated** — from the closed set {unwritable `$RUNNER_TEMP`, a read that failed for a reason other than the object being absent}. The run proceeds with no extension text and the job emits a `::warning::` naming which applied, mirroring the deny floor's existing `::warning::devflow deny-list floor helper (filter-runner-tools.sh) not found at any TRUSTED source` arm.
- **The base ref's content was never established** — from the closed set {base-ref fetch failure, empty `BASE_REF`, no trusted source for the materialization helper}. Whether an extension exists on the base ref is unknowable on these arms, because every base-ref read in this job sits inside the fetch-success branch, so the run cannot say which of the two arms above it would otherwise have taken. It emits a distinct notice — the trusted materialization was never attempted and the reviewer is running with no extension text — rather than a reason-naming warning that would assert a fact the run never established. That notice is a `::warning::` on every such run, which is correct: a base ref that cannot be read is not the ordinary shape.

The unconditional truncation remains the guard for the case where the base ref pins a plugin version whose loader predates this change: that loader resolves the repo-root path, finds the truncated file, and loads nothing. With `devflow_version` base-ref-resolved, which loader runs is a maintainer's pin rather than a pull request's choice.

`scripts/load-prompt-extension.sh` emits a stderr breadcrumb naming the directory it resolved and which branch selected it, and `skills/review/phases/phase-3-agents.md`'s `EXTENSION-STATUS:` contract is amended in the same change to classify on **stdout** explicitly and to carry the dispatched reviewer's own resolved root as a required field — so a run in which the variable failed to reach the orchestrator's shell, and equally one in which it failed to reach the dispatched Task's shell, is visible rather than silent.

The `/devflow:implement` tier keeps its current behavior. The `/devflow:review-and-fix` tier keeps its current behavior for its own two extension loads and is left knowingly unprotected for the Phase-3 load it performs after checking out the PR head; the reason is recorded in the residual list below rather than asserted as an ordering property.

## User Impact

Maintainers gating merges on `Devflow Review` get a reviewer whose instructions come from their own base branch, read by a plugin version their own base branch pins. A consumer whose base ref carries a committed `review.md` **and** a `devflow_version` at this change onward keeps it working unchanged **once the environment-propagation measurement below confirms the variable reaches the reader** — that measurement is pending, and until it lands this continuity is an expectation rather than a guarantee; a consumer whose base ref pins an earlier version loses the extension until they bump, which the residual list below states rather than glosses. A repository whose base-ref fetch fails sees a named warning in the job log instead of a silently degraded review, and one whose runner does not propagate the variable sees the loader's resolved-root breadcrumb instead of a silent no-op.

## Technical Context

**Relevant Classes/Files**

- `scripts/load-prompt-extension.sh` — the root-resolution block, the new breadcrumb, and the header paragraph asserting that the file's bytes are trusted.
- `scripts/devflow-cloud-writer-contract.json` and `lib/test/cloud_writer_contract.py`, with `lib/test/cloud_writer_deps.py` — the SHA-256 pin over the loader's bytes, the generator that refreshes it, and its dependency side.
- `.github/workflows/devflow-runner.yml` — the `Checkout repository` step, the `cfg` and `extract` steps that resolve `devflow_version`, the `vendor-plugin` step that consumes it, the `baseprovision` step (its `BASE_REF` env, its fetch-success branch, and its `FLOOR_HELPER`/`FLOOR_SOURCE` initialization), the `harden_hooks` step's `displaced_paths` output and the `compose` step's `HARDENED_PATHS` binding, the read-only-safety comment block above the review `TOOLS=` region, and the `Run Claude Code` step.
- `scripts/render-grounding-block.sh` and `skills/review/phases/phase-0-setup.md` — the issue-#504 displaced-path attribution chain (Phase 0.1's attribution rule and Phase 0.1.5's `.devflow/tmp/displaced-paths.txt`).
- `skills/review/SKILL.md` — the standardized extension-load step and the sentence directing the agent to treat the printed text as appended instructions.
- `skills/review/phases/phase-3-agents.md` — the `{SUPPLIED_EXTENSION_COMMAND}` substitution and the `EXTENSION-STATUS:` return contract.
- `skills/review-and-fix/references/shadow-review.md` — the authoritative statement of the shadow pass's non-base-ref provenance check, of which `docs/shadow-review.md` is the documentation mirror; its account of what that check is for changes once a base-ref materialization exists on the review tier, so the pair moves together.
- `.github/workflows/matcher-probe.yml` — the re-runnable evidence source for the environment-propagation measurement.
- `lib/test/modules/prompt-extension-reader.sh`, its checked-in provenance record `lib/test/modules/prompt-extension-reader.inventory.md`, its registry entry under `test_modules` in `scripts/workflow-flight-recorder-registry.json`, and the loader's entry in the focused-test map `lib/test/modules/coverage-map.json`.
- `lib/test/run.sh` — the `#402`/`#404`/`#458` trusted-materialization pin block and its `emit_tools` driver.
- `.devflow/config.schema.json` and `scripts/emit-git-env.sh` — the two sites carrying the issue-#295 repo-root-reader enumeration that names the loader.
- `install.sh` — the workflow copy loop, and the absence of any `.devflow/prompt-extensions/` creation.
- `lib/intervention-surfaces.md`, `CONTRIBUTING.md`, `CLAUDE.md` — prose describing extensions as consumer-owned, silent on checkout provenance.

**Verified: the loader has neither an environment-variable surface nor a file-pointer surface today.** `scripts/load-prompt-extension.sh` runs under `set -euo pipefail`, accepts one positional `SKILL_NAME` plus an optional `--section '<## heading>'`, refuses any other `--`-prefixed token with exit 2, and reads no environment variable at all. `claim=loader-no-env-surface revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=2cae98a3f73a47205f0e54ac3a72811ebc728dc7`

**Verified: the review job's only checkout is the PR head, and the plugin version is resolved from that checkout.** `.github/workflows/devflow-runner.yml`'s `Checkout repository` step passes `ref: ${{ inputs.ref }}`, whose input description reads *"PR-review callers MUST pass the PR head SHA so the run sees the PR's own content"*; the `cfg` step reads `.devflow/config.json` from that tree, the `extract` step emits `devflow_version` from it (`echo "devflow_version=$(echo "$CONFIG_JSON" | jq -r '.devflow_version // empty')"`), and the `vendor-plugin` step consumes it as `ref: ${{ steps.extract.outputs.devflow_version }}`. The adjacent comment records the accepted residual this issue closes: *"Only `ref` is PR-influenced; the source repo is fixed, so a PR can at most select an official-repo commit."* `claim=review-tier-checks-out-pr-head revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=2baa79740811f21e83780e33f4f38ddff69744e9`

**Verified: the sibling tiers' protection, stated precisely.** `.github/workflows/devflow.yml`'s `config`, `gate`, and `command` jobs and `.github/workflows/devflow-implement.yml`'s `config`, `gate`, and `claude` jobs all use `actions/checkout@v6` with no `ref:`. The implement workflow states its own case: *"The implement job needs no such step: it checks out the DEFAULT branch (never a PR head), so its hook scripts are already trusted."* For the command tier the operative property is that **standalone `/devflow:review` performs no PR checkout at all** — `skills/review/phases/phase-0-setup.md` scopes the local-head path to a wrapping skill (*"When `head_override` is absent (standalone `/devflow:review`, the default) use the API head as above; do **not** diff against local `HEAD`"*) and fetches the diff through `gh pr diff` — so the tree stays at the default branch for the whole run. For `/devflow:review-and-fix` the picture is mixed: its own two loads happen at skill start, but `skills/review-and-fix/references/loop-control.md`'s Step 0.5 runs `gh pr checkout $PR_NUMBER` **before** Step 1 runs the shared engine, so that tier's Phase-3 `requesting-code-review` load resolves against the PR head. That is recorded as a residual below, not as protection. `claim=sibling-tiers-default-branch revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=de907ee684118f030adbaa772020842477687ff7`

**Verified: two extension reads on the review tier resolve from the PR head, and they sit at different environment depths.** `skills/review/SKILL.md` carries the standardized `load-prompt-extension.sh review` step, run in the orchestrator's own shell. `skills/review/phases/phase-3-agents.md` embeds `.devflow/vendor/devflow/scripts/load-prompt-extension.sh requesting-code-review` inside `{SUPPLIED_EXTENSION_COMMAND}`, and its own text states that *"The dispatched final-pass reviewer runs as a `general-purpose` Task"* — so that load runs one environment hop further out, inside a dispatched subagent's shell. `claim=two-pr-head-extension-reads revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=5f393b2c75b114f501d1f8aab6173980add80be0`

**Verified: four trusted `$RUNNER_TEMP` closures exist, all in one file, and none covers prompt extensions.** Executed whitespace-normalized repo-wide search over the git index:
`git ls-files -z | xargs -0 python3 -c '<collapse \s+ to one space; report files matching devflow-trusted-[a-z]+>'`
Hit list, complete: `.github/workflows/devflow-runner.yml` — 19 hits, names `devflow-trusted-compose`, `devflow-trusted-floor`, `devflow-trusted-gitenv`, `devflow-trusted-hooks`. `claim=trusted-closure-inventory revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=066bc05e3da0b2eb11df7b3093feef187117b40e`

**Verified: nothing in the tracked tree — no pin, no comment, no documentation sentence — associates the prompt-extension path with base-ref trust.** Executed whitespace-normalized repo-wide co-occurrence search (any `prompt-extension` match within 300 normalized characters of `FETCH_HEAD`, `RUNNER_TEMP`, `devflow-trusted`, `trusted base`, `base-ref`, `base ref`). Hit list, complete: `.github/workflows/devflow-runner.yml` — 1 hit, inspected and incidental (a grant token adjacent to an unrelated provisioning comment); `skills/review-and-fix/references/shadow-review.md` — 2 hits, the paragraphs that state the shadow check is performed *without* a base-ref read. `claim=no-trust-source-coverage revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=d5aad881d3f36feec8908db4365cd308016c27c5`

**Verified: `Bash(git checkout:*)` is granted on the review profile**, which is why the trusted copy is placed outside the PR-writable workspace rather than written over the workspace copy alone, and why the truncation alone was never sufficient as a boundary. `lib/review-profile.tokens` carries `Bash(git checkout:*)` alongside `Bash(*/load-prompt-extension.sh:*)` and `Bash(.devflow/vendor/devflow/scripts/load-prompt-extension.sh:*)`. `claim=review-profile-git-checkout-grant revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=4d98cc33b61c0afb35b82a316c47b59ac2ca9a92`

**Verified: the loader's behavioral module floor is an equality pin across three sites.** `scripts/workflow-flight-recorder-registry.json` registers the module under `test_modules` as `{"path": "lib/test/modules/prompt-extension-reader.sh", "minimum_assertions": 102, …}`, `lib/test/run.sh` passes the same number positionally at its `devflow_run_full_suite_module` call site, and the module emits a matching tally — all three move in the same change when assertions are added. `claim=module-floor-equality-triple revision=94203da5698bf0e9c74ab73d9bc26c9d51993758 identity=1c5111185ae47c2cff1b9c4078395e4d9646ad9f`

**Verified: the mechanism's unstated in-repo dependencies, each probed.** (1) `git show <ref>:<path>` exits 128 with empty stdout for a path absent at that ref, and 0 for a present one — observed on a scratch repository. Exit 128 is git's *generic* fatal status, so it does not by itself distinguish absence from a corrupt object, from an unresolvable `FETCH_HEAD`, and from a dubious-ownership refusal; the absence arm is therefore gated on a positive `git cat-file -e "FETCH_HEAD:<path>"` refutation, the discriminator idiom this workflow already uses for `.claude/settings.json`. (2) `scripts/load-prompt-extension.sh` exits 0 with zero stdout bytes for an absent extension file, and likewise for a present-but-empty one — observed against the repository's own helper, which is what makes both the fail-closed arm and the truncation produce "no extension" rather than an error. (3) `scripts/devflow-cloud-writer-contract.json` carries a SHA-256 pin over `scripts/load-prompt-extension.sh`, so editing the loader turns the required `lib + python tests` check red until `python3 lib/test/cloud_writer_contract.py generate` refreshes it in the same commit.

**Verified: the `git show FETCH_HEAD:` reads resolve on this job's shallow checkout.** The review job checks out at `fetch-depth: 50` by default, and `baseprovision` does not depend on that depth: it runs its own `git fetch --depth=1 origin "$BASE_REF"` and reads `FETCH_HEAD` only inside that fetch's success branch. A fetch that fails leaves the branch unentered, which is the fail-closed arm this change reuses rather than adds.

**Verified: the displaced-path channel exists and today has exactly one producer.** `skills/review/phases/phase-0-setup.md` Phase 0.1 runs `git status --porcelain` and warns on output, and its issue-#504 rule attributes any path listed in the run's engine-ground-truth block to the trusted-source floor as *"expected displacement, NOT a PR defect …"*; Phase 0.1.5 persists that list to `.devflow/tmp/displaced-paths.txt`. The `compose` step binds `HARDENED_PATHS: ${{ steps.harden_hooks.outputs.displaced_paths }}` — a single step's output — and on that step's skip arm it publishes `printf 'displaced_paths=\n'`, an empty value. A second producer, and a binding that joins both, are therefore part of this change rather than an assumed capability.

**Unverified, and load-bearing for the fallback arm — flagged as an assumption.** `claude-code-action` removes sensitive paths and restores them from the base branch before installing plugins, which is why this repository vendors under `.devflow/vendor/` rather than `.claude/`; the repository states that set only by its `.claude/` member and says nothing about `.devflow/`. Whether the truncated workspace copies survive that pass is therefore **not established here** — assumption, confirm before implementing. It is load-bearing exactly once: on the arm where the variable does not reach the loader and the truncation is the only control. A restore from the PR head would undo the truncation there; a restore from the base branch would replace it with base bytes, which is a different guarantee from the one this issue states. The pinned action's own source, and a `matcher-probe.yml` observation, are the two evidence sources.

**Verified: a consumer checkout is not guaranteed to contain `.devflow/prompt-extensions/`.** `install.sh` creates `.claude-plugin` and `.github/workflows .github/actions` and creates no prompt-extensions directory; a redirect into a missing directory fails, and the default step shell's `-e` would abort the review job. Even in this repository the protected set has a member with no live file: `git ls-files .devflow/prompt-extensions/` lists `requesting-code-review.md.example` and no `requesting-code-review.md`.

**Verified: GitHub's workflow-file provenance differs between the two lanes this repository routes to** (relied-on third-party behavior, checked against official documentation — https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows). `pull_request_target` *"runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the `pull_request` event does"*; `pull_request` runs from the PR merge commit. The same page states that *"with the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository."* `.github/workflows/devflow-review.yml`'s inline event dedupe routes bot actors to `pull_request_target` and human actors to `pull_request`.

**Unverified, and load-bearing — flagged as an assumption, per environment hop, with the measurement made a follow-up.** Whether a step-level `env:` entry on the `Run Claude Code` step is visible to a command run through the agent's Bash tool is **not established here**, and the two protected loads sit at different depths: the `review` load runs in the orchestrator's shell (hop one), while the `requesting-code-review` load runs inside a dispatched `general-purpose` Task (hop two). Every existing `env:` entry on that step is consumed by the CLI process itself, and this repository's own comment beside `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS` records that even the CLI-level effect was measured by `matcher-probe.yml` rather than assumed. The design is written so that the failure direction is safe at both hops — an unpropagated variable makes the loader resolve the repo-root path, where it finds the truncated file and prints nothing — so a propagation failure costs the feature, never the boundary. The loader's resolved-root breadcrumb, surfaced at hop two through the amended `EXTENSION-STATUS:` field, is what makes such a failure observable rather than silent.

**Residual unprotected surface, stated explicitly.** This change moves the review tier's appended-prompt bytes behind the trusted boundary; it does not make the reviewer's whole input base-ref-derived. The residuals below are the ones this change is aware of, and the list is not asserted as exhaustive:

1. **The plugin tree itself remains PR-selectable, and that is a separate issue.** `BAKED_MARKETPLACES` carries the literal `./` — the PR-head checkout root — and `.claude-plugin/marketplace.json` is an ordinary tracked file whose `source` field a pull request can edit. Repointing it supplies the whole skill tree, the loader included, from PR-head bytes, which would moot both the trusted closure and the version pin. Whether `claude-code-action` wipes and restores `.claude-plugin/` the way it does `.claude/` is **not established here** — assumption, confirm before implementing. This issue closes the extension channel and does not claim to close that one; it is filed as its own issue.
2. **The composite actions under `.github/actions/` are PR-head-supplied on both lanes.** The review job resolves `uses: ./.github/actions/read-project-config`, `uses: ./.github/actions/vendor-plugin`, and `uses: ./.github/actions/setup-project-env` from the workspace, after the `ref: ${{ inputs.ref }}` checkout. Those files sit outside `.github/workflows/`, so GitHub's `workflows` push permission does not gate them, and they are not the marketplace manifest, so residual 1 does not cover them. A pull request that edits `vendor-plugin` supplies the whole plugin tree — the loader included — regardless of where the `devflow_version` input came from. This belongs to the same separately-filed issue as residual 1.
3. **The diff channel remains.** For any pull request that edits `.devflow/prompt-extensions/review.md` — the only pull requests on which this attack exists — the added bytes still reach the reviewer as part of the diff it is required to read (`Bash(gh pr diff:*)` is on the review profile). The change moves those bytes from the *appended-instructions* channel to the *reviewed-data* channel, where the engine-ground-truth block's existing untrusted-data framing is the compensating control. That is a real reduction, not an elimination.
4. On the human `pull_request` lane the workflow file itself comes from the PR merge commit, so a collaborator able to push `.github/workflows/` changes can alter the review job directly. The hardening raises that from an ordinary file edit to a workflow edit, which GitHub gates behind the `workflows` push permission and which is conspicuous in review. On the bot `pull_request_target` lane the workflow file is base-branch-derived, and there this change closes the boundary.
5. When `vendor-plugin` resolves `vendor_source` to a value other than `fetch` — `self` in DevFlow's own repository, `committed` in a consumer that committed the plugin tree — the skill tree itself is PR-head-supplied, so the hardening is defense in depth there. On the `fetch` branch it is a closed boundary **for the extension channel**, subject to residuals 1 and 2.
6. The protected set is the extensions the review tier loads. Every other file under `.devflow/prompt-extensions/` stays on the untouched PR-head path, and the drift guard named in the acceptance criteria is what keeps the protected set matched to the tier's actual loads.
7. **The base-ref-pinned-old-loader window.** A consumer whose base ref pins a `devflow_version` predating this change gets a loader that ignores the variable, so the truncation is the only control there and `Bash(git checkout:*)` can restore the workspace copy in session. With `devflow_version` read from the base ref, that window is a maintainer's pin rather than a pull request's choice, and it closes when they bump. During the same window a consumer whose base ref carries a committed extension loses it, which the User Impact section states rather than glosses.
8. `/devflow:review-and-fix` checks out the PR head in Step 0.5 and then runs the shared engine, so its Phase-3 `requesting-code-review` load reads PR-head bytes. That tier is human-invoked, is not the required merge gate, and its invoking actor already holds write access — so it is left out of scope here, knowingly, rather than covered.

**Documentation Drift.** The Step 1 pass returned `DRIFT FOUND` on both legs. Named sites:

- `docs/DEVFLOW_SYSTEM_OVERVIEW.md` — the security model's base-ref trust boundary enumerates `provision_env`, the `allowed_tools` list, and the `setup` block, and omits both the prompt extension and `devflow_version`; the prompt-extension section documents only the skill-name containment check as the load's security property.
- `docs/cloud-setup.md` — the read-only-profile paragraph justifies the grant with *"they never mutate the tree"* and the descriptor *"consumer-owned"*, neither of which speaks to authorship on a PR-head checkout; the **Trusted-ref rule** paragraph states the base-ref rule per-file for `.claude/settings.json` with no adjacent statement for `.devflow/prompt-extensions/`.
- `docs/cloud-allowlist.md` — the `load-prompt-extension.sh` grant-surfaces section frames the directory-agnostic wildcard's risk only as the extension failing to load; it is also where the pending propagation measurement is recorded.
- `docs/shadow-review.md` — its provenance check's purpose changes once a base-ref materialization exists on the review tier.
- `.devflow/config.schema.json` and `scripts/emit-git-env.sh` — both carry the issue-#295 repo-root-reader enumeration naming the loader, whose ambient-`GIT_DIR` failure is a silent miss; after this change the loader anchors on the repo root only on its fallback branch, and the two enumerations are a coupled pair.
- `scripts/load-prompt-extension.sh` header and the `devflow-runner.yml` read-only-safety comment — both assert the trust premise this change falsifies.
- `docs/install.md` — the two-channel upgrade note is where the base-ref-pinned-old-loader window belongs.

**Architecture Alignment.** The change extends two established patterns. The base-ref config read gains `devflow_version` alongside `provision_env`, the `allowed_tools` list, and the `setup` block, which makes the plugin version a maintainer input on the same footing as the tool grants — and carries the same consequence the repository already documents for trigger-time-resolved keys: a pull request that bumps `devflow_version` does not affect its own review. The trusted closure adds one more member to the `$RUNNER_TEMP` family whose existing members are enumerated in the trusted-closure inventory above, governed unchanged by that step's trust-rule comment: *"materialized from the base ref below, ONLY inside the successful base-ref fetch branch — FETCH_HEAD elsewhere (e.g. left by actions/checkout) can point at the PR HEAD."* It diverges in one stated respect, covered by an acceptance criterion below: the closure's creation, the workspace directory's creation, the truncation, and the variable export sit **outside** the fetch-success branch, because those are the actions that must hold when the branch is not entered. The environment-variable override follows the `DEVFLOW_GH` / `DEVFLOW_JQ` / `DEVFLOW_BASH` convention: honored at top precedence when set and non-empty, inert when unset, and inert when set to the empty string.

**Judgment-surface note.** This change introduces no new LLM judgment over third-party text, and no new semantic judgment over it. It narrows the input provenance of an existing one — the reviewing agent's own prompt — so the input-is-data guard class does not attach here; the hostile-input case named in the Testing Strategy is carried anyway, as the direct positive control for the boundary this change establishes.

**Dependencies.** GitHub Actions (`$RUNNER_TEMP`, step outputs, step-level `env:`) and `git show`. Issues #402, #404, #458, #504, and #505 are merged precedents this builds on, not open prerequisites.

**Data/Schema Considerations.** No configuration key is added. `devflow_version` changes the *channel* it is read from, which moves it into the repository's documented trigger-time-resolved class: a change to it inside a pull request is inert for that pull request's own review and takes effect after merge.

**Cross-layer Impact.** Workflow layer (`.github/workflows/devflow-runner.yml`, `.github/workflows/matcher-probe.yml`), helper layer (`scripts/load-prompt-extension.sh`, `scripts/render-grounding-block.sh`, and the extracted materialization helper), prompt layer (`skills/review/SKILL.md`, `skills/review/phases/phase-0-setup.md`, `skills/review/phases/phase-3-agents.md`), generated-artifact layer (`scripts/devflow-cloud-writer-contract.json`, `lib/test/cloud_writer_deps.py`), test layer (`lib/test/run.sh`, `lib/test/modules/prompt-extension-reader.sh`, its inventory record, `lib/test/modules/coverage-map.json`), install layer (`install.sh`), and docs.

## Acceptance Criteria

- [ ] `.github/workflows/devflow-runner.yml` resolves the `devflow_version` **input** it passes to `vendor-plugin` from the trusted base ref rather than from the PR-head checkout. The criterion is scoped to that input: the composite action that consumes it is itself PR-head-supplied, which residual 2 records and this issue does not close.
- [ ] A pull request that changes `.devflow/config.json`'s `devflow_version` does not change the plugin version used by its own review run.
- [ ] The step that produces the base-ref `devflow_version` is declared **before** the `vendor` step, and the value the `vendor` step receives is non-empty whenever the base ref carries one.
- [ ] The review job completes normally on the `vendor-plugin` **fetch** branch — the thin-install shape this repository's own `self` branch never exercises.
- [ ] `scripts/load-prompt-extension.sh` resolves its extension directory from `DEVFLOW_PROMPT_EXTENSION_ROOT` when that variable is set and non-empty, composing `${DEVFLOW_PROMPT_EXTENSION_ROOT}/${SKILL_NAME}.md`, and resolves from the existing repo-root path when that variable is unset, and likewise when it is set to the empty string.
- [ ] The existing `SKILL_NAME` guard rejecting a value that contains `/`, and rejecting a value that contains `..`, with exit 2 runs before any filesystem access on both resolution paths.
- [ ] `scripts/load-prompt-extension.sh` writes a stderr breadcrumb naming the directory it resolved **only when `DEVFLOW_PROMPT_EXTENSION_ROOT` is set**, so the observed output of every caller that runs without it — every `skills/*/SKILL.md` load site outside the review tier, and every local and interactive run — is byte-unchanged.
- [ ] The loader's stdout bytes and its exit code are unchanged with `DEVFLOW_PROMPT_EXTENSION_ROOT` unset, across every input shape the existing `prompt-extension-reader` module exercises.
- [ ] Every extension-classification site that runs inside the review job — `skills/review/SKILL.md`, `skills/review/phases/phase-3-agents.md`'s `EXTENSION-STATUS:` contract, and the vendored `skills/requesting-code-review/SKILL.md` the Phase-3 Task invokes, which inherits the job-scoped variable like every other command in the job — classifies `loaded-with-content` and `loaded-empty` on the command's **stdout** explicitly, and an empty extension under the new stderr breadcrumb still maps to `loaded-empty`. Both edits are prompt-surface edits and are routed through this repository's `superpowers:writing-skills` discipline.
- [ ] `skills/review/phases/phase-3-agents.md`'s return contract carries the dispatched reviewer's resolved extension-directory root as a required field.
- [ ] Every `load-prompt-extension.sh` invocation under `skills/` is textually unchanged — those in `skills/*/SKILL.md`, which the `LPE_EXPECT_LINE` standardized-step scan and the issue-#275 portable-anchor pins cover, and the vendored-literal command in `skills/review/phases/phase-3-agents.md`, which no scan covers and which is therefore checked by reading it.
- [ ] `.github/workflows/devflow-runner.yml` creates `$RUNNER_TEMP/devflow-trusted-prompt-ext/`, creates `.devflow/prompt-extensions/` in the workspace, truncates to empty the workspace copy of each extension in the protected set — creating an empty file for a protected name the checkout carries none of — and exports `DEVFLOW_PROMPT_EXTENSION_ROOT` pointing at the closure, on every review run, with all four actions placed outside the base-ref fetch-success branch and outside any conditional on a helper resolving.
- [ ] The review job completes normally on a checkout that contains no `.devflow/prompt-extensions/` directory and on one that contains the directory with none of the protected files.
- [ ] `.github/workflows/devflow-runner.yml` populates the closure from the base ref — the skill names `review` and `requesting-code-review`, passed to the helper as explicit arguments — inside the existing base-ref fetch-success branch and nowhere else.
- [ ] Each materialized file is byte-identical to the base ref's copy, including a file carrying no trailing newline and a file carrying a UTF-8 BOM, verified with `cmp`.
- [ ] A protected name the base ref does not carry contributes zero bytes to the closure, no warning, and no extension text — the ordinary shape for a consumer that committed none. A shell redirect creates its target before the read runs, so the helper removes a target left zero-length by a failed read rather than leaving it behind.
- [ ] When the base ref was read and carries the file but the closure was not populated, the job emits a `::warning::` whose text names which failure applied — from the closed set {unwritable `$RUNNER_TEMP`, a read that failed for a reason other than the object being absent} — and the reviewing agent receives no extension text.
- [ ] When the base ref's content was never established — from the closed set {base-ref fetch failure, empty `BASE_REF`, no trusted source for the materialization helper} — the job emits a distinct notice stating that the trusted materialization was not attempted, never a reason-naming warning that would assert whether an extension exists, and the reviewing agent receives no extension text.
- [ ] An extension file present on the PR head and absent on the base ref contributes zero bytes to the reviewing agent's prompt.
- [ ] An extension file present on the base ref and absent on the PR head is materialized and reaches the reviewing agent's prompt.
- [ ] The branch-selecting logic and the `::warning::` composition live in a `scripts/*.sh` helper the suite drives directly, per this repository's rule that branch-selecting and message-composing inline workflow shell is extracted into a suite-drivable helper; the workflow invokes that helper only from a trusted source, by the same rank ladder `filter-runner-tools.sh` uses, and when no trusted source resolves the closure is left empty and the warning names that case.
- [ ] A suite assertion derives the set of skill names reachable as `load-prompt-extension.sh <name>` invocations from `skills/review/` and fails when it differs from the protected set the workflow declares.
- [ ] The truncated extension paths reach the engine-ground-truth block's displaced-paths section through a named producer of their own, and the `compose` step's `HARDENED_PATHS` binding joins that producer with `harden_hooks`' `displaced_paths` — including on the arm where `harden_hooks` skips and publishes an empty value — and `skills/review/phases/phase-0-setup.md`'s Phase 0.1 attribution enumeration is extended from {content delta, mode-only delta} to include the **untracked** delta a newly created empty extension file produces, so Phase 0.1 attributes their `git status --porcelain` output to the trusted-source floor rather than warning about them.
- [ ] `.github/workflows/matcher-probe.yml` carries a job that measures whether a step-level `env:` entry on a `claude-code-action` step is visible both to a command run through the orchestrator's Bash tool and to one run inside a dispatched Task, and `docs/cloud-allowlist.md` records that job and names the measurement as pending a maintainer-dispatched run — the implementing run adds the job and the documentation entry, and does not dispatch the probe.
- [ ] `scripts/devflow-cloud-writer-contract.json` carries the post-change SHA-256 of `scripts/load-prompt-extension.sh`, discharged through the granted suite command alone: `lib/test/run.sh` is RED while the pin is stale and green once it is refreshed. The regeneration command itself (`lib/test/cloud_writer_contract.py generate`) is an interpreter-head invocation that `devflow_implement.allowed_tools` does not grant, and a grant this pull request ships would be inert for its own run, so the criterion names the granted signal rather than the ungranted command.
- [ ] `lib/review-profile.tokens` is unchanged by this change, discharged through the granted suite command alone: `lib/test/run.sh` drives `lib/generate-capability-profiles.py --check`, so a manifest-to-literal drift and a widening of the resolved review token list each turn the suite red.
- [ ] `scripts/load-prompt-extension.sh`'s header sentence asserting that the file's bytes are trusted, and `.github/workflows/devflow-runner.yml`'s read-only-safety comment justifying the grant on tree non-mutation, both state the review tier's trusted-source rule after the change; and the two coupled path-containment sentences that name `.devflow/prompt-extensions/` as the un-escapable root — the loader header's and `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s — are restated as containment within the selected root.
- [ ] `docs/DEVFLOW_SYSTEM_OVERVIEW.md`'s base-ref trust boundary enumeration names the prompt extension and `devflow_version`, `docs/cloud-setup.md`'s Trusted-ref rule carries an adjacent statement for `.devflow/prompt-extensions/`, and the issue-#295 repo-root-reader enumeration is updated in both `.devflow/config.schema.json` and `scripts/emit-git-env.sh` to record that the loader anchors on the repo root only on its fallback branch.
- [ ] `lib/test/run.sh` reports zero failures, and the `prompt-extension-reader` module's `minimum_assertions` value is updated in `scripts/workflow-flight-recorder-registry.json`, at the `lib/test/run.sh` call site, in the module's `lib/test/modules/prompt-extension-reader.inventory.md` record, and in the emitted tally in the same change.

## Implementation Notes

**Approach**

The mechanism lands in these files.

1. `.github/workflows/devflow-runner.yml` moves the `devflow_version` resolution onto the trusted base ref, and the step ordering is the load-bearing part. `baseprovision` is declared **after** `vendor`, so it cannot supply the value: a `steps.baseprovision.outputs.*` reference inside the `vendor` step resolves to the empty string, and `vendor-slice.sh` refuses an empty `DEVFLOW_REF` with *"refusing to track mutable main"* — killing the review job on every `fetch`-branch consumer while this repository's own `self` branch, which ignores `DEVFLOW_REF`, stays green. The base-ref read is therefore a step of its own placed above `vendor`: it fetches the base ref and reads `devflow_version` from it, and the `vendor-plugin` step consumes that output. Because it performs the base-ref fetch earlier than today, the change states what `FETCH_HEAD` holds for every later reader of it, and `baseprovision`'s own fetch-success branch keeps its existing semantics. The fail-closed direction is the one the job already uses: when the base-ref read does not resolve, the existing degraded path governs.

2. `scripts/load-prompt-extension.sh` gains one resolution branch. Where it currently composes `ext_file` from `_devflow_root`, it first tests `DEVFLOW_PROMPT_EXTENSION_ROOT` for a non-empty value and, when present, composes `ext_file` from it directly, skipping the `git rev-parse --show-toplevel` call entirely. The `*/*` and `*..*` name guard keeps its current position ahead of all filesystem access, so containment holds relative to whichever directory was selected. The trusted-root branch writes the resolved-root breadcrumb to stderr; the repo-root branch does not, so no caller that leaves the variable unset sees any change in observed output. No call site changes, and `scripts/devflow-cloud-writer-contract.json` is regenerated in the same commit because it pins this file's bytes.

3. `skills/review/SKILL.md`, `skills/review/phases/phase-3-agents.md`, and the vendored `skills/requesting-code-review/SKILL.md` — every classification site reached inside the review job, since `$GITHUB_ENV` is job-scoped and the Phase-3 Task's commands inherit it — are amended in the same change so the stderr breadcrumb cannot perturb the classification it feeds: each restates its `prints text` / `prints nothing` test on **stdout** explicitly, and the dispatched reviewer's resolved extension-directory root becomes a required `EXTENSION-STATUS:` field so hop-two propagation is observable. Both are prompt-surface edits, routed through this repository's `superpowers:writing-skills` discipline. The sibling `skills/*/SKILL.md` load sites outside the review tier need no edit, because the variable is published only in the review job and so none of them ever observes the breadcrumb.

4. A new `scripts/materialize-trusted-prompt-extensions.sh` owns the workflow's branch selection and message composition, so `lib/test/run.sh` drives each arm. It takes the protected skill names (`review` and `requesting-code-review`) and the target directory as explicit arguments, performs the `git show FETCH_HEAD:.devflow/prompt-extensions/<name>.md > "<target>/<name>.md"` reads as **direct redirects** and removes a target the redirect created but the read never filled, distinguishes "the base ref carries no such file" — established by `git cat-file -e` reporting the object absent, never inferred from exit 128 alone — (silent, no warning) from the five-member failure set, and prints the reason-naming warning on stderr for the latter. The direct redirect is a deliberate divergence from the `baseprovision` step's `_floor_raw` pattern: `VAR=$(git show …)` strips every trailing newline and `printf '%s\n' "$VAR"` re-adds exactly one, which would make the byte-identity criterion above unsatisfiable. It uses bash builtins for every value that decides an emitted result, since `tr`, `sed`, `wc`, `cut`, and `head` are not preflight-guaranteed and a missing one fails open.

5. The workflow splits the rest into an unconditional part and a conditional part. The unconditional part — `mkdir -p "$RUNNER_TEMP/devflow-trusted-prompt-ext"`, `mkdir -p .devflow/prompt-extensions`, the truncating `: > .devflow/prompt-extensions/<name>.md` over each protected name, a step output listing the paths it touched, and the `DEVFLOW_PROMPT_EXTENSION_ROOT` assignment appended to `$GITHUB_ENV` — selects no branch and composes no user-facing message, so it stays inline; it runs on a failed base-ref fetch exactly as on a successful one, and on an unresolvable helper exactly as on a resolved one. The workspace `mkdir -p` is what keeps the redirect from aborting the job under the default `bash -e` step shell on a consumer that never created the directory. The conditional part calls the helper from inside `baseprovision`'s fetch-success branch, resolving the helper itself through the same trusted-source rank ladder `filter-runner-tools.sh` uses, and emits the no-trusted-source warning inline when neither rank resolves, mirroring the `tools` step's existing inline fail-closed arm. Because the closure exists and the variable points at it regardless, an unresolvable helper leaves the closure empty rather than sending the loader back to the repo root. The variable arrives through `$GITHUB_ENV`, so no command line carries a leading `VAR=value` prefix, which the cloud matcher denies as a shape.

6. The truncation step's own path-list output joins `harden_hooks`' `displaced_paths` at the `compose` step, whose `HARDENED_PATHS` binding becomes a join of the two producers rather than a single reference — so the paths reach the grounding block even on the arm where `harden_hooks` skips and publishes an empty value. `scripts/render-grounding-block.sh` renders them unchanged, Phase 0.1 attributes their working-tree delta to the trusted-source floor, and Phase 0.1.5 persists them to `.devflow/tmp/displaced-paths.txt` for the later phases that route HEAD verification through `git show`.

**Code Patterns**

- The `baseprovision` step's existing base-ref config read, which already supplies `provision_env`, the `allowed_tools` list, and the `setup` block — `devflow_version` joins it rather than gaining a new mechanism.
- That step's `FLOOR_HELPER=''` / `FLOOR_SOURCE='absent'` pre-initialization ahead of the branch, so the fetch-failed and empty-`BASE_REF` paths leave the values set under `set -u`.
- The two-rank trusted-source consuming ladder with the `::warning::` fail-closed arm in the `tools` step.
- The `harden_hooks` step's `displaced_paths` output feeding `HARDENED_PATHS` into `render-grounding-block.sh` — the shape the truncated paths reuse, extended to two producers.
- `scripts/describe-denial-count.sh` is the reference for extracting a branch-selecting, message-composing chunk of workflow shell into a suite-drivable helper.
- `lib/resolve-gh.sh`'s `DEVFLOW_GH` handling is the reference for a non-empty environment override taken at top precedence and inert when empty.

**Testing Strategy**

The boundary is executable at these levels, and the plan names each: the loader's CLI contract (stdout bytes, exit code, stderr) as a unit boundary; the new helper's per-arm behavior (populated files, warning text, filesystem effects) as a unit boundary; the workflow's unconditional part and its two-producer `HARDENED_PATHS` join as an integration boundary driven the way `lib/test/run.sh`'s `emit_tools` driver already drives the deny floor; and the environment-propagation question as a live cloud measurement in `matcher-probe.yml`, whose result is evidence recorded by a maintainer-dispatched run rather than a gate the implementing run ticks.

*Loader assertions*, added to `lib/test/modules/prompt-extension-reader.sh` (its `minimum_assertions` floor moves in the same change, across `scripts/workflow-flight-recorder-registry.json`, the `lib/test/run.sh` call site, the module's inventory record, and the emitted tally). The input-shape matrix for the new variable is enumerated as a closed set — exactly these eight, complete by construction, being the product of the variable's presence states and the target's filesystem states that the resolution branch distinguishes:

1. unset — resolves from the repo root, stdout byte-identical to today.
2. set to the empty string — resolves from the repo root, stdout byte-identical to today.
3. set to whitespace only — treated as a directory name, the file is absent, exit 0 with empty stdout.
4. set to a non-existent directory — file absent, exit 0 with empty stdout.
5. set to a path that is a regular file rather than a directory — file absent, exit 0 with empty stdout.
6. set to an existing directory holding no `<skill>.md` — exit 0 with empty stdout.
7. set to an existing directory holding `<skill>.md` — that file's bytes verbatim, compared with `cmp` for byte exactness including UTF-8 content, and the repo-root copy demonstrably not read (the two fixtures hold different bytes).
8. set to an existing directory, with a `SKILL_NAME` containing `/` and one containing `..` — exit 2, empty stdout, guard fires ahead of any read.

Each of these fails first for the right reason: the resolution branch does not exist yet, so every case that sets the variable reads the repo-root copy and returns the wrong bytes. Two further assertions cover the breadcrumb: each branch writes a stderr line naming the directory it resolved (failing first because no breadcrumb exists), and an empty extension under that breadcrumb still yields empty **stdout**, which is the operand the amended `EXTENSION-STATUS:` contract classifies on.

*Helper assertions*, added to `lib/test/run.sh` driving `scripts/materialize-trusted-prompt-extensions.sh` over a temporary git fixture, mirroring the `emit_tools` driver shape. Every arm below starts from a fixture whose PR-head workspace copies are **present and non-empty**, so the claimed suppression is actually exercised rather than asserted over an already-absent input.

- Base ref carries every protected file — each materialized byte-for-byte, no warning.
- Base ref carries none of them, with `git cat-file -e` confirming the objects absent — the closure holds no file for those names (the redirect's zero-length target is removed), **no warning**, exit 0: the ordinary extension-less consumer.
- A read that fails for a reason other than absence (a `FETCH_HEAD` that does not resolve, with `git cat-file -e` not confirming absence) — nothing materialized, **warning** naming the read-failure reason, which is the arm exit 128 alone would have routed to silence.
- Base ref carries a strict subset — that subset materialized, no warning for the absent names.
- Target directory unwritable — nothing materialized, warning naming the unwritable-temp reason.
- Empty `BASE_REF` — the helper reads no `FETCH_HEAD` path and emits the not-attempted notice, never a reason-naming warning about a file it never looked for.
- A base-ref file with no trailing newline, and one carrying a UTF-8 BOM — each round-trips byte-exactly through `cmp`, which is the assertion the direct-redirect shape exists to satisfy.
- Warning-text arm ordering is asserted per arm, so a reordered branch misattributing the reason turns the suite red.

*Workflow-wiring assertions*, driving the unconditional part the way `emit_tools` drives the tools step:

- The closure directory is created, the workspace directory is created, the workspace copies are truncated, and the variable is exported on the fetch-failure arm, on the empty-`BASE_REF` arm, and on the no-trusted-helper arm — the three arms where the conditional part does not run. Each fails first because today's workflow does none of it.
- The unconditional part completes with exit 0 on a fixture with no `.devflow/prompt-extensions/` directory and on one where a protected name has no file — the two shapes that abort the job today under `bash -e`.
- On every failure arm the exported variable names the closure directory rather than the empty string, so the loader's repo-root fallback is not reachable from the workflow.
- The truncated paths appear in the value bound to `HARDENED_PATHS` on the arm where `harden_hooks` publishes an empty `displaced_paths`, which is the join the single-producer binding cannot express today.
- The `devflow_version` handed to `vendor-plugin` is the base-ref value on a fixture whose PR-head config names a different one — failing first because today's `extract` step reads the PR-head checkout.

*Adversarial and security assertions*, which are the point of the change:

- **The named attack path**, mirroring the existing `#404 attack: PR-head vendored helper NOT executed` assertion: a fixture whose PR-head `review.md` contains instruction-shaped text directing the reviewer to approve unconditionally, and whose base-ref `review.md` contains different, benign text. Assert the loader, run under the workflow's resolved environment, emits the benign bytes and that the instruction-shaped bytes appear nowhere in its stdout. This fails first because today's loader returns the PR-head bytes — the exact wrong behavior, not a missing feature.
- **The version-selection arm**: a fixture whose PR-head config pins a `devflow_version` different from the base ref's, asserting the base-ref value is the one used. This is the assertion that makes the loader-version choice a maintainer input.
- **The old-loader arm**, which is the guarantee-class case: run a loader that ignores the variable against a workspace whose copy the workflow truncated, and assert it emits nothing. This proves the mechanism holds on the path where the cooperating half is absent, which is the entire reason the truncation exists.
- **The unpropagated-variable arm**: run the post-change loader with the variable absent against that same truncated workspace, and assert it emits nothing and writes the repo-root breadcrumb — the observable that turns a silent feature loss into a diagnosable one.
- **The drift guard's positive control**: plant a `load-prompt-extension.sh <name>` invocation under `skills/review/` for a name outside the declared protected set and assert the new suite assertion fails on it.

*Boundary and state assertions*: the helper is idempotent across two consecutive runs on the same fixture; the unconditional part is idempotent over an already-truncated workspace and over an already-created closure.

`governing conventions consulted: CLAUDE.md, CONTRIBUTING.md, docs/, .devflow/prompt-extensions/` — `CLAUDE.md`'s six-shape adversarial matrix governs config-JSON consumers; the surface here is a single environment-variable path and its filesystem target, so the matrix enumerated above is that input type's analogue rather than the JSON six-shape set, and the narrowing is stated here rather than left silent.

**Documentation Needed**

- `docs/DEVFLOW_SYSTEM_OVERVIEW.md`
- `docs/cloud-setup.md`
- `docs/cloud-allowlist.md`
- `docs/shadow-review.md`
- `docs/install.md`
- `CLAUDE.md`
- `CONTRIBUTING.md`
- `lib/intervention-surfaces.md`

**Potential Gotchas**

- The suppression must not depend on the materialization. Placing the truncation inside the fetch-success branch reintroduces the exact defect this issue closes, and placing the variable export there does the same: on a fetch failure the workspace copy survives and the loader reads it.
- The redirect that truncates needs its directory to exist. `install.sh` creates no `.devflow/prompt-extensions/`, and the default step shell's `-e` turns a failed redirect into a dead review job — `mkdir -p` first.
- "The base ref carries no extension" is the common case, not a failure. Warning on it would put a `::warning::` on every review run in every consumer that never opted in.
- The materialization belongs strictly inside the base-ref fetch-success branch. `FETCH_HEAD` outside that branch can point at the PR head, which is the failure the existing trust-rule comment was written to prevent.
- Command substitution strips trailing newlines and drops NUL bytes, so the `_floor_raw` shape cannot carry the byte-identity criterion. Use a direct redirect for this closure.
- A stderr breadcrumb is observed output on a runner that merges stderr into tool results, and the standardized load prose in every `skills/*/SKILL.md` classifies on "prints text". Scoping the breadcrumb to the variable-set case keeps every site that runs without it byte-unchanged; the two that do set it move their classification to stdout in the same change, because leaving it on "printed text" would let an empty load report as `loaded-with-content`.
- `HARDENED_PATHS` binds to one step's output today, and `harden_hooks` publishes it empty on its skip arm. A second producer needs its own output and a join, not an assumed append.
- `$RUNNER_TEMP` does not exist off Actions, so the variable stays unset on the local and interactive tiers and those runs keep today's behavior.
- The helper must be executed from a trusted source. A helper resolved from the PR-head checkout would be PR-editable, which is the issue-#404 rejection applied one level up — and because the helper only populates, an unresolvable helper degrades to an empty closure rather than to PR-head bytes.
- No new tool grant is required: the existing `Bash(*/load-prompt-extension.sh:*)` token already admits trailing arguments, the variable arrives through the step's `env:` rather than a command prefix, and the new helper runs as a workflow step rather than an agent command. A change that nonetheless widens the resolved review token list fails `lib/generate-capability-profiles.py --check` closed until `lib/review-profile.tokens` is updated in the same pull request.
- Moving `devflow_version` to the base ref makes it trigger-time-resolved: the pull request that bumps it will not see the bump in its own review, and that is the intended consequence, not a defect to work around.
- Editing `scripts/load-prompt-extension.sh` breaks the SHA-256 pin in `scripts/devflow-cloud-writer-contract.json`, and `lib/test/run.sh` stays RED until it is refreshed in the same commit. The regenerating command is not granted on the implement tier, so a cloud run reaches this as a red suite rather than as a command it can issue — refresh it from a tier that grants the interpreter head.
- Adding assertions to `lib/test/modules/prompt-extension-reader.sh` ships CI red unless the `minimum_assertions` value moves in `scripts/workflow-flight-recorder-registry.json`, at the `lib/test/run.sh` call site, in the module's inventory record, and in the emitted tally together.

The01Geek:
## Summary
- The automated cloud reviewer read its own appended prompt extensions from the **pull request's head checkout**, so a PR author could write the instructions used to review their own PR. The review job now materializes them from the **trusted base ref** into a `$RUNNER_TEMP` closure, and truncates the workspace copies unconditionally so an older reader finds nothing there.
- The `devflow_version` input passed to `vendor-plugin` is likewise resolved from the base ref, so a PR can no longer choose which plugin commit — and therefore which loader — reviews it.
- Suppression is **unconditional**; population is conditional. Every arm on which the closure is not populated degrades to an *empty* closure rather than back to the PR-head file.

## Changes

**`devflow-runner.yml` (review job)**: New unconditional `promptext` step creates the closure, creates `.devflow/prompt-extensions/`, truncates each protected extension, and exports `DEVFLOW_PROMPT_EXTENSION_ROOT`; a new `baseversion` step resolves `devflow_version` from the base ref above the `vendor` call; population happens inside the existing fetch-success branch only, through a trusted-source rank ladder; a new `displaced_join` step joins the truncated paths with `harden_hooks`' displaced paths.

**`scripts/materialize-trusted-prompt-extensions.sh` (new)**: Owns the branch selection and `::warning::`/`::notice::` composition. Establishes absence *positively* (`git cat-file -e`) before licensing silence, rejects an empty trusted read, removes a target left zero-length by a failed read, and writes by direct redirect so materialized bytes are byte-identical (BOM and no-trailing-newline included).

**`scripts/load-prompt-extension.sh`**: Adds the `DEVFLOW_PROMPT_EXTENSION_ROOT` branch — top precedence, inert when unset *and* when empty, name guard before any filesystem access on both branches. With the variable unset, stdout and exit code are byte-unchanged.

**Prompt surfaces**: The three extension-classification sites classify on **stdout** explicitly; `phase-3-agents.md` requires the dispatched reviewer's `resolved-root` and now **consumes** it, so a hop-two propagation failure is recorded rather than read as a successful load; Phase 0.1 attributes displaced paths by directory prefix (git collapses a wholly-untracked directory).

**`matcher-probe.yml` + `scripts/env-propagation-probe-verdict.py` (new)**: An un-run probe measuring whether a step-level `env:` entry reaches both an orchestrator command and a dispatched Task. A hop counts as *reported* only on an observed value, so a run that measured nothing yields `INCONCLUSIVE` instead of a false negative.

**Tests**: 9 helper arms over real git fixtures with `cmp` byte checks and a named-attack arm; the `promptext` and `displaced_join` steps are *extracted and executed*, not grepped; a closed presence × target-state matrix for the loader; a drift guard deriving the protected set from `skills/review/` with a non-vacuous positive control.

## Resolves
Resolves #874

## Test Plan
- [x] `lib/test/run.sh` — 13052 passed, 0 failed, 0 skipped
- [x] `shellcheck --severity=warning -e SC1091` over every tracked non-`lib/test` `.sh`
- [x] `shellcheck --extended-analysis=false lib/test/run.sh`
- [x] `ruff check` over every tracked `.py`
- [x] `python3 lib/test/regenerate-artifacts.py` reconciles clean
- [x] A PR committing a plain file at `.devflow/prompt-extensions` no longer aborts the review job (reproduced, then fixed and covered)
- [x] Base-present/head-absent extension is materialized **and** reaches the loader (end-to-end arm)

## Post-Merge Verification
- [ ] The review job completes normally on the `vendor-plugin` **fetch** branch — the thin-install shape this repository's own `self` branch never exercises.

## Visual Changes
N/A

## Breaking Changes
None for behavior, but **one upgrade-ordering note for consumers**: the two halves arrive through different channels — the workflow via `install.sh`, the reader that honors `DEVFLOW_PROMPT_EXTENSION_ROOT` via the `devflow_version` vendor fetch. Re-running `install.sh` without advancing `devflow_version` truncates the workspace copies while the pinned reader still reads the repo-root path, so a committed `review.md` stops loading until the pin advances. Take both halves together.

**Scope**: this boundary covers the **automated** review runner (`devflow-runner.yml`) only. The manual `/devflow:review` comment path in `devflow.yml` also checks out the PR head and is **not** covered — a disclosed residual, recorded in `docs/DEVFLOW_SYSTEM_OVERVIEW.md`.

Generated via `/devflow:implement 874`


The01Geek:
Closed the three Important findings in `ed8e4916`. No production behavior changed — one agent-prompt correctness fix and two test-coverage additions.

**Important 1 — fixed, and it was consequential.** Confirmed the chain: `compute_verdict` has `if not (before and after): return "INCONCLUSIVE"`, so an agent that honors "the **five** … and nothing else" literally stops after Action 5, skips Action 6 (the AFTER control), and forces INCONCLUSIVE — the probe would run, cost a dispatch, and measure nothing.

Fixed **count-free** rather than `five` → `six`. Two coupled numerals in one prompt is the rot this repo's own convention tells us to avoid, and the next action added would re-break it.

**Important 2 — fixed.** All five rc-2 usage arms are now driven. Each pins **its own** diagnostic rather than a shared prefix: a reordered guard would otherwise refuse for the wrong reason with every arm still exiting 2, which is precisely the arm-ordering gap `describe-denial-count.sh` exists to close. Each arm also asserts nothing was materialized, since an arm that exited 2 *after* writing would satisfy an rc-only check.

**Important 3 — fixed.** Both degraded arms now driven:
- the unappendable-summary arm points `GITHUB_STEP_SUMMARY` at a **directory**, not a `chmod`-ed file, so `open(dir, "a")` raises `IsADirectoryError` for every uid — this arm cannot self-skip on a host that ignores mode bits, unlike the mode-bit arms elsewhere in the suite;
- it carries a **positive control** proving a writable summary really is appended, so it cannot pass against a helper that never writes one;
- the nesting arm derives depth from the interpreter's own recursion limit and emits **raw text**, because building the fixture through `json.loads` would raise in the generator.

I checked whether the `RecursionError` guard was even reachable — `parse_execution_file` catches bare `Exception`, and `RecursionError` is one, so a deep file could plausibly be swallowed as "unparseable" and never reach `collect()`. Empirically it does reach the guard: the verdict reads `nested too deeply to walk`. The guard is live, not dead code.

### Verification (control green in each case)

| Mutation | Result |
|---|---|
| `except OSError` → `except SystemExit` | summary arm crashed **rc 1** |
| `except RecursionError` → `except SystemExit` | nesting arm crashed **rc 1** |

- **`lib/test/run.sh` → 13080 passed, 0 failed, 0 skipped**
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` clean; `mutation-routing-worktree` gate clean
- `actionlint` — the one SC2129 style note is pre-existing on an unrelated job

Two things worth recording for whoever touches this next. My first pass collided with the existing `a10base`/`a10ws`/`a10closure` fixture names and broke the pre-existing asymmetry assertion — renamed to a `usag*` prefix. And `grep -F "--base-ref requires a value"` parses the pattern as an **option**; three arms needed `grep -qF --`.

No changeset added: these changes are `lib/test/` plus the repo-internal `matcher-probe.yml`, both covered by this PR's existing changeset.

PR #897, opened 2026-07-28
Reconcile the #881 memo coupled pair and cover the untested memo paths

The01Geek:
## Summary

Follow-up to **#886** (issue #881), addressing review findings that landed after that PR merged. Four prose defects — three of them false against HEAD — and five untested paths that #886's own memoization introduced.

Every test added here was verified RED under a mutation of the code it guards; the mutations and their verdicts are in the Test Plan.

## Changes

**`docs/DEVFLOW_SYSTEM_OVERVIEW.md`** — two false claims in the #881 paragraph:

- It said the two-limb memo-safety argument was *"mirrored at the shard-dispatch site"* in `harness-python-guards.sh`. That file carried only the key-purity limb; the argument-less `_load_mutation_census_module` and its module-statelessness limb appeared nowhere in it. The declared coupled pair was half-stale on the commit that declared it, which is the failure mode the coupled-invariant discipline exists to prevent. Fixed on the shell side, so the sentence is now true.
- It said a re-presented source *"is parsed once rather than once per presentation"*. With `_IMAGE_PARSE_CACHE_SIZE = 2`, presenting a third distinct image evicts the first, so re-presenting it re-parses — confirmed by running the sequence. Now scoped to each memo's bound, noting an evicted entry is recomputed and never answered wrongly.

**`lib/test/modules/harness-python-guards.sh`** — carries both limbs, so the mirror the overview claims actually exists.

**Both mirrors + the test docstring** — described the per-source memos as *"keyed on the source name and text alone"*. The two linter memos key on the **text alone**. Corrected in place; the load-bearing "no `repo_root`, no filesystem state" claim was already true and is unchanged.

**`lib/test/mutation-pin-census.py`** — the bound comment claimed headroom *"with margin"* and *"a margin that keeps them as the tree grows"*. The swept population sits a few files under the bound, and this repo adds one every time a durable test module is extracted. Restated honestly, without transcribing a count that would rot.

**`lib/test/test_pin_corpus_lint.py`** — five test gaps:

- `test_census_reuses_every_audited_source_within_one_build` cleared only `_logical_lines` while `_definition_scan` and `_extract_rows` — its **callers**, memoized on the same `(name, text)` pairs — stayed warm. A co-resident census made the hit count read `0`, and the failure message then sent the reader to raise a bound that was not the cause. It now **warms every memo first, then clears all three**, which makes the requirement self-enforcing: reducing the clear turns *this* test RED instead of arming a failure for whichever test later shares its process. The message names both candidate causes.
- The two outer census memos had no reuse detector at all — their reuse is *across* builds, which the within-build pin structurally cannot see.
- The linter's two-image memos had no hit-count guard, so a lowered bound would silently stop hitting.
- `_load_mutation_census_module`'s failed-exec cleanup (the `sys.modules.pop` before the re-raise, plus the retry `lru_cache` permits by not memoizing the raise) was untested.
- The `bisect` line-number derivation had no equivalence assertion. It is now driven through `_function_definitions` against an independent counting oracle — deliberately not by re-running `bisect` in the test, which would restate the implementation rather than test it.

`MemoizedParseContractTests`' docstring dropped its "two contracts" ordinal, which had already rotted past the properties the class holds.

## Test Plan

- [x] Each new test verified RED under a mutation of the code it guards, with the unmutated control green:

  | Mutation | Test | Verdict |
  |---|---|---|
  | reduce the three-memo clear to `_logical_lines` | within-build reuse pin | FAILED |
  | `_SOURCE_PARSE_CACHE_SIZE` 48 → 4 | across-build outer-memo reuse | FAILED (2) |
  | `_IMAGE_PARSE_CACHE_SIZE` 2 → 0 | linter image-memo reuse | FAILED |
  | drop `sys.modules.pop` on failed exec | census load-failure cleanup | FAILED |
  | start-line `+ 1` → `+ 2` | line-number counting oracle | FAILED (4) |
  | end-line `+ 1` → `+ 2` | line-number counting oracle | FAILED (4) |
  | `_newline_offsets` returns empty | newline-index pin | FAILED (3) |

- [x] `lib/test/run-module.sh harness-python-guards` → **38 passed, 0 failed**; `test_pin_corpus_lint.py: executed 130 test(s)` (was 125).
- [x] The coupled triple (registry `minimum_assertions`, the `run.sh` boundary operand, the emitted tally) is unchanged at **38** — the added assertions raise the test count, not the module's assertion tally, and `minimum_assertions` is a floor.
- [x] `ruff check` and `shellcheck --severity=warning -e SC1091` clean on all four changed files.
- [ ] Full `lib/test/run.sh` — running; result posted below before this is called ready.

**Disclosed non-finding.** Substituting `bisect_right` for `bisect_left` at both production call sites leaves the oracle test green. That is not a coverage gap: both call sites pass a position that is never itself a newline offset (a name character, and the closing `}`), which is the only input on which the two differ. The off-by-one mutations above are what actually exercise the derivation.

**Not fixed, deliberately.** `_definition_scan`'s anonymous nested-tuple return was raised as a `NamedTuple` candidate. It is a style suggestion against a correct implementation, on the production parse path, so it is left out of a follow-up whose subject is defects — worth doing on its own.

## Resolves
Follow-up to #886 / issue #881. No acceptance criterion of #881 regresses: no test is removed, no audited source set narrows, and `grep -cF 'test_pin_corpus_lint.py' lib/test/run.sh` still returns `0`.

## Breaking Changes
None. Confined to `lib/test/` and `docs/`, which `install.sh` does not vendor, so no consumer observes it — hence no changeset, per the repo's internal-only rule.


The01Geek:
Addressed the REJECT on `8f25aa2c` in `2f147070`.

**VC-8 (the REJECT driver) — confirmed and fixed.** The claim was literally false: `_DEFINITION_RE` and `_DEFINITION_TEXT_RE` are module-level `dict`s. They are built once at import and never written to, so the ordering-independence argument the limb supports still holds — but the universal as written did not. Both coupled sites now read "nothing is mutated after import", and each names the regex dicts so the claim is checkable rather than asserted. Correcting a false prose claim is this PR's own stated mission, so this belonged here.

**Important #1 — confirmed, and the diagnosis was exactly right.** `test_linter_image_memos_are_reused_within_one_extraction` presents a single image, so one live key stays resident at any `maxsize >= 1`. I verified the failure mode directly: with `_IMAGE_PARSE_CACHE_SIZE` lowered to `1`, **that test still passes**. The bound of 2 buys *cross-extraction* reuse, which nothing covered.

Split in two rather than softened:
- the original keeps the key-match property, with its over-claim about the bound removed and a pointer to the test that does guard it;
- `test_linter_image_memo_bound_retains_a_second_image` drives two distinct images, then re-presents the first and asserts it takes no new miss. At `maxsize=1` the second evicts the first and it goes RED.

**Suggestion #2 — fixed.** The oracle anchored on the first occurrence of `name(` and of `body`, so a call site preceding a definition (or two byte-identical bodies) would compare against the wrong span and go RED on a *correct* derivation. It now pairs each body occurrence with its nearest preceding header and accepts the reported pair if any candidate yields it. An off-by-one shifts every candidate at once, so it still fails — re-verified below.

**Suggestion #1** (`_definition_scan`'s anonymous tuple → `NamedTuple`) is deliberately still not taken: it is a style change on the production parse path, and this PR's subject is defects.

### Mutation verification (control green in each case)

| Mutation | Test | Verdict |
|---|---|---|
| `_IMAGE_PARSE_CACHE_SIZE` 2 → 1 | new bound test | **FAILED** (old test still passed — the gap, demonstrated) |
| start-line `bisect + 1` → `+ 2` | counting oracle | **FAILED** (4 subtests) |
| end-line `bisect + 1` → `+ 2` | counting oracle | **FAILED** (4 subtests) |

- `python3 lib/test/test_pin_corpus_lint.py` → **131 tests, OK**
- `lib/test/run-module.sh harness-python-guards` → **38 passed, 0 failed** — equal to the registry floor and unchanged, since the Python file folds into one assertion (added tests raise the test count, not the tally)
- `ruff` + `shellcheck --severity=warning -e SC1091` clean on all changed files

Issue #671, opened 2026-07-21
Close plugin packaging gaps: CI validation of the plugin tree, agent frontmatter, license delivery, and manifest metadata


The01Geek:
## Problem Statement

The repo's packaging surface has a verification hole and a delivery hole that reinforce each other.

`claude plugin validate .` at the repo root prints `✔ Validation passed` while checking only the marketplace manifest — the plugin tree, including every agent and skill frontmatter, is never examined. Under that unwatched surface, two agent files carry YAML frontmatter that fails to parse, and neither CI nor any local run reports it.

Separately, the vendored plugin slice that consumers actually receive omits `LICENSES/`. Since the vendored agent and skill files deliberately carry no in-file attribution marker, a consumer's materialized plugin tree contains third-party Apache-2.0 and MIT prompt files with no license text and no attribution of any kind.

Alongside these, the committed config carries a dead provider-routing block no code path reads, the plugin and marketplace manifests are missing metadata that listing surfaces and tooling consume, and `CITATION.cff`'s version has drifted behind the manifest with nothing in the repo bumping, reading, or validating it.

## Current Behavior

**The root-level validator covers only one manifest.**
- Verified: `claude plugin validate . --strict` → `Validating marketplace manifest: .claude-plugin/marketplace.json` / `✔ Validation passed`. The plugin manifest and all component frontmatter go unexamined because the repo root holds both manifests and the tool resolves the marketplace one.
- Verified: `claude plugin validate --help` lists exactly one option, `--strict`. There is no manifest-selection flag.
- Verified: validating a plugin-only tree (a scratch directory holding `.claude-plugin/plugin.json` plus the component directories) reports `✘ Validation failed` with two frontmatter errors.

**Two agents have unparseable frontmatter.**
- Verified: a PyYAML `safe_load` sweep over all 28 `agents/*.md` + `skills/**/SKILL.md` frontmatter blocks fails on exactly two files and parses the other 26.
- `agents/checklist-verifier.md` — its `description` is an unquoted plain scalar containing `` `verification_mode: "agent"` ``. The `: ` makes YAML read a nested mapping key.
- `agents/silent-failure-hunter.md` — same defect from `Context: Daisy has just finished implem…`. Its description additionally contains literal `\n` escape sequences and three `<example>` blocks narrating a persona named "Daisy", carried over from the upstream `pr-review-toolkit` agent. Verified: `git grep -l Daisy` matches this file alone.
[...]

Issue #456, opened 2026-07-13
Report skipped checks in `lib/test/run.sh`, and stop a skipped gate being recorded as a clean pass


The01Geek:
## Problem Statement

`lib/test/run.sh` has exactly two counters — `PASS` and `FAIL`. A check that **self-skips** increments neither. It prints a `NOTE` line and is otherwise invisible: the terminal summary (`N passed, M failed`) and the exit status (`[ "$FAIL" -eq 0 ]`) are **byte-identical** whether a gate ran and passed or never ran at all.

So a reader — human or agent — cannot distinguish **"the gate passed"** from **"the gate never fired."**

This is not hypothetical, and it is not confined to local runs. The `#434` stale-prose self-scan is **skipping in CI on every run**, because `actions/checkout` fetches a single commit by default and `origin/main` therefore does not resolve. CI run [29221871448](https://github.com/The01Geek/devflow-autopilot/actions/runs/29221871448) reports:

```
NOTE  #434 self-scan skipped — origin/main not resolvable in this checkout
5746 passed, 0 failed
```

The gate meant to catch stale counted prose has been dead in CI, while CI reported green.

It also skips locally whenever the working tree is dirty — the normal state while developing. During PR #442 this laundered a real failure: the suite reported `0 failed`, the stale-prose gate had not run, and two `STALE` rows only surfaced when a blinded reviewer re-ran the suite on a clean tree.

The stakes are highest for DevFlow's own automation. Per issue #405 and `CLAUDE.md`, an **in-env suite run is the verification gate** during a cloud `/devflow:implement`, and `/devflow:review-and-fix` records the outcome as `verification_evidence`. An agent reading `0 failed` records `result: pass` — off a gate that may never have executed. That recorded `pass` is the laundering, and it is what this issue closes.

## Current Behavior

- `run.sh` derives `PASS` and `FAIL` by counting `PASS`/`FAIL` lines in `RESULTS_FILE`, prints `echo "$PASS passed, $FAIL failed"`, and exits on `[ "$FAIL" -eq 0 ]`.
- Three checks self-skip today, each via a bare `printf '  NOTE  …'` that touches no counter:
  - `#423 T6b` — the host's Python build cannot express the strict-ASCII condition the check probes.
  - `#434` — `origin/main` is not resolvable in this checkout.
  - `#434` — the working tree is dirty (the scan grades committed `HEAD`).
[...]

PR #907, opened 2026-07-28
ci: install the Claude Code CLI so the #671 plugin-validate gate arms

The01Geek:
## The problem

The `#671` `claude plugin validate --strict` gate in `lib/test/run.sh` **self-skips on every CI run**. No runner has the `claude` CLI on PATH, so the gate takes its `blocking-gate` skip branch. In the run cited below it was the suite's only skip:

> run `30274021296` @ `fb5256b7` — `NOTE #671 claude plugin validate --strict (plugin tree) skipped [blocking-gate] — claude CLI not on PATH`, `12523 passed, 0 failed, 1 skipped`

Because a skip exits 0 (issue #456), the gate was disarmed **silently** while CI stayed green.

## What I verified before changing anything

Every claim in the report reproduces:

| Claim | Result |
|---|---|
| Staged tree is `plugin.json` + `agents skills commands hooks` only | ✅ confirmed in source |
| Staged tree `--strict` → `EXIT=0` | ✅ reproduced |
| Add root `CLAUDE.md` → `EXIT=1` | ✅ reproduced (`root: CLAUDE.md at the plugin root is not loaded as project context`) |
| `ci.yml` installs no `claude` CLI | ✅ confirmed |
| Validation is auth-free | ✅ `ANTHROPIC_API_KEY` + `CLAUDE_CODE_OAUTH_TOKEN` unset, `HOME` redirected → `EXIT=0` |
| Root `.mcp.json` / `.claude/settings.json` not inspected | ✅ planted deliberately-corrupt files in both; still `EXIT=0` |

**The "not vacuous" correction also holds, and I pushed on it** — it is the reason this fix is worth making. Against the staged tree I mutated `skills/review/SKILL.md` and `agents/code-reviewer.md`:

| Mutation | Exit |
|---|---|
| Malformed YAML frontmatter in a **skill** | `1` |
| Malformed YAML frontmatter in an **agent** | `1` |
| No frontmatter block at all | `1` |
| Empty `SKILL.md` | `1` |
| `name:` key removed | `0` — `name` is optional, inferred from the directory |
| `skills/review/` deleted entirely | `0` — absence is not an error |

So the gate really does descend into all 18 skills + 10 agents. Arming it buys genuine manifest/frontmatter coverage.

## The change

Install a **pinned** CLI (`2.1.212`, current `stable`) in the `shard` job, then verify it resolves on PATH.

- **Pinned, not `stable`/`latest`** — same reasoning as the lint job's ShellCheck pin, but sharper: here the CLI *is* the validator, so an unpinned upgrade that adds a new warning would redden this **required** check on an unrelated PR with no repo change.
- **`set -o pipefail` is load-bearing** — `curl | bash` with a failed download otherwise feeds bash an empty script and exits 0, restoring the exact silent-self-skip this PR removes. The separate `claude --version` step then fails loudly (`GITHUB_PATH` only applies to *later* steps, so it cannot be folded in).
- **Installed on every shard uniformly**, matching the existing `fetch-depth: 0` rationale, so a gate that later moves into a module shard does not quietly revert to self-skipping.
- **No secret needed**, which matters because this job runs arbitrary PR code (see the `on:` comment on why `pull_request_target` is avoided).

A `run.sh` guard asserts the shard job installs the CLI, so removing the step cannot silently disarm the gate again. It mirrors the neighbouring `#456 ci.yml: the shard job checkout sets fetch-depth: 0` pin — the identical failure class, a `ci.yml` setting whose absence *disarms* a gate rather than failing it. It matches on the installer URL (a renamed step still arms the gate) and is scoped to the `shard` job.

## Scope I deliberately did not take

**Root `CLAUDE.md` stays outside the staged tree.** Adding it would turn the gate RED immediately — the warning above is real, not hypothetical. The staged tree exists precisely to isolate the plugin from repo-root files (see its comment re: `marketplace.json`). Worth noting for a separate issue: the repo is its own marketplace, so for a marketplace install the repo root *is* the plugin root and that warning is user-visible. That is a product question, not a CI one.

## Verification

- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` → clean
- `actionlint` with the CI job's `SHELLCHECK_OPTS` → clean
- New guard: passes on the real `ci.yml`; returns `no` under two negative controls — install line stripped, and the step moved to a different job (proving job-scoping, not a file-wide match)
- Full `lib/test/run.sh` result reported in a follow-up comment

**Changeset**: `.changeset/ci-install-claude-cli.md` (`bump: patch`). The `ci.yml` + `lib/test/run.sh` half is internal-only — `ci.yml` is not in `install.sh`'s workflow copy loop (`devflow`, `devflow-runner`, `devflow-implement`, `devflow-review`, `telemetry-push`), so it never reaches consumers — but the two new `scripts/*.sh` helpers are enumerated engine surface, which is what the changeset covers.

### Effect on the CI skip population — measured, not predicted

This arms `#671`; it does **not** drive CI to zero skips, and an earlier draft of this
description wrongly predicted that it would. Measured on both sides:

| run | skips |
|---|---|
| `main` @ `30399500061` | 2 — `#671` (CLI not on PATH) + `#434` (working tree dirty) |
| this branch @ `30395343556` | 1 — `#434` alone |

`#434`'s dirty-tree skip is **pre-existing and orthogonal** — it has nothing to do with the
CLI. Its cause is now diagnosed rather than left open: `ci.yml` points
`DEVFLOW_SHARD_TALLY_DIR` at `<workspace>/shard-tally-out`, *inside* the working tree
instead of `run-shard.sh`'s already-ignored `.devflow/tmp/` default; `run-shard.sh` creates
it before invoking the suite; and it matched no `.gitignore` rule, so `git status
--porcelain` reported `?? shard-tally-out/`. A `.gitignore` entry in this PR closes it, so
this change re-arms **two** blocking gates rather than one — a gate disarmed by a path
choice is the same failure class as one disarmed by a missing binary.

No total is predicted here; read the skip population off a run's own `NOTE` lines.

## Deferred / recorded decisions

- **Unpinned installer fetch (round-5 Suggestion 11).** `curl -fsSL … | bash` refetches
  `install.sh` unchecksummed on a required check that runs arbitrary PR code. *Why deferred:*
  the version argument pins the CLI but not the installer, which the step's own comment
  discloses, and the blast radius is bounded — the job is `contents: read` and holds no
  secrets. *Revisit if:* the job ever gains a secret or write scope, or upstream publishes a
  checksum/pinned-ref install path.
- **`sleep` exit status is unchecked** (round-6 Important 3). `[ "$delay" -gt 0 ] && sleep
  "$delay"` discards `sleep`'s status, so an absent or interrupted `sleep` collapses the
  backoff and makes the just-printed "retrying in Ns" false. *Why deferred:* it fails loud
  (a red required check), never silently, and `sleep` is present on every supported runner.
  *Revisit if:* the retry is reused somewhere a collapsed backoff would be silent, or a
  runner without `sleep` appears.
- **Two message/guard precision items** (round-6 Important 4, Suggestion 4): the empty-output
  path renders `got: ` with a blank operative half, and a whitespace-only command string
  passes the never-retry-nothing guard. *Why deferred:* both fail closed with the correct
  exit code and are driven by `run.sh`; the defect is message/guard precision, not behaviour.
  *Revisit if:* either helper gains a caller outside `ci.yml`, where a blank diagnostic or a
  whitespace command would not be caught by the fixed call site.
- **Pushback on round-5 Suggestion 1** (settled). The finding correctly spotted the
  contradiction but named the wrong side as false: it reasoned that a fresh `actions/checkout`
  is clean, so `ci.yml`'s clause must be the falsehood. The two runs tabled above show the
  dirty-tree arm actually firing, so `ci.yml`'s clause was accurate and the `#434` block's
  "BOTH skip arms are inert in CI" was not. Corrected the latter, kept the former — and
  round 6 then supplied the cause, which this PR fixes. *No revisit condition outstanding.*
- **Partial pushback on round-6 Suggestion 1 / VC-15.** The finding is right that
  "a fresh `bash -c` inherits no shell options" is a false absolute — an exported
  `SHELLOPTS` does carry them. Its *stated mechanism* is not right, though: bash does not
  automatically export `SHELLOPTS`. Measured on bash 5.3 and 3.2, `SHELLOPTS` is absent from
  a child `bash -c`'s environment (`env | grep -c '^SHELLOPTS'` → 0) and `pipefail` reads
  `off` in the child; it propagates only after an explicit `export SHELLOPTS`. So the
  suggested rewording would have replaced one inaccurate mechanism with another. The comment
  now asserts no inheritance mechanism at all and rests on two directly checkable facts.
  *Revisit if:* a supported runner ships a bash that does export `SHELLOPTS` by default,
  which would not change the comment (it makes no claim either way) but would matter to
  anyone reasoning about nested option propagation here.

---

**Verification evidence:** `lib/test/run.sh` — **13343 passed, 0 failed, 0 skipped**
(captured at `.devflow/tmp/verification-pr907.log`). CI for this head: all six checks green,
**13340 passed, 0 failed**, zero `NOTE` lines — the first run with **both** blocking gates
armed. `completion-check: pass — all completion evidence current, bound, and durable`.
Branch-sync (same turn): fetch ok, ahead 0 / behind 0 vs remote, behind 0 vs base, working
tree clean at `2fbb87d9`.


The01Geek:
-

The01Geek:
-

The01Geek:
**Shepherd wind-down record.** Head `5a07096d`, `behind_by` 0, pushed fast-forward from `6e930424`.

## Original CI failure — single root cause

The round-3 commit added `scripts/assert-cli-version.sh` but not its `lib/test/modules/coverage-map.json` entry. That one omission drove all three failing shards: `modules-pin` and `monolith` via `#591 coverage-map guard` arm1 (a git-tracked depth-1 unit matching a coverage pattern with no owner), and `modules-rest` via `#619 A1/A2 regenerate-artifacts` downstream (the `coverage-map-ratchet` row cannot report clean while the map is drifted). The monolith `HostCapabilitySkipChannelTests` / `ModuleRunnerTests` failures were the same guard failing through the module-runner path.

## Concurrency note

A sibling agent working from `.claude/worktrees/ci-claude` was on this PR simultaneously and pushed the identical coverage-map fix as `2d3dd0e4`, then later `6e930424`. Its work was **not** force-pushed over — the duplicate was discarded twice and `5a07096d` was rebased on top, carrying only what `6e930424` did not cover.

## What `5a07096d` adds

- Two `documented_falsehood` fixes in `ci.yml` comments: the "suite's ONLY skip" claim (false against the base ref — main runs report 2 skips) and "the literal is also pinned in run.sh" (only key-presence was pinned).
- Two coverage gaps closed with executable arms: stderr stream-routing for both `::error::` messages (the existing `2>&1` arms would stay green if `>&2` were dropped), and two parser boundary rows over `got="${actual%% *}"`.
- One silent failure: `devflow_pkg_validate_rc` discarded validator output; it now replays that output plus the observed `claude --version` on an unexpected rc, with both call sites updated in the same commit.
- Comment corrections: a self-referential shard count, a wrong pin precedent (ShellCheck vs ruff), a "the next step" misdirection, a mismatched cross-reference, and the stale `#671` skip reason.

The agent's own changeset was dropped — `6e930424`'s supersedes it.

## Verification — read the caveat

Full local suite run twice, both green, both read from the log's own summary line; the final pre-rebase run was **`13313 passed, 0 failed`** with no `, K skipped` clause and zero `FAIL` lines. Lints clean throughout (shellcheck, ruff, actionlint, coverage-map guard).

**However, that run was on the pre-rebase commit.** After rebasing onto `6e930424` the specific re-applied arms were driven individually (all pass) and the lints and coverage guard re-run, but **the full suite was not re-run on the final `5a07096d`** — wind-down was explicit not to start new verification. CI on the pushed head is the outstanding check.

## Review state

The `review-and-fix` loop ran through Phase 3 (five agents; `type-design-analyzer` correctly gated out) but **did not reach a formal terminal verdict** — wind-down landed during the fix step, so Step 2.6's shadow pass never ran. The final-pass reviewer's own assessment was *"Ready to merge: with fixes"*, its one Important blocker being the changeset, which `6e930424` supplies.

**Not addressed:** the cloud DevFlow Review on `2d3dd0e` returned REJECT and those findings were not worked.

The01Geek:
-

PR #911, opened 2026-07-28
chore: differentiate vendored skill/agent descriptions from upstream

The01Geek:
## Problem

Three related metadata defects, all invisible in this repo alone and only observable in a session with the upstream plugins also installed:

1. **Both vendored skills** (`receiving-code-review`, `requesting-code-review`) carried `description:` bytes identical to `superpowers` 6.1.1 **and** 6.2.0, while their bodies have diverged 9.2x and 2.2x — hard forks with identical triggering metadata. Both pairs were live in one session's skill listing with the same description and no basis to pick the DevFlow fork.
2. **Six vendored agents** (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `type-design-analyzer`, `code-explorer`, `code-architect`) carried descriptions byte-identical to the currently-listed Anthropic plugins they were vendored from. Only `silent-failure-hunter` already differed. A user with both plugins installed pays for and sees duplicates.
3. **Two subagent-only skills** (`retrospective`, `retrospective-audit`) ended their descriptions with "Invoked as a subagent — do not call it directly" while remaining model-invocable and menu-visible — `disable-model-invocation` appeared in exactly one skill, `init`.

## Change

- Each vendored skill/agent description now names its DevFlow engine role or trigger surface ahead of its unchanged triggering conditions. Bodies, tools, and model frontmatter are untouched. The two vendored skills stay repo-agnostic (no `lib/test/run.sh`, no internal step numbers) so consumer installs read correctly.
- `disable-model-invocation: true` on both retrospective skills, matching `init`.
- Regenerates `scripts/devflow-cloud-writer-contract.json`, which SHA-256-pins both review `SKILL.md` assets.
- Adds `.changeset/vendored-review-skill-descriptions.md` (`bump: patch`) — this touches the engine surface (`skills/`, `agents/`).

## Why gating the retrospective skills is safe

`/devflow:retrospective-weekly` dispatches Stage A and Stage B by instructing each subagent to **read** that stage's `SKILL.md` file, not through the Skill tool. Verified in `skills/retrospective-weekly/SKILL.md` — the dispatch prompt is `Read and follow "${CLAUDE_SKILL_DIR:-…}"/../retrospective/SKILL.md`. Model-invocation gating cannot break the loop.

## Verification

- `lib/test/validate-frontmatter.py .` → `OK 28 frontmatter files parsed`
- `lib/test/cloud_writer_contract.py check` → `closure OK` (after regeneration)
- Changeset parses under `scripts/consolidate-changesets.py`'s `_parse_changeset`
- Full `lib/test/run.sh` — running locally at the time of opening; result reported in a follow-up comment. CI `lib + python tests` is the merge gate.

## Note on scope

`silent-failure-hunter` already differed from upstream in bytes (DevFlow reflowed it in `80166b0c`), but like the other six it never named DevFlow, so it is included in this change: **all seven** vendored agents get a DevFlow role sentence. Problem item 2 above enumerates the six that were still byte-identical to upstream — that is the pre-existing-state count, not the change's scope.

## Writing-skills evidence

`superpowers:writing-skills` invoked before touching any `SKILL.md`; edits are frontmatter-only description/flag changes with no body or workflow content. RED-phase subagent pressure scenarios were **not** run — this session's constraints bar dispatching agents unless requested. The baseline failure is the observed one: identical metadata live in a single skill listing.

---

## Review-and-fix pass (local, PR #911)

Writing-skills evidence: the two vendored `SKILL.md` description edits in `439935cf` were routed through `superpowers:writing-skills` in a context-isolated subagent, which treated the `description:` field as the skill's trigger predicate and applied RED/GREEN. RED: a mechanical predicate check over both files failed 6 of 12 assertions against the narrowed text (neither description contained its upstream trigger conditions; both scoped triggering to a DevFlow antecedent). GREEN: 12/12 after the prepend-a-role-sentence rewrite. Frontmatter constraints (<=1024 chars, third-person, `Use when...` phrasing preserved) and the vendored-skill repo-agnostic constraint were re-checked; no body text changed.

Verification evidence: `lib/test/run.sh` on `b22c8f17` (this head) — **13274 passed, 0 failed** (the summary rendered the `K==0` arm, so the skip population is empty). Captured at `.devflow/tmp/verification-pr911-b22c8f17.log`. An earlier identical run on `439935cf` is superseded by this one. `python3 lib/test/regenerate-artifacts.py` reconciled clean (cloud-writer manifest, capability-profile literals, coverage-map ratchet); `python3 lib/test/validate-frontmatter.py .` → `OK 28 frontmatter files parsed`.


The01Geek:
**Finding rejected:** the Phase-3 `code-reviewer` Critical asserting that **five** `pr-review-toolkit` agents diverged from upstream.

**The finding.** It inferred, from `git log -S'## When to invoke'`, that DevFlow authored the description tail at import time, and concluded five vendored agent descriptions were no longer byte-identical to upstream.

**Verdict: not a defect — refuted by direct evidence.** The inference was from git history, not from the upstream bytes. Comparing our descriptions against the installed upstream plugins (`~/.claude/plugins/marketplaces/claude-plugins-official/plugins/{pr-review-toolkit,feature-dev}`) raw-line-by-raw-line:

| Agent | Ours | Upstream | Identical |
|---|---|---|---|
| `code-reviewer` | 985 B | 985 B | yes |
| `comment-analyzer` | 505 B | 505 B | yes |
| `pr-test-analyzer` | 516 B | 516 B | yes |
| `type-design-analyzer` | 565 B | 565 B | yes |
| `code-explorer` | 208 B | 208 B | yes |
| `code-architect` | 243 B | 243 B | yes |
| `silent-failure-hunter` | 423 B | 1441 B | **no** |

The `See "When to invoke"…` tail is present in upstream too, so its presence in our copies is not evidence of local authorship. Exactly **one** agent diverges — `silent-failure-hunter`, reflowed by DevFlow in `80166b0c` — and that one is fixed in this PR rather than rejected.

This rejection is recorded so a later review round re-raising the five-agent framing does not re-spend a fix cycle on it. It yields only to new evidence that the byte comparison above did not consider; a contrary opinion on the same evidence does not reopen it.

_Separately deferred (recorded, not fixed, pre-existing and untouched by this diff):_ `lib/test/validate-frontmatter.py`'s `REQUIRED_KEYS` is a subset test with no unknown-key arm, so a misspelled `disable-model-invocation` would fail open silently. Worth hardening if that flag is ever relied on as a guarantee.

The01Geek:
**Shepherd wind-down record.** Head `a483d343`, `behind_by` 0. Full local suite on this head: **13274 passed, 0 failed, 0 skipped**. `regenerate-artifacts.py` all clean, `cloud_writer_contract.py check` → `closure OK`, `validate-frontmatter.py .` → `OK 28`, `stale-prose-lint.py --rev HEAD` rc 0. No pin targets any literal removed (`git grep -nF` over `lib/test/ scripts/ .github/ docs/ skills/ agents/` → none).

## Resolved

Both review findings are closed, and the disputed Critical is settled in both directions:

- **Horn 1 (five agents diverged from upstream) — rejected**, consistent with the earlier rejection comment. Byte comparison of `git show b82d3b2:agents/<a>.md` against the installed upstream plugins: six print `IDENTICAL`, only `silent-failure-hunter` prints `DIFFERS`. The `See "When to invoke"…` sentence **is present upstream** — the reviewer inferred DevFlow authorship from a `git grep` over our tree alone.
- **Horn 2 (was "a DevFlow reflow of upstream's text" accurate?) — resolved: it was accurate.** Upstream carries the same first two sentences then an `Examples:` / `<example>` tail; DevFlow replaced that tail with the pointer sentence. The characterisation was correct. The reviewer's suggested fix was taken anyway, because dropping the count removes the dispute surface entirely and cannot rot.

Also dropped from the changeset: the `9.2x` size multiple and the unverifiable "already divergent at its own import commit" clause — that clause was the engine's INCONCLUSIVE `VC-15`, so removing it clears the Rule-2 REJECT driver too. `docs/DEVFLOW_SYSTEM_OVERVIEW.md:506` was reworded narrowly to ground the hazard on the shared **name** rather than on description identity; the always-use-the-plugin-qualified-name rule is retained verbatim.

## The review-and-fix pass was CUT SHORT, not converged

The full gated Phase 3 roster ran and all four agents returned **finding no defect in the shipped diff** — three independently confirmed the two fixes as correct. But Step 2.5/3 disposition, the shadow pass, and Loop Exit were never reached before wind-down. Treat this as un-converged.

## Open — all paperwork, none in shipped code

1. **Important (2/4 agents)** — the PR description's "Note on scope" section is false against HEAD: it says `silent-failure-hunter` "was left alone", but the diff modifies it, and Problem item 2 says six agents where the changeset correctly says seven. A reader would file a follow-up for work that already shipped.
2. **Important (1/4) — time-sensitive.** The changeset carries no `(#911)` PR citation, which `CONTRIBUTING.md`, the implement extension, and `consolidate-changesets.py`'s own docstring all mandate. **After merge the changeset is deleted**, so this becomes fixable only by hand-editing `CHANGELOG.md`. Worth fixing before merge.
3. **Suggestion (3/4)** — `prepend-a-role-` / `sentence` is broken across lines in the changeset; markdown will render it as `prepend-a-role- sentence` in the published CHANGELOG.
4. **Suggestion (1/4)** — "so neither skill's non-DevFlow trigger surface is narrowed" is an unmeasurable behavioral prediction shipping to users; the checkable claim is that the trigger conditions are preserved verbatim.

Noted, not filed: `agents/code-reviewer.md`'s description is now 1067 chars (no repo-enforced cap found on `agents/*.md`), and the PR body's `Verification evidence:` figure still cites `439935cf` rather than HEAD.

Issue #628, opened 2026-07-20
Calibrate quantitative claims in the code-explorer/code-architect agents and remove their inert KillShell/BashOutput grants


The01Geek:
## Problem Statement
`/devflow:implement` dispatches two discovery/planning subagents — `devflow:code-explorer` (Phase 2.1) and `devflow:code-architect` (Phase 2.2 Path B) — whose bodies mandate confident, decisive output while granting no arithmetic capability, no reliable counting channel (under current Claude Code tool semantics — a harness-behavior premise — Grep's count mode reports matching lines rather than occurrences, and search results truncate at output limits), and no instruction to mark unverified numbers. The failure mode is calibration, not a missing tool: neither body requests word counts, sizes, or estimates — the only quantitative ask is `file:line` precision, which Read/Grep fully supports — so an agent that volunteers an unrequested measurement class (a word count, a size total, budget arithmetic) does so with mandated confidence and no calibration path. A run incident was observed in which an architect blueprint carried a budget-arithmetic error of roughly 6,100 words, caught only because the orchestrator independently re-derived the number (a run observation, not verifiable from this repository).

This repo's own budget contracts show why plausible measurements matter here: the review-bundle budget is counted by `_rb_words`, never `wc -w`, because a measurement that reproduced at one desk and nowhere else shipped into the record (CLAUDE.md's review-bundle bullet).

## Current Behavior
- `agents/code-architect.md` and `agents/code-explorer.md` declare identical `tools:` frontmatter: `Glob, Grep, LS, Read, NotebookRead, WebFetch, TodoWrite, WebSearch, KillShell, BashOutput`. Neither grants `Bash`; under current Claude Code Agent-tool semantics a subagent without `Bash` has no shell session for `BashOutput` to read or `KillShell` to kill, so both grants are inert — a harness-behavior premise, not a repo-verifiable fact, falsified if the Agent tool ever exposes parent or background shell sessions to subagents.
- `agents/code-architect.md` instructs: "Make decisive choices - pick one approach and commit." and "Make confident architectural choices rather than presenting multiple options." `agents/code-explorer.md` closes with: "Always include specific file paths and line numbers." No sentence in either body addresses quantitative claims the agent cannot derive from tool output.
- `skills/implement/phases/phase-2-implement.md` §2.2 Path B consumes the architect's blueprint directly ("Hold this blueprint in your context — do NOT commit it") with no obligation to re-derive quantitative claims before relying on them.
[...]

Issue #504, opened 2026-07-15
Cloud review: surface the #458-displaced Stop-hook paths as engine ground truth and route their HEAD verification through `git show`


The01Geek:
## Problem Statement

The cloud auto-review tier deliberately dirties the PR-head workspace before the engine runs: `devflow-runner.yml`'s "Harden Stop-hook script sources" step (issue #458) displaces the ten Stop-hook-closure files with trusted base-ref copies (else fail-closed stubs) so a PR cannot obtain unmediated shell via the base-branch Stop hooks. That floor is correct and stays untouched — but the review engine has no awareness of it. The engine's truthfulness contract tells every Phase-3 agent to "verify the claim against the shipped code (read the named symbol …)", and the Phase 4.1.6 sweep does the same — all through **working-tree reads**, which for the displaced paths return **base-ref bytes, not the reviewed head**.

Observed live on PR #495 (branch `worktree-issue-489`, which edits two closure files): repeated review runs manufactured a Critical `documented_falsehood` — the **non-demotable REJECT class** — against diff-added wording that is correct at HEAD (the committed "ships (#489)" text read as "false" because the working-tree copy carried the base branch's older "forthcoming" wording). Two such runs are documented in the surviving `devflow:review-progress` record; the run-keyed comment is edited in place, so earlier rounds are not durably recorded. The verdicts survived only because Phase 4 aggregation scoped to the committed diff. Every future PR touching a closure file pays the same burned agent budget and carries the same wrong-REJECT risk — and the closure files are exactly the ones DevFlow's own telemetry/hook work keeps editing.

`docs/DEVFLOW_SYSTEM_OVERVIEW.md` §15's #458 bullet already discloses this as residual (3) and names the fix: *"A future refinement would surface the hardened paths to the review engine so their verification routes through `git show $PR_HEAD_SHA:<path>`; until then the divergence is disclosed here."* This issue implements that refinement.

## Current Behavior

[...]

Issue #141, opened 2026-06-26
Internalize the pr-review-toolkit review agents as first-party DevFlow assets (seam 2 of #139)


The01Geek:
## Problem Statement

Follow-up to #139 (seam 2 of 3). The parent issue internalizes every external Anthropic asset DevFlow's engine dispatches as a first-party DevFlow file so the retrospective-weekly self-improvement loop can evaluate and improve it. Its /devflow:implement run scoped the first PR to the `feature-dev` agents only (per implement Phase 2.2.5 — the issue body mandates a one-PR-per-source-plugin split: `feature-dev`, then `pr-review-toolkit`, then `superpowers` + dependency removal). This issue covers the **`pr-review-toolkit`** seam: the five Phase-3 review agents that today sit permanently outside the self-improvement loop because they are external plugin files DevFlow cannot edit.

## Current Behavior

`skills/review/SKILL.md` Phase 3.1 and `skills/review-and-fix/SKILL.md` dispatch the five `pr-review-toolkit:*` agents (`code-reviewer`, `silent-failure-hunter`, `comment-analyzer`, `type-design-analyzer`, `pr-test-analyzer`). `scripts/resolve-review-overrides.py` hardcodes these identifiers in its allowlist, and `.devflow/config.schema.json` / `config.example.json` key `devflow_review.agent_overrides` on them. The retrospective loop cannot target any of these — they are not files in this repo.

## Desired Behavior

The five `pr-review-toolkit` review agents are vendored in as first-party DevFlow files under `agents/`, the engine dispatches the `devflow:`-namespaced identifiers, the resolver/config/tests/docs reference only the internalized identifiers, and `pr-review-toolkit` is removed from the plugin's dependency set — so the retrospective loop can evaluate and propose improvements to the review roster.

## User Impact

The DevFlow maintainer (and the retrospective loop acting on their behalf) gains the ability to continuously improve the review agents that do the bulk of /devflow:review's work — not just the orchestration around them. Consumers lose one more companion-plugin install dependency.

## Technical Context

[...]

Issue #711, opened 2026-07-22
Make the test suite's repository-tree enumeration worktree-immune


The01Geek:
## Problem Statement

The local test suite carries one assertion that enumerates the repository tree with a raw
recursive filesystem walk instead of asking git what is committed. In any checkout that has
sibling git worktrees under `.claude/worktrees/`, that walk descends into every one of them and
counts their copies of the file it is trying to prove unique — so the check fails with a number
that has nothing to do with the repository's actual state.

Sibling worktrees under `.claude/worktrees/` are this repository's own working mode: the
`EnterWorktree` tool creates every worktree it makes inside `.claude/worktrees/`, and this
checkout currently holds nine of them carrying their own baseline file. So the failure is not an
exotic environment — it is the ordinary state of a working checkout.

CI never sees it. A fresh `actions/checkout` has no `.claude/worktrees/`, so the required
`lib + python tests` job stays green while every local run in a worktree-carrying checkout eats
a red suite.

This directly attacks a load-bearing contract. `CLAUDE.md` makes `0 failed` the thing a run's
verification rests on, and is emphatic that a non-green suite is never laundered into a pass. A
guaranteed, environment-dependent FAIL that the required gate cannot reproduce trains the
operator to explain away red — which is exactly the habit the contract exists to prevent. The
failure text makes it worse: `only one committed prompt-mass baseline exists` reads as a stale
artifact bug in the repository, sending the reader to hunt a duplicate that was never committed.

## Current Behavior

The suite check named `#551 only one committed prompt-mass baseline exists`, in the
`issue #551: mandatory prompt-byte census + prose-cutover policy` block of `lib/test/run.sh`,
computes its count with an inline `python3 -c` expression that calls
`Path(root).rglob("prompt-mass-baseline.json")` and filters out only the substring
`.devflow/vendor/`. Its root argument is the repository root.

Verified — executed 2026-07-21 in this checkout:

- The assertion's own expression returns `10`; the assertion expects `1`.
- The nine surplus hits are all of the form
  `.claude/worktrees/<name>/lib/test/prompt-mass-baseline.json`, across the worktrees
  `review-pr-625`, `issue-audit-preamble`, `pr-622`, `fix-625`, `pr-625`, `issue-704`,
  `pr-639b`, `issue-655`, `issue-690`.
- `git ls-files -- '*prompt-mass-baseline.json'` returns exactly one line,
  `lib/test/prompt-mass-baseline.json`.

[...]

PR #939, opened 2026-07-29
fix(agents): declare explicit tools: allowlists on the eight inheriting subagents

The01Geek:
## What

The three `checklist-*` agents and the five vendored `pr-review-toolkit` review agents carried **no `tools:` key**, so each inherited the full subagent toolset — `Edit`, `Write`, `MultiEdit`, `NotebookEdit`, `Task` included. Five of them state a read-only advisory working-tree policy *in their own bodies* ("never modify working-tree source files, the index, HEAD, or branch state"). That policy was requested in prose and unenforced by the harness.

Each agent now declares only what its body and its dispatch prompt actually use:

| Agent | `tools:` | Why |
|---|---|---|
| `checklist-deduper` | `Read` | Pure JSON→JSON merge; touches no file |
| `checklist-generator` | `Read, Grep, Glob` | Reads the cached diff by path, greps to ground claims |
| `checklist-verifier` | `Read, Grep, Glob, Bash, Write` | Steps 2–3 name Read/Grep/Glob; `git show`/`git cat-file` for #504 routing; **Write** for the verdict file |
| `code-reviewer` | `Read, Grep, Glob, Bash` | `git diff` scope, `git show` for #504, `mktemp` mutation-checks |
| `comment-analyzer` | `Read, Grep, Glob, Bash` | same |
| `pr-test-analyzer` | `Read, Grep, Glob, Bash` | same |
| `silent-failure-hunter` | `Read, Grep, Glob, Bash` | same |
| `type-design-analyzer` | `Read, Grep, Glob, Bash` | same |

`Task` is dropped everywhere — none of the eight dispatches a subagent (the "task" hits in `code-reviewer`/`comment-analyzer` are the English word).

## Two findings that shaped this

**`checklist-verifier` genuinely needs `Write`.** Its own body never mentions writing, but `skills/review/phases/phase-2-verification.md` instructs the dispatched agent: *"Write your verdict JSON to the file path {VERDICT_FILE} using the Write tool."* The grant is derived from the dispatch prompt, not from the agent body — checking only the body would have broken Phase 2.

**An empty `tools:` value is a silent fail-open, not the tightest restriction.** Measured on Claude Code 2.1.220 by probing all four forms against a control (`tools: Read`):

| Frontmatter form | Parsed as |
|---|---|
| `tools:` (bare/null) | `All tools` |
| `tools: []` | `All tools` |
| `tools: ""` | `All tools` |
| key omitted | `All tools` |
| `tools: Read` (control) | `Read` |

So there is no zero-tool agent, and an emptied value reverts to inheriting everything. The new suite rows therefore prove each value is **non-empty before** testing what it omits — otherwise every omission row would pass vacuously on exactly the regression they exist to catch. Agent frontmatter also has no deny/disallow key, so an allowlist is the only lever.

## What this does *not* claim

This is a **name-level** boundary, not read-only enforcement. The six agents holding `Bash` can still reach the working tree through it (`sed -i`, `git checkout --`, `git reset`). What it closes is the tool-shaped path — an agent reaching for `Edit`/`Write` on a file it was told only to report on. The Phase-3 dirty-tree snapshot guard in `phase-3-agents.md` remains the behavioral backstop.

Per-tool argument scoping (`Bash(git show:*)`, `Write(.devflow/tmp/**)`) is a possible follow-up: such entries parse and round-trip intact, but whether they are **enforced** as scoping — rather than read as a tool name that matches nothing — is not measured, and getting it wrong would silently strip Bash from a reviewer. Left out deliberately.

## Upstream provenance (checked against the sources)

Fetched verbatim via `gh api .../contents/...` (not a summarizer), so these are the upstream bytes:

- **The five `pr-review-toolkit` agents declare no `tools:` key upstream either.** `anthropics/claude-plugins-official@main`'s `plugins/pr-review-toolkit/agents/*.md` carry only `name`, `description`, `model`, `color`. So the inherit-everything state was **inherited verbatim from Anthropic**, not introduced when DevFlow vendored them — and this PR is a deliberate DevFlow divergence from upstream, on files already marked "MODIFIED by DevFlow".
- **The two `feature-dev` agents DO declare one upstream**, and it is the read-only idiom: `Glob, Grep, LS, Read, NotebookRead, WebFetch, TodoWrite, WebSearch, KillShell, BashOutput` — identical for both, and DevFlow's copies are that list minus `KillShell, BashOutput` (dropped as inert in #628). Note what it omits: **no `Bash`, no `Edit`/`Write`**. That is the shape Anthropic gives an analysis-only agent.
- **The three `checklist-*` agents are first-party DevFlow** — no upstream to compare.
- **Nothing to compare on the superpowers side.** `skills/requesting-code-review/code-reviewer.md` has no frontmatter upstream or locally (it is a prompt template dispatched as `general-purpose`, not an agent definition), and neither vendored skill declares `allowed-tools`.

The one place this cuts against the PR: upstream's read-only agents get **no `Bash` at all**, whereas five agents here keep it. That is not copyable — DevFlow's reviewers need `git diff`/`git show` for scope and #504 displaced-path routing, and the `mktemp` mutation-check the bodies mandate. It does sharpen where the remaining exposure is, and it is the same gap the "not read-only" caveat above names: `Bash`, not the four tree-mutation tools, is what still reaches the tree.

## Coupled sites

- Two `#141` comments in `lib/test/run.sh` recorded the old state ("these carry NO `tools:` key (they inherit all tools)") — corrected in the same change.
- New `#agent-tools` block in `lib/test/run.sh` asserts, per agent: a non-empty `tools:` value; `Edit`/`MultiEdit`/`NotebookEdit`/`Task` absent; and `Write` present **iff** the agent is `checklist-verifier` (asserted both ways, so neither a dropped grant that breaks Phase 2 nor a spread grant that breaches the policy is silent).
- `.changeset/subagent-tool-allowlists.md`, `bump: patch` (engine surface).

## Verification

- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` — clean (ShellCheck 0.11.0).
- `python3 lib/test/validate-frontmatter.py .` — `OK 28 frontmatter files parsed`.
- Full `lib/test/run.sh` running locally; result reported in a follow-up comment. No focused module covers `run.sh` main-shell assertions.

### DIFF
--- a/.changeset/subagent-tool-allowlists.md
+++ b/.changeset/subagent-tool-allowlists.md
@@ -1,0 +1,34 @@
+---
+bump: patch
+---
+
+### Changed
+
+- **Every DevFlow-authored subagent now declares an explicit `tools:` allowlist.** The three
+  `checklist-*` agents and the five vendored `pr-review-toolkit` review agents
+  (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`,
+  `type-design-analyzer`) previously carried no `tools:` key, so each inherited the full
+  subagent toolset — including `Edit`, `Write`, `MultiEdit`, `NotebookEdit` and `Task` —
+  while five of them state a read-only advisory working-tree policy in their own bodies
+  ("never modify working-tree source files, the index, HEAD, or branch state"). That policy
+  was requested in prose and unenforced by the harness. Each agent now names only what its
+  body and its dispatch prompt actually use: `Read, Grep, Glob` for `checklist-generator`,
+  `Read` for `checklist-deduper` (a pure JSON-to-JSON merge), and `Read, Grep, Glob, Bash`
+  for the five review agents, whose `git diff` / `git show` scope reads and `mktemp`
+  mutation-checks need Bash. `checklist-verifier` additionally keeps `Write`, because the
+  Phase 2.1b dispatch prompt instructs it to write its verdict JSON to `{VERDICT_FILE}`.
+  `Task` is dropped everywhere: none of the eight dispatches a subagent.
+
+  **This is a name-level boundary, not read-only enforcement.** The six agents that hold
+  `Bash` can still reach the working tree through it (`sed -i`, `git checkout --`, `git
+  reset`). What the change closes is the tool-shaped path — an agent reaching for `Edit`
+  or `Write` on a file it was told only to report on. The Phase-3 dirty-tree snapshot guard
+  remains the behavioral backstop, and per-tool argument scoping (`Bash(git show:*)`) is a
+  possible follow-up: such entries parse and round-trip, but whether they are *enforced* as
+  scoping is not yet measured.
+
+  Restricting an agent has one non-obvious failure mode, measured on Claude Code 2.1.220 and
+  now guarded: a `tools:` key whose value is **empty** — bare, `[]`, or `""` — parses
+  identically to omitting the key, so the agent inherits every tool. An emptied value is a
+  silent fail-open of this boundary rather than the tightest possible restriction, which is
+  why the new suite rows prove each value is non-empty before testing what it omits.
--- a/agents/checklist-deduper.md
+++ b/agents/checklist-deduper.md
@@ -1,6 +1,7 @@
 ---
 name: checklist-deduper
 description: Merges multiple batches of checklist items from checklist-generator into a single deduped checklist. Preserves traceability by recording which original IDs were merged. Does NOT judge correctness or re-tag items.
+tools: Read
 model: sonnet
 color: violet
 ---
--- a/agents/checklist-generator.md
+++ b/agents/checklist-generator.md
@@ -1,6 +1,7 @@
 ---
 name: checklist-generator
 description: Use when an orchestrator needs to enumerate every verifiable claim in a code diff (dependency interactions, test-mock alignment, data format assumptions, API contracts) and return a JSON checklist for independent verification. Does NOT judge correctness.
+tools: Read, Grep, Glob
 model: opus
 color: blue
 ---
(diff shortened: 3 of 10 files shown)


The01Geek:
## Verification result

**CI: all green**, including the required `lib + python tests`, plus `lint (shellcheck + actionlint + ruff)` and all four shards (`monolith`, `modules-large`, `modules-pin`, `modules-rest`) — run [30498799259](https://github.com/The01Geek/devflow-autopilot/actions/runs/30498799259).

**Local full suite: `14079 passed, 1 failed, 1 skipped`** — so it is *not* a clean local pass, and both non-passing rows trace to the same cause in this checkout, not to the diff:

```
FAIL  #664 scanner: clean on the tree as it stands
      lint-gh-api-repo-path: SKIPPED .worktrees/issue-810/: unreadable (IsADirectoryError)
      lint-gh-api-repo-path: audited 252 of 253 files (1 skipped)
SKIP  #434 stale-prose self-scan [blocking-gate] — working tree dirty
```

An untracked sibling git worktree at `.worktrees/issue-810/` exists in this working tree (it predates this branch). It makes the tree dirty, which arms the `#434` `blocking-gate` self-skip, and the `#664` scanner's working-tree enumeration selects the directory and refuses to report clean when a selected path cannot be audited. Re-running that scanner alone reports **only** the unreadable path — no violation in any changed file — and CI's fresh checkout, which has no worktrees, is green on both.

Worth noting as a possible follow-up, unrelated to this PR: the `#711` worktree exclusion documented for these scanners covers `.claude/worktrees/`, but worktrees created at `.worktrees/` are not excluded, so any local full-suite run in a checkout using that layout fails `#664` and skips `#434`.

**New rows from this change: 48 `#agent-tools` assertions, all PASS** (6 per agent × 8 agents).

PR #950, opened 2026-07-30
fix(lint): make the structural-pin-ok hatch usable via a three-step routing ladder (#948)

### LINKED ISSUE #948: The structural-pin-ok escape hatch is unusable: make the gate a three-step ladder (from #885)
Discovered while doing #885. Not a blocker for it — #885 shipped with the limit recorded in `CONTRIBUTING.md` — but it leaves a real maintenance trap in place.

## What the tag is for

A pin is a test asserting that an exact sentence appears exactly once in a file. Most of those sentences are read only by the runtime agent, so the pin freezes wording rather than protecting behavior; #375/#666/#810 stopped new ones and #885 retired 194 of the old ones. But some pinned sentences **are** load-bearing — they contain a marker a script parses, or the exact spelling of a command a runner is permitted to execute. For those, the intended escape hatch is a `# structural-pin-ok: <category> -- <why>` tag on the pin's own line.

**The hatch does not work.** Three layered reasons, all measured on `main` at `657854cc`.

## The measurements

674 retained existence-only pins across `lib/test/run.sh` and the modules:

| | count |
|---|---|
| carry a **valid** tag | **8** |
| carry **no** tag | 631 |
| carry a tag whose category is **not one of the eight legal names** | **35** |
| whose literal the gate classifies as prose (so no tag can help — see below) | 229 |

## The three failures

**1. The gate never consults the tag for a prose-resolving pin.** `scan_changed_sources` runs `_literal_prose_resolution` *before* the declaration check and `continue`s on a hit. Probed directly: adding a valid tag to one retained pin produced

```
MUTATION-ROUTING  lib/test/run.sh:1283  assert_pin_unique  re-opens the diff at every threshold value
  literal resolves into prose at skills/receiving-code-review/SKILL.md:199;
  the assert_pin_unique helper does not change the verdict
rc=3
```

**2. The prose test asks the wrong question.** It asks whether the sentence *reads like* prose, not whether anything *depends on* it. The interesting cases are both at once: an ordinary English sentence that spells out a command name a cloud grant must match exactly, or a field name a renderer enumerates. #885's per-site pass found many — that is why 229 pins are retained *and* classified prose. The check cannot distinguish them; a human reading the surrounding code can.

**3. 35 pins are already sitting on an invalid tag, and touching one turns the required gate RED.** They use free-text categories (`contract-presence over skill prose`, `grammar-declaration presence`) that predate the eight-name vocabulary. The gate only classifies a site whose lines are in the diff's added set, so they are invisible until someone edits the line — and then there is often no valid form to edit into. This bit #885 twice. One site (`lib/test/modules/create-issue-contract.sh`) is now **permanently frozen** with a warning comment above it: fixing its category surfaces a *second* failure, `typed structural declaration target cannot be inspected`, because its target is a concatenated in-module bundle the lint cannot resolve. Its trailing tag text is knowingly left stale because no edit to it can pass.

## Proposal: a three-step ladder, failing closed

Replace the single prose test with an ordered ladder in `scan_changed_sources`:

1. **Does a program demonstrably read this?** Mechanically decidable in many cases — the literal, or the token it names, appears in a `scripts/**`, `lib/**` (non-test) or `.github/**` file. Pass; no human needed.
2. **If not: does `lib/test/pin-corpus-adjudications.tsv` already record this literal as a boundary?** If yes, honour the `# structural-pin-ok:` tag and pass. The tag becomes a *pointer to an authorized decision*, not a self-granted permission — you cannot tag your own pin into legitimacy, because a ledger row needs an authorization bundle and cannot be self-granted either.
3. **Neither?** Report the finding. This is the pin the policy exists to remove.

### Three constraints, each load-bearing

- **Step 1 must fail toward step 2, never toward rejection.** A grep-based consumer check misses generic consumers by construction — a script that walks a routing table row by row never mentions any individual row. If "found no consumer" meant "reject", the gate would push maintainers into deleting pins that genuinely are load-bearing: this repo's named fail-open class, a guard whose comparand its producer does not always emit. "I did not find one" must mean "ask the ledger", not "there is not one".
- **Be honest about where the control lives.** After this change the gate is *routing*, not judging. The real safeguard is the review of ledger changes, which is the strong part — a ledger edit needs an exact delta manifest and a branch based on the current `origin/main` tip. That is a fine place for it to sit; it should not be described as the gate getting smarter.
- **Step 2 requires the ledger row and the tag to agree, and the gate must fail closed if it cannot read the ledger.** An unreadable or absent ledger must not silently satisfy step 2.

## Out of scope

- **The uninspectable-target case.** Neither half of this fixes a pin whose target the lint cannot resolve; that site stays frozen until target resolution improves. Do not close this issue claiming otherwise.
- **Re-litigating #885's verdicts.** Those 674 `boundary` rows were decided on recorded evidence of a consumer.
- **Retrofitting tags onto all 674 pins.** Even with this change, copying each ledger rationale into a source comment would create hundreds of uncoupled duplicates of a record that already has one canonical home. The point of step 2 is that a tag can *reference* the decision, not that every pin must carry one.

## Sequencing

Pairs naturally with #946, which will add ~45 fresh ledger judgments — exactly the population step 2 would then be able to surface. Either order works; doing this one first means #946's retained pins can carry their reason from the start.

A useful side effect worth stating plainly: this **unfreezes maintenance**. Today a stale comment on some retained pins cannot be corrected at all. Afterwards, a pin with a real ledger row can carry its reason where a reader will see it *and* be edited normally.

The01Geek:
Closes #948.

## The ladder as implemented

`lib/test/pin-corpus-lint.py`'s `scan_changed_sources` no longer answers a changed pin
site with a single prose test. Per site, in this order:

**Before the ladder (unchanged arms).** The retired-wording-literal revival block, then
`_typed_pin_inspection_error`, then — the one arm I moved — a declaration that is
**present but ungrammatical** (unknown category, empty rationale, duplicate marker). That
grammar check now runs *ahead* of the routing ladder so neither a program consumer nor a
ledger row can route around a malformed tag. The ~35 standing pins on a pre-vocabulary
free-text category therefore stay findings, exactly as today.

1. **A program demonstrably reads it → pass, no human needed.** The literal, or a
   machine-identifier-shaped token it names, occurs in the *operative* text of a tracked
   `scripts/**`, `lib/**`-outside-`lib/test/`, or `.github/**` file. No tag and no ledger
   row are required for this arm.
2. **Otherwise, the ledger already recorded the decision → pass.**
   `lib/test/pin-corpus-adjudications.tsv` carries this literal as `boundary` **and** the
   site carries a valid `# structural-pin-ok:` declaration. The tag is a *pointer to an
   authorized decision*, never a self-granted permission.
3. **Neither → the finding stands**, and its text now names which half of step 2 was
   missing.

`load_machine_consumer_sources` builds the step-1 corpus once per gate run: population
from an index-reading `git ls-files -- scripts/ lib/ .github/` (no `--others`, no
root-anchored recursive walk — issue #711), contents from the worktree, `lib/test/**`
excluded. For `#`-comment languages the comment regions are **subtracted**: a literal
quoted in a `lib/*.sh` comment is not a program reading it.

### The token-matching rule I chose, and why

A "distinctive token" is a whitespace-split word, trimmed of surrounding quoting and
sentence punctuation, that is **≥ 8 characters, contains ≥ 3 letters, and has a
machine-identifier shape**: contains `_`, contains `/`, contains `:`, starts with `--`,
has an interior `alnum.alnum`, or is kebab-case with a digit or ≥ 3 segments. Matching is
whole-token (neither neighbour may be `[\w-]`).

The shape requirement is the whole rule: a common English word is never distinctive
however long it is, so `configuration`, `documentation`, `verdict` and `INCONCLUSIVE`
cannot satisfy step 1. Two-segment kebab words that read as ordinary English
(`fail-closed`, `best-effort`) are excluded on purpose — a kebab token qualifies only with
a digit or a third segment. `DEVFLOW_BASH`, `scripts/workpad.py`, `--tick-progress`,
`devflow:workpad`, `config-get.sh`, `phase-4-verdict` and `mutation-routing-worktree` all
qualify. A literal with no such token yields no tokens at all, which routes its pin to
step 2 — never to a finding.

## The three constraints

- **Step 1 fails toward step 2, never toward rejection.** "No consumer found" is only ever
  the absence of *found* evidence — a helper that walks a routing table row by row names
  no individual row — so it hands the site to step 2. Every way step 1 can come up short
  (no corpus supplied, an untracked consumer, an unreadable or non-UTF-8 file, a
  comment-only mention, a literal with no distinctive token) routes down, and the
  unreadable-file case emits a stderr breadcrumb rather than failing silently.
- **Step 2 fails closed.** An absent row, a row in any other bucket, an empty ledger and
  an unestablished ledger are all findings. An *unreadable* ledger never reaches the
  ladder at all: `analyze_adjudication_changes` raises `InfrastructureError` first, so the
  public command exits **2**, not 0 — driven end-to-end below.
- **The gate routes; it does not judge.** Step 1 asks a mechanical question. Step 2 defers
  to an authorization granted elsewhere. The real control over step 2 is the **review of
  ledger changes**, which is separately delta-gated and needs an exact branch change
  manifest against the current `origin/main` tip. This is not the gate getting smarter,
  and the code comment and `CONTRIBUTING.md` both say so in those words.

## The deliberate decision: ledger row present, tag absent → **finding**

Step 2 requires *both* halves. Rationale, recorded in the test's own comment:

- It keeps this change a pure **loosening for tagged pins** and never a loosening for
  untagged ones — a tagless pin is a finding today and stays one, so nothing red today
  silently goes green except via a tag or a demonstrable consumer.
- The gate only asks about a site whose own lines changed. When a maintainer edits that
  line, the next reader of it is owed the reason *at the site*; a ledger-only pass would
  leave the gate with no on-site signal at all.
- It does not conflict with "don't retrofit tags onto the 674" — retrofitting is still not
  the ask, because the question is never posed until the line is edited.

## Scope decision surfaced by a pre-existing test

`test_authorized_retired_revival_cannot_launder_committed_prose_target` went red on the
first draft, and correctly. A **retired** wording literal's revival has its own stronger
contract (deliberate revival authorization **plus** a genuinely declared structural
boundary — `CONTRIBUTING.md`: "a new boundary row alone does not make the revival valid"),
and both ladder steps rest on exactly that boundary row. So the ladder is scoped to the
**retained** population: a retired literal keeps its pre-#948 behaviour verbatim, message
text included. That test passes unchanged, and a new unit test drives the scope directly.

## Which arm is covered by which test

All in `lib/test/test_pin_corpus_lint.py`. The new `PinRoutingLadder948Tests` fixture pins
a whitespace-bearing visible Markdown sentence — a literal the lint classifies as prose,
i.e. one for which *no* declaration and *no* ledger row was reachable before this change.
Each test asserts the deciding message, not just the count, because a reordered ladder
would otherwise stay green.

| Arm | Test |
|---|---|
| Step 1 passes with no tag and no ledger row (+ RED-first control: same site, no corpus → finding) | `test_step_one_passes_with_no_tag_and_no_ledger_row` |
| Step 1 accepts a distinctive token, not only the whole literal | `test_step_one_accepts_a_distinctive_token_not_only_the_whole_literal` |
| **Negative control** — an unrelated file mentioning common words does not pass step 1 | `test_step_one_ignores_a_common_word_in_an_unrelated_file` |
| A consumer whose only mention is a `#` comment does not pass step 1 | `test_step_one_ignores_a_mention_that_is_only_a_comment` |
| The corpus excludes `lib/test/`, `docs/`, `skills/` and repo-root prose | `test_step_one_corpus_excludes_the_suite_and_prose_surfaces` |
| The token rule itself (7 qualifying, 7 non-qualifying shapes) | `test_distinctive_token_shapes_are_machine_identifiers` |
| **Step 2 passes a prose-resolving pin** — the case impossible today | `test_step_two_passes_a_prose_pin_with_a_tag_and_a_boundary_row` |
| **Anti-self-grant** — tag, no ledger row → finding | `test_step_two_rejects_a_tag_with_no_ledger_row` |
| **Deliberate decision** — ledger row, no tag → finding | `test_step_two_rejects_a_ledger_row_with_no_tag` |
| A row present but not `boundary` (6 buckets) → finding | `test_step_two_rejects_a_row_in_any_other_bucket` |
| **Fail-closed** — unestablished and empty ledger → finding | `test_step_two_fails_closed_on_an_unestablished_ledger` |
| A different literal's boundary row does not carry over | `test_a_different_literals_boundary_row_does_not_carry_over` |
| **Positive control** — wording-only pin with nothing behind it → still a finding (prose and non-prose targets) | `test_a_wording_only_pin_with_nothing_behind_it_is_still_a_finding` |
| Invalid/missing category → finding *even with* a consumer and a boundary row (arm order) | `test_an_invalid_category_is_a_finding_ahead_of_the_whole_ladder` |
| **Arm order** — one site, 8 configurations, each asserting the deciding arm's message | `test_each_configuration_selects_its_documented_arm` |
| Retired literal keeps the pre-#948 revival contract (scope) | `test_a_retired_literal_keeps_the_pre_948_revival_contract` |
| **End-to-end step-1 wiring** through `git ls-files` + worktree reads: undeclared pin over a literal a tracked `scripts/` program reads → rc 0; comment-only mention → rc 3 | `StaticPinWorktreeCompositionTests.test_step_one_consumer_passes_the_public_worktree_ladder` |
| **End-to-end fail-closed** — unreadable ledger → rc 2, on a pin that exits 0 with a readable one | `StaticPinWorktreeCompositionTests.test_unreadable_ledger_is_infrastructure_not_a_step_two_pass` |

**Arm I could not test:** an end-to-end step-2 pass through the public command. Reaching it
needs a `boundary` row for the fixture literal *plus* an authorized delta manifest, and
adding a row without one is itself a finding (`unauthorized pin adjudication delta`) — so
the fixture would end up testing the bundle machinery, which has its own tests. Step 2 is
driven at the `scan_changed_sources` boundary instead, with the production wiring
(`current_adjudications=adjudication_analysis.current`) unchanged.

## Explicitly not done

- The ~35 pins whose tag carries an invalid category are **not** fixed. They become
  fixable after this lands; touching them here would balloon the diff.
- The uninspectable-target case (`typed structural declaration target cannot be
  inspected`) is **not** fixed. Neither half of the proposal addresses it, and the frozen
  site in `lib/test/modules/create-issue-contract.sh` stays frozen.
- No tags retrofitted onto the retained population; no existing `boundary` adjudication
  re-litigated.

## Coupled prose updated in the same change

- `CONTRIBUTING.md` — the paragraph stating that the gate "runs the prose-resolution check
  *before* the declaration check … **regardless of any declaration**" is now false; it
  states the ladder, where the real control sits, and both out-of-scope cases.
- `lib/test/run.sh` — the `#871` comment asserting that the prose arm is one "no
  `# structural-pin-ok:` declaration rescues" now records the step-2 pointer contract.

## Verification

Run on this branch, rebased onto current `main` (`0de0fbc1`, after #947 merged), every
script invoked as a direct leading token:

- `python3 lib/test/test_pin_corpus_lint.py` → `Ran 155 tests … OK`
- `python3 lib/test/pin-corpus-lint.py mutation-routing-worktree <repo-root>` → rc 0
- `python3 lib/test/test_pin_corpus_classifier.py` → `Ran 15 tests … OK`
- `python3 lib/test/coverage_map_guard.py .` → rc 0
- `git ls-files '*.py' | xargs -r ruff check` → `All checks passed!`
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` → rc 0
- the full `lib/test/run.sh` — result posted as a comment on this PR

Writing-skills evidence: not applicable — this change touches no `SKILL.md` and no prompt
surface.

### DIFF
--- a/.changeset/issue-948-pin-gate-routing-ladder.md
+++ b/.changeset/issue-948-pin-gate-routing-ladder.md
@@ -1,0 +1,22 @@
+---
+bump: patch
+---
+
+### Changed
+
+- The pin-corpus authoring gate now **routes** a changed pin site through an ordered
+  three-step ladder instead of returning on a single prose test (issue #948): (1) a
+  program in `scripts/**`, `lib/**` outside `lib/test/`, or `.github/**` demonstrably
+  reads the literal or a machine-identifier-shaped token it names — pass; (2) otherwise
+  the delta-gated ledger `lib/test/pin-corpus-adjudications.tsv` already records the
+  literal as `boundary` **and** the site carries a valid `# structural-pin-ok:`
+  declaration — pass, the marker acting as a pointer to that authorized decision; (3)
+  neither — the finding stands, naming which half was missing. Step 1 can only ever
+  route a site down to step 2, and step 2 fails closed: a tag with no ledger row, a
+  ledger row with no tag, a row in any other bucket, and an unestablished ledger are all
+  findings. The `# structural-pin-ok:` escape hatch, previously unreachable for any
+  literal the lint classified as prose, is usable again — so a retained pin with a
+  recorded boundary decision can carry its reason at the site and be edited normally.
+  A declaration whose grammar is invalid is still decided *before* the ladder, a retired
+  wording literal's revival keeps its stronger pre-existing contract, and an
+  unresolvable declaration target is still unfixable.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -170,21 +170,45 @@ an executable structural boundary and must carry
 
 **The marker is a declaration required of a NEW or CHANGED pin — it is not a retention
 badge you can retrofit onto the standing population (issue #885).** The gate scopes the
-requirement to a site whose every physical line is in the diff's added set, and it runs
-the prose-resolution check *before* the declaration check: a pin whose literal resolves
-into visible Markdown prose draws
-`literal resolves into prose at <target>:<line>` **regardless of any declaration**.
-A large majority of the retained pins resolve that way (a past-time snapshot, not a
-live figure: 229 of them measured on the post-sweep population at the #885 sweep
-commit, kept as provenance for the probe below and deliberately not re-rendered) — they
-are retained because a tool or consumer reads the *thing the literal names* (a marker, a
-grant-matched invocation shape, a schema field set), a distinction the lint's
-prose boundary cannot see. So adding a marker to one of those turns the required gate
-RED and its only gate-satisfying disposition would be deletion, which the recorded
-adjudication contradicts. The per-pin retention record therefore lives where it is
-delta-gated and auditable — `lib/test/pin-corpus-adjudications.tsv`, surfaced per row in
-the census — and not as several hundred uncoupled copies in source comments. Touch a
-retained pin's lines only when you are prepared to answer the gate for it.
+requirement to a site whose every physical line is in the diff's added set. For such a
+site — once its declaration *grammar* has been checked, which happens first and which no
+later arm routes around — the gate **routes** through an ordered three-step ladder
+(issue #948), and the routing is deliberately not a judgement:
+
+1. **A program demonstrably reads it.** The literal, or a machine-identifier-shaped
+   token it names, occurs in the operative (comment-stripped) text of a tracked
+   `scripts/**`, `lib/**`-outside-`lib/test/` or `.github/**` file. Pass; no
+   declaration and no ledger row needed. A grep-shaped search misses a *generic*
+   consumer by construction — a helper that walks a routing table row by row names no
+   individual row — so "found none" routes to step 2 and never to a finding.
+2. **The ledger already recorded the decision.** `lib/test/pin-corpus-adjudications.tsv`
+   carries this literal as `boundary` **and** the site carries a valid
+   `# structural-pin-ok:` declaration. The marker is a *pointer to an authorized
+   decision*, never a self-granted permission: a tag with no ledger row is a finding,
+   and so is a ledger row with no tag. Step 2 fails closed — an absent, unestablished
+   or non-`boundary` row never satisfies it, and a ledger the gate cannot read is an
+   infrastructure failure before the ladder runs at all.
+3. **Neither.** The finding stands: `literal resolves into prose at <target>:<line>`,
+   naming which half of step 2 was missing. This is the pin the policy exists to remove.
+
+Be honest about where the control sits: after step 2 the gate is only routing, and the
+real safeguard is the **review of ledger changes**, which is separately delta-gated and
+needs an exact branch change manifest. A large majority of the retained pins resolve
+into prose (a past-time snapshot, not a live figure: 229 of them measured on the
+post-sweep population at the #885 sweep commit, kept as provenance for the probe below
+and deliberately not re-rendered) — they are retained because a tool or consumer reads
+the *thing the literal names* (a marker, a grant-matched invocation shape, a schema
+field set), a distinction the lint's prose boundary cannot see; step 2 exists so that
+such a pin can carry its reason at the site *and* be edited normally, which before #948
+it could not. The per-pin retention record still lives where it is delta-gated and
+auditable — `lib/test/pin-corpus-adjudications.tsv`, surfaced per row in the census —
+and not as several hundred uncoupled copies in source comments, so retrofitting a marker
+onto the standing population is still not the ask. Two cases the ladder does **not**
+reach: a pin whose declared target the lint cannot resolve at all
+(`typed structural declaration target cannot be inspected` — e.g. a concatenated
+in-module bundle) is still unfixable, and a *retired* wording literal's revival keeps its
+own stronger contract below. Touch a retained pin's lines only when you are prepared to
+answer the gate for it.
 
 `lib/test/pin-corpus-adjudications.tsv` contains only the current active adjudication
 state. Every addition, removal, or change to that table must be authorized by an exact
--- a/lib/test/pin-corpus-lint.py
+++ b/lib/test/pin-corpus-lint.py
@@ -63,6 +63,27 @@ authoring time:
   scan exits 0. The lower-level ``mutation-routing`` synthetic-fixture command
   remains for legacy self-tests.
 
+  **The static classifier ROUTES; it does not judge (issue #948).** For a changed
+  site whose declaration grammar is valid, it walks an ordered three-step ladder:
+  (1) a program in ``scripts/**``, ``lib/**`` (non-test) or ``.github/**``
+  demonstrably reads the literal or a distinctive token it names — pass, no human
+  needed; (2) otherwise, the delta-gated ledger
+  ``lib/test/pin-corpus-adjudications.tsv`` already records this literal as
+  ``boundary`` **and** the site carries a valid ``# structural-pin-ok:``
+  declaration — pass, honouring the tag as a *pointer to an authorized decision*;
+  (3) neither — report the finding. Step 1 can only ever route a site to step 2
+  (a grep-shaped consumer search misses a generic consumer by construction, so
+  "found none" means "ask the ledger", never "reject"), and step 2 fails closed:
+  an absent, unestablished or non-``boundary`` ledger row never satisfies it, and
+  an unreadable ledger is an infrastructure failure long before this ladder runs.
+  The real control over step 2 is therefore not this classifier but the review of
+  ledger changes, which is separately delta-gated and needs an exact branch
+  manifest — do not read the ladder as the gate getting smarter. The ladder is
+  scoped to the RETAINED population: a *retired* wording literal's revival keeps
+  its pre-#948 contract (deliberate authorization plus a genuinely machine-shaped
+  target), since both ladder steps rest on the very boundary row that contract
+  says cannot on its own make a revival valid.
+
 **Fail-closed:** a call site the scanner cannot resolve statically (the literal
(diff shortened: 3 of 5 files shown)


The01Geek:
## Verification — all on the final tree (`e4dbce31`, clean worktree, rebased onto `main` at `0de0fbc1`)

Every script invoked as a direct leading token; the full suite launched through a python
wrapper with `start_new_session=True` and SIGINT restored to `SIG_DFL`, never backgrounded
with `&`.

```
$ python3 lib/test/test_pin_corpus_lint.py
Ran 155 tests in 182.502s

OK

$ python3 lib/test/pin-corpus-lint.py mutation-routing-worktree <repo-root>
rc=0

$ python3 lib/test/test_pin_corpus_classifier.py
Ran 15 tests in 3.635s

OK

$ python3 lib/test/coverage_map_guard.py .
coverage_map_guard rc=0

$ git ls-files '*.py' | xargs -r ruff check
All checks passed!
ruff rc=0

$ shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh
rc=0     (ShellCheck 0.11.0)

$ lib/test/run.sh
13949 passed, 0 failed
```

The suite summary carries **no skip clause**, so the skip tally is zero — `0 failed` *and*
`0 skipped`, both required. The total matches the expected count for the current `main`
exactly, and the log ends with its own summary line (checked, because a truncated log
cannot be read as `0 failed`); an independent scan of the log finds zero `FAIL` and zero
`NOTE` lines.

CI on this same sha: `lib + python tests` **success**, all four shards **success**, `lint
(shellcheck + actionlint + ruff)` **success**.

Issue #946, opened 2026-07-30
Bring review-and-fix-contract's wrapper-routed pins into the pin corpus (prerequisite deferred from #885)


The01Geek:
Deferred from #885 (PRs #942 + #944), which delivered the re-adjudication and swept 194 of the 239 prose-bucketed pins. This is the remaining 45, plus the prerequisite that currently blocks them.

## The blocker

`lib/test/modules/review-and-fix-contract.sh` routes its pins through a **module-private wrapper**:

```
_raf_pin_unique() { assert_eq "$1" "1" "$(_raf_pin_count "$2" "$3")"; }
```

`pin-corpus-classifier.py`'s `EXISTENCE_HELPERS` does not know that wrapper, and the file is absent from `PIN_CORPUS_SOURCES` (which lists `lib/test/run.sh` plus six modules). So **none of its pins has a census row**, and `CONTRIBUTING.md`'s retirement gate takes **arm 0** — the census cannot answer, so every one of them retains. #885's AC8 expected this module's count to fall materially; it could not, and the reason is a fail-closed authorization rule rather than effort.

## Measured scope (at 42dce581)

| Quantity | Value |
|---|---|
| `_raf_pin_unique` call sites in `review-and-fix-contract.sh` | 45 |
| That module's `minimum_assertions` (exact-pinned tally) | 67 |
| Audited modules outside `PIN_CORPUS_SOURCES` carrying **zero** pins | 5 of 6 |
| Standing corpus after #885 | 675 sites, all `boundary` |

Worth stating because it bounds the work: this module is the **entire** invisible population. The other five audited-but-uncensused modules (`experiment-records`, `harness-python-guards`, `prompt-extension-reader`, `retrospective-lifecycle`, `workflow-flight-recorder`) carry no pins at all, so there is no second instance of this gap to find later.

## Suggested shape

1. **Teach the corpus the wrapper.** Add `lib/test/modules/review-and-fix-contract.sh` to `PIN_CORPUS_SOURCES` and make the classifier resolve `_raf_pin_unique` (`pin-corpus-lint.py` already discovers first-positional-forwarding wrappers to a fixpoint for `coverage_map_guard`'s label derivation — reuse that rather than special-casing the name). Expect the census to grow by ~45 rows, each initially `unclear` or mechanically bucketed, and **every newly-`unclear` site needs a table row or the classifier hard-errors** (`missing adjudication for <key>`).
[...]

Issue #843, opened 2026-07-26
Refresh the frozen pin identities so the pin corpus can track a renamed cap and cover issue-797's new operative sentences


The01Geek:
## Dependencies
Unblocked — #797 landed (PR #837, merged 2026-07-26), so the parent's Phase 4.1.5 behavior-inert prose cap is on main and this deferred chunk can start.

## Problem Statement
Follow-up to #797. Issue #797 widened the review engine's Phase 4.1.5 over-grade cap into a behavior-inert prose cap. Two consequences could not be discharged inside that PR, both caused by the pin corpus being frozen against changed identities. The maintainer is left with a suite that prints a test name describing a cap the engine no longer has, and with a merge gate whose most consequential new prose is unguarded.

**Correction, filed after PR #836 landed on main.** This issue was filed while the mutation-pin machinery still existed. PR #836 has since **retired** it: the issue-#810 gate now requires the audited-source census and its checked-in inventory to remain **empty**, `assert_pin_red_under` and its wrappers are gone from `lib/test/run.sh`, and `CLAUDE.md` now states that behavioral regressions require ordinary executable tests exercising the relevant interface or contract. So the second half of this issue is **not** "admit a new mutation-pin identity" — that route is closed by design. It is "decide what coverage, if any, an unpinnable prompt-prose rule can carry under the post-#836 regime."

## Current Behavior
`.devflow/logs/residual-prose-retirement-manifest.tsv` freezes each retained pin's identity — source, helper, assertion **name**, literal, and target — at a base revision, and `lib/test/test_residual_prose_retirement_manifest.py::test_current_tree_realizes_the_retirement_manifest` requires every one of them to still exist verbatim in the tree.

Two residues follow in the shipped tree:

- `lib/test/run.sh` carries `assert_pin_unique "291(AC1): review 4.1.5 carries the deterministic in-code-comment cap"`. Section 4.1.5 no longer carries an in-code-comment cap; it carries the behavior-inert prose cap. Every sibling label in the same block was renamed by PR #837, and this one could not be, because the name is a frozen identity column. PR #837 disclosed the constraint in an adjacent comment rather than leaving the staleness unattributed.
[...]

Issue #366, opened 2026-07-09
Guard /devflow:implement against nested-Skill tail-call early-stop: add a Skill-completion re-anchor and make the Skill rule exclusionary


The01Geek:
## Problem Statement

A `/devflow:implement` run stopped mid–Phase 3.3 without finalizing, leaving the workpad frozen at `Status: 🚀 Reviewing` and requiring a manual resume. The cause is structural: invoking a nested `Skill` in the middle of a phase behaves as a **tail call**, not a subroutine call. When the nested skill's body ends in an interactive "ask the user / apply with approval" step, that step becomes the *run's* terminal step, and nothing restores the interrupted phase continuation. This maintainer has observed early-stop runs "a few times recently"; this issue pins down and fixes the `Skill`-return variant of it.

DevFlow runs across multiple coding agents (Claude Code, GitHub Copilot CLI, Codex CLI, Cursor, Gemini CLI) and operating systems. The fix must work on all of them and must not depend on any single harness's features.

## Current Behavior

Concretely, the following sequence terminates the run mid-phase:

1. Inside Phase 3.3's fix loop, the orchestrator needed to edit `CLAUDE.md`. `CLAUDE.md`'s own Conventions section mandates (line 58): *"Updating `CLAUDE.md`? Use `/claude-md-management:revise-claude-md` … never hand-edit without running at least one."*
2. But `skills/implement/SKILL.md`'s **Skill rule** (line 12) sanctions only three skills: *"Use the Skill tool for `simplify` and `review-and-fix` during code review and `pr-description` for PR documentation."*
3. A Phase-3 fix that must touch `CLAUDE.md` therefore has **no legal path**: hand-editing violates `CLAUDE.md`, and invoking the mandated skill leaves the sanctioned set.
4. The orchestrator invoked `claude-md-management:revise-claude-md`. That skill's body is written for interactive use and terminates at *"Step 5: Apply with Approval — Ask if the user wants to apply the changes."* The nested skill's body arrives as a *user* turn, so its terminal step became the *run's* terminal step. The Phase 3.3 continuation was never restored, and the run ended with `Status` still `🚀 Reviewing`.

Why the existing guards did not catch it:

[...]

PR #961, opened 2026-07-30
test(pins): re-adjudicate the 44 wrapper-routed review-and-fix pins (#946 step 2)

The01Geek:
Step 2 of #946. Step 1 (#949) taught the pin corpus `review-and-fix-contract.sh`'s module-private `_raf_pin_unique` wrapper and gave its 44 pins census rows, all adjudicated `boundary` — the fail-closed default for an unanswered question, not a finding that they are load-bearing. This PR answers the #843 governing question per site.

**No pin is removed here.** `minimum_assertions` stays 67 and no pin's own line is touched. The step-3 sweep is a separate PR, because the delta gate nets merge-base -> HEAD and a row that moves to a prose bucket and is then deleted with its pin cancels out.

## Verdict split (44 sites)

| Outcome | Count |
|---|---|
| **Retirable** (`prose-sole-copy`) | **28** |
| Retained — verified consumer | 1 |
| Retained — arm-1 multi-home (`counted_occurrences >= 2`) | 3 |
| Retained — wrapped second home | 1 |
| Retained — cross-file phase contract | 1 |
| Retained — extraction-inventory record, not prompt prose (`counted_occurrences 0`) | 4 |
| Retained — adjudicated on evidence by #885, not re-litigated (a stated non-goal) | 6 |

The 28 retirable sites are the guardrail, convergence, fix-delta-gate, verification, telemetry-bypass, continuation and prompt-composition sentences in `skills/review-and-fix/references/{loop-control,fixing,loop-exit,shadow-review,fix-delta-gate,error-handling}.md` and the three guard-class literals in `.devflow/prompt-extensions/review-and-fix.md`.

## Consumer hunt

Every site got a real hunt, not a pattern-match: the literal read in its home with surrounding context, then a search for a consumer across `scripts/`, non-test `lib/`, `.github/`, `agents/`, `skills/`, plus the generic reconcilers that can consume a literal without containing it. `scripts/devflow-cloud-writer-contract.json` was excluded throughout (it SHA-256-pins whole files; counting it would void the policy). Near-misses checked and ruled out:

- `lib/efficiency-trace.sh` parses exactly one `fixing.md` literal — the `FIX_COMMIT_SUBJECT_PREFIX` commit-subject template — and otherwise reads iteration-record JSON *field names* and a closed provenance-value set, never a prose sentinel. This is what clears the highest-risk candidate, the `park-calibration gate clean: no parked finding matched` sentinel.
- `lib/test/lint-subagent-extension-handoff.py` splits these files at markdown headings but keys heading text solely off `declared_non_dispatch`, which is empty; the `shadow-review.md` registry record carries `handoff: "inherited"`, which `audit()` exempts. Probed empirically: `scan_file` returns the same result with the two pinned `####` headings deleted.
- `extract-command-heads.py` / `extract-command-shapes.py` are fence-scoped. `.devflow/prompt-extensions/review-and-fix.md` carries **zero** fenced blocks and is in no scanned population, so the `type <fn> >/dev/null 2>&1` recipe is never reconciled against `lib/capability-profiles.json`. Mutation-probed on `loop-control.md`: mutating all three `max_iterations` recipe literals leaves both extractors' output byte-identical.
- `scripts/load-prompt-extension.sh` reads the extension as whole bytes (`--section` is used only by create-issue); `scripts/workflow-flight-recorder-registry.json` references these files by whole-file glob. Neither is a consumer of a literal inside them.

## Adversarial refutation pass

Two retirable verdicts were refuted and moved back to retained:

- **`keeps the full roster regardless of \`iterations\``** — this is the module's only pin targeting the review-*engine* bundle rather than the review-and-fix bundle. The rule is owned by `skills/review/phases/phase-3-agents.md` while review-and-fix's recovery path is the surface consuming its consequence: the #291-class cross-file phase contract CONTRIBUTING records under arm 3. Retained on that ground, not on a machine reader.
- **`Topic-priming is a second, distinct leak channel`** — `docs/shadow-review.md` carries the same sentence differing by one word (`a second, distinct channel`), and `skills/review-and-fix/references/shadow-review.md` states it a third time. That is a duplicated statement whose divergence check the pin buys, so arm 1 governs and a pin-only removal is not an accepted disposition.

One refutation was itself refuted, keeping the site retirable: a claimed verbatim second home for `the behavior you broke and the executable test you observed go RED` in `.devflow/prompt-extensions/implement.md` did not survive an exact-string grep — the extension's sentence is a reword this substring pin never held.

A calibration note worth recording: a first pass over-applied "wrapped second home" to any `docs/` paragraph restating the same *rule* in different words. Checked against #885's recorded standard (its rationales read "only soft match elsewhere is an unrelated sentence, not the same content"), a paraphrase at a different level of abstraction is not a wrapped home — the wrapped check is for the same literal carried as adjacent/wrapped fragments. Applying #885's standard is what keeps the two passes comparable.

## Census

719 rows: **691 `boundary`, 28 `prose-sole-copy`** (was 719 / 719 boundary / 0). Regenerated under the two-commit inventory-free protocol against the re-adjudication commit, after the final rebase onto `origin/main`.

## Coupled prose

`CONTRIBUTING.md`'s "Current state — arm 2 ... population is empty again" section and the two `CLAUDE.md` sentences asserting a boundary-only census and an arm-0 blind spot over `_raf_pin_unique` all became false with this change and are corrected in the same commit. The `CLAUDE.md` edits are the required-edit carve-out (#366), made directly and recorded here.

Closes nothing on its own — #946 remains open for step 3.


The01Geek:
## Review finding addressed — the Major is fixed, and one review note was wrong

**The finding is correct and is fixed as described.** Two `prose-sole-copy` rationales asserted a categorical no-reference claim that `scripts/devflow-cloud-writer-contract.json` contradicts (it whole-file SHA-256-pins both `error-handling.md` at :38 and `loop-exit.md` at :42). The bucket is unchanged — a whole-file SHA is an identity/change detector, not a semantic reader, and counting it as a consumer would make every prose pin retain and void the policy — but the rationale is the record step 3 relies on, and "nothing references it at all" would walk a maintainer straight into a RED `lib + python tests` check.

Both strings now state the accurate form **and** name the obligation explicitly, e.g.:

> Sole home; no tool parses this sentence. The file is whole-file SHA-pinned by `scripts/devflow-cloud-writer-contract.json`, which is an identity/change detector rather than a semantic reader, so it is not a consumer of this literal — but editing the file obliges re-running `python3 lib/test/cloud_writer_contract.py generate` in the same change or the required `lib + python tests` check goes RED.

The delta bundle's `current_state` for both keys is updated in lockstep (still 34 rows, 2 changed).

## Correction to the review: census regeneration WAS required

The review noted "no census regeneration needed". That was checked rather than trusted, and it is wrong. `pin-corpus-classifier.py` resolves the adjudications table **at the recorded `# revision:`**, not from the working tree, so regenerating against the prior revision reproduced the superseded string verbatim while the corrected table sat in the tree:

```
$ grep -c 'nothing in lib/, scripts/, or .github/ references error-handling.md at all' .devflow/logs/pin-corpus-inventory.tsv
1
```

The census is therefore regenerated under the two-commit inventory-free protocol against the correction commit (`3f8c58cf`). It now carries the corrected strings (`grep -c 'identity/change detector rather than a semantic reader'` -> `2`, old string -> `0`) with bucket totals unchanged: **719 rows, 691 `boundary`, 28 `prose-sole-copy`**.

Worth recording for the next pass: the frozen-census tests do not catch this. `test_frozen_inventory_matches_its_recorded_revision` asserts structure, row count and metadata, not rationale bytes, so a table-only rationale edit ships a census disagreeing with its own source and every gate stays green. That is why this was verified empirically rather than inferred from a green suite.

## Re-verification on the corrected head

```
$ python3 lib/test/pin-corpus-lint.py mutation-routing-worktree "$(git rev-parse --show-toplevel)"
(no output)  RC=0

$ python3 lib/test/test_pin_corpus_classifier.py                  Ran 17 tests      OK
$ python3 lib/test/test_residual_prose_retirement_manifest.py      Ran 13 tests      OK
$ python3 lib/test/test_red_on_removal_retirement_manifest.py      Ran 2 tests       OK
$ python3 lib/test/test_pin_corpus_lint.py                         Ran 173 tests     OK
$ python3 lib/test/coverage_map_guard.py .                         rc 0, no output
$ lib/test/run-module.sh review-and-fix-contract                   67 passed, 0 failed

$ lib/test/run.sh
14092 passed, 0 failed        # terminal line of the log; 0 lines matching '^  FAIL'; no skip clause
```

CI on the corrected head, all green: `lib + python tests`, `lint (shellcheck + actionlint + ruff)`, `shard (monolith)`, `shard (modules-pin)`, `shard (modules-large)`, `shard (modules-rest)`.

`merge-base(origin/main, HEAD) == origin/main` tip `2a8a79c8`.

On the advisory: the `CLAUDE.md` edits are the required-edit carve-out (#366) — three sentences that became factually false with this change (boundary-only census; the `_raf_pin_unique` arm-0 blind spot) — made directly and recorded in the PR body.

Issue #375, opened 2026-07-09
Wave 1 — harden lib/test/run.sh pin tooling: mutation-taking assert_pin_red_under, pin-in-comment lint, wrapped-literal meta-guard, and an evidence-based behavioral-fix-pin rule


The01Geek:
## Problem Statement

`lib/test/run.sh` pins are the suite's mechanism for keeping agent-executed prose contracts alive, and two retrospective parents independently found the pin *tooling* — not author discipline — to be the failure point. On PR #364 (issue #370's evidence), five defects traced to pins that could not fail on the regression they guard: the framework's only removal-proof assertion, `assert_pin_red_on_removal`, reports `PASS->FAIL` for **any** present-and-unique literal, so a pin on a motivating header "proves" exactly as much as a pin on the operative mechanism. On PR #340 (issue #371's evidence), a coupled-site sweep and two source pins all missed a contract sentence because it lives in wrapped adjacent Python string literals — no single source line contains it, so line-based `git grep` finds nothing. Iteration cost attributed by the parents: the framing-pin class consumed fix iterations on PR #364, and the wrapped-literal class consumed PR #340's iteration 1 entirely.

This ticket is Wave 1 of the consolidated retrospective plan: it lands the mutation-taking pin primitive and the two mechanical guards that later waves' new pins are required to use.

## Current Behavior

**The removal helper strips whole lines, so it cannot distinguish operative from framing.** Verified at HEAD, `lib/test/run.sh` lines 1717–1724:

```bash
assert_pin_red_on_removal() {  # name literal [file]   (file defaults to $MAXI_SKILL)
  local t file="${3:-$MAXI_SKILL}" before after; t="$(probe_tmp "$1 (removal setup)")" || return 0
  before="$(probe_assert assert_pin_unique 'probe-present' "$2" "$file")"
  grep -vF "$2" "$file" > "$t"
  after="$(probe_assert assert_pin_unique 'probe-removal' "$2" "$t")"
  assert_eq "$1" "PASS->FAIL" "$before->$after"
  rm -f "$t"
}
```

The `grep -vF` on line 1720 deletes every whole line containing the literal; the post-strip re-check then necessarily reports FAIL for any literal that was present and unique. The helper proves "the pin notices its own line vanishing" — never "the pin notices the guarded behavior regressing."

**Issue #370 Recommendation 3, verbatim:**

[...]

PR #969, opened 2026-07-30
test(pins): derive dispatch rosters instead of pinning unresolvable literals

The01Geek:
Unblocks PR #965 (the plugin rename) from `main`, so it can be brought current and go green.

## The defect (pre-existing on `main`, not introduced by the rename)

Four guard sites in `lib/test/run.sh` grep a fixed dispatch literal out of a runtime-concatenated skill bundle:

| line on `main` | literal | haystack |
| --- | --- | --- |
| 28387 | `subagent_type: devflow:$fdagent` | `$IMPL_SKILL_BUNDLE` |
| 28570 | `**devflow:$a**` | `$REVIEW_BUNDLE` |
| 28577 | `devflow:$a` | `$MAXI_BUNDLE` |
| 28815 | `/devflow:requesting-code-review` | `$REVIEW_BUNDLE` |

Three interpolate a loop variable into the pinned literal, and all four name a `mktemp`'d bundle path. `lib/test/pin-corpus-lint.py`'s static classifier can resolve neither, so **any** edit to one of those lines is an unanswerable finding. Verified on `main` by touching only those four lines (no rename involved):

```
$ python3 lib/test/pin-corpus-lint.py mutation-routing-worktree .
MUTATION-ROUTING	lib/test/run.sh:28386	raw-presence	<unresolved-literal>	missing structural declaration
MUTATION-ROUTING	lib/test/run.sh:28569	raw-presence	<unresolved-literal>	missing structural declaration
MUTATION-ROUTING	lib/test/run.sh:28576	raw-presence	<unresolved-literal>	missing structural declaration
MUTATION-ROUTING	lib/test/run.sh:28814	raw-presence	/devflow:requesting-code-review	missing structural declaration
EXIT=3
```

…and adding a `# structural-pin-ok:` declaration is correctly refused, because the inspectability it asserts does not exist:

```
MUTATION-ROUTING	lib/test/run.sh:28386	raw-presence	<unresolved-literal>	typed structural declaration literal cannot be inspected
MUTATION-ROUTING	lib/test/run.sh:28569	raw-presence	<unresolved-literal>	typed structural declaration literal cannot be inspected
MUTATION-ROUTING	lib/test/run.sh:28576	raw-presence	<unresolved-literal>	typed structural declaration literal cannot be inspected
MUTATION-ROUTING	lib/test/run.sh:28814	raw-presence	/devflow:requesting-code-review	typed structural declaration target cannot be inspected
```

`git diff origin/main pr/965 -- lib/test/run.sh` shows those lines change **only** `devflow:` → `prflow:`. So this is a latent `main` defect that a skipped gate concealed, and it belongs here rather than in #965.

## Option chosen, and why the two offered options do not work

Note the failure mode above: **three of four fail on the LITERAL, not the target.** So "teach the gate to resolve bundle members" (option 1) fixes exactly one of the four — the other three are still `<unresolved-literal>` and never reach the target check. Option 2 (convert to the helper pin form) routes through the same `resolve_arg`/`path_vars` machinery and fails identically. Neither offered option unblocks the set.

This PR takes the third route: **remove the unresolvability at the source** by replacing four fixed-wording source-presence pins with one executable derivation each. `dispatched_roster_missing` extracts the dispatch identifiers a bundle actually carries and reports which roster members are **missing** by name. That is what `CLAUDE.md` #375/#666/#810 prescribe — guard executable behavior, not prose presence — so the gate is satisfied honestly rather than by a declaration claiming something untrue. **`pin-corpus-lint.py`'s policy is unchanged**: nothing is loosened, and no new declaration is added.

It fails **closed** by construction: an empty extraction (renamed namespace, unreadable bundle, absent `grep`) reports *every* roster member missing, so a derivation that quietly matches nothing is RED, not green. It is deliberately a membership test rather than set equality, so adding a sixth dispatched agent — guarded by that agent's own rows — does not turn an unrelated PR red.

## Demonstration: the guards still bite

Planted two dropped dispatches (`**devflow:type-design-analyzer**` → `**type-design-analyzer**` in `skills/review/phases/phase-3-agents.md`; `subagent_type: devflow:code-architect` → `subagent_type: code-architect` in `skills/implement/phases/phase-2-implement.md`), then ran the full suite:

```
  FAIL  #139 implement bundle dispatches both feature-dev agents (derived subagent_type roster; names any missing id)
         expected:
         actual:   code-architect
  FAIL  #141 review engine dispatches every roster agent via its **devflow:<name>** prompt block (derived; names any missing id)
         expected:
         actual:   type-design-analyzer

14174 passed, 9 failed, 1 skipped
```

Both rows fired and **named the dropped identifier** — better diagnosis than the `expected yes / actual no` the greps produced. (The other 7 failures are the expected collateral of mutating SHA-256-pinned skill assets: the cloud-writer runtime manifest. The 1 skip is the expected dirty-tree `#434`.) Restored, and the suite is green again — summary below.

## The deeper finding: the gate could be skipped silently

`require_current_adjudication_base` raises **before** both `scan_changed_sources` passes. That is the correct fail-closed direction, but at the caller it is indistinguishable from any other rc 2 — which is why this gate went un-run for two whole sessions while four unanswerable sites sat unreported.

`scan_static_pin_changes` now writes a completion sentinel to stderr **only** on the path where both passes finished, and `lib/test/run.sh` asserts it. Its absence is now a named failure instead of an anonymous infrastructure exit. Demonstrated against the real precondition (a branch carrying an adjudication-table delta on a non-current base):

```
MUTATION-ROUTING-INFRASTRUCTURE	adjudication branch is not based on current origin/main: merge-base 0b126486…, base tip 6ce9ac33…
--- rc=2 ---
run.sh assertion '#967 ... actually ran' would read: expected yes / actual no
```

…and on a completed run:

```
MUTATION-ROUTING-STATIC-SCAN-COMPLETED
EXIT=0
```

`lib/test/test_pin_corpus_lint.py` drives both directions, and `_public_rc` now asserts the sentinel on every one of its ~25 callers (present when the run completes, absent on rc 2).

## Prediction for #965 (reasoning, not a guarantee)

**I expect #965 to go green on the `#810` gate once it is brought current onto this branch.** Measured, not assumed: I merged this branch into #965's head, resolved the three `run.sh` conflicts the way its author would (take the derivation side, apply the rename inside it), and ran the real gate on the result:

```
--- merge-base with origin/main: c6951f9eb43a7d5f591f6db1beb25d98c0a5de08 ---
MUTATION-ROUTING-STATIC-SCAN-COMPLETED
EXIT=0
```

The mechanism: the three loop-variable greps and the fourth fixed-literal grep no longer exist on either side, so the rename cannot touch them. It touches the new *call sites* instead — and those carry no `grep` at all, so `extract_guard_sites` yields no site for them.

Residual risks I cannot rule out: #965 must still resolve its own unrelated conflicts (`plugin.json`, `install.sh`) and re-run its generators; and this says nothing about any non-`#810` failure #965 may carry.

## Changeset

**None, deliberately.** Every changed path is under `lib/test/` — the suite's own tree. `install.sh` does not ship `lib/test/`, so nothing reaches a consumer repo as an update; `.devflow/prompt-extensions/implement.md` puts internal-only test changes outside the changeset rule.

## Verification

- Full suite, direct leading token, foreground — summary in the PR comment below.
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` (ShellCheck 0.11.0) — rc 0
- `git ls-files '*.sh' | grep -v '^lib/test/' | xargs -r shellcheck --severity=warning -e SC1091` — rc 0
- `git ls-files '*.py' | xargs -r ruff check` — All checks passed
- `python3 lib/test/regenerate-artifacts.py` — all artifacts reconciled, exit 0 (covers `cloud-writer-manifest`, `capability-profile-literals`, `plugin-identity-regions`, `coverage-map-ratchet`)
- `python3 lib/test/test_pin_corpus_lint.py` — 174 tests, OK

Writing-skills evidence: not applicable — no prompt surface (`skills/**`, `agents/**`, `.devflow/prompt-extensions/**`) is touched by this PR.


The01Geek:
Verification — full local suite, direct leading token, foreground, on `21fb3b13` (current with `origin/main` @ `c6951f9e`):

```
14204 passed, 0 failed
```

Zero failed, zero skipped (the summary is the byte-identical no-skip form; no `Failure recap` block, no `SKIP` lines).

Rows introduced or reworked by this change, all PASS in that run:

```
  PASS  #139 implement bundle dispatches both feature-dev agents (derived subagent_type roster; names any missing id)
  PASS  #141 review engine dispatches every roster agent via its **devflow:<name>** prompt block (derived; names any missing id)
  PASS  #141 fix-loop skill references every roster agent (review-and-fix roster; names any missing id)
  PASS  #142 review engine dispatches /devflow:requesting-code-review (derived final-pass call-site; names it when missing)
  PASS  #967 mutation-routing static pin classifier actually ran (a precondition raise cannot silently skip it)
```

PR #980, opened 2026-07-31
test: widen the self-trigger-loop guards to every declared command namespace

The01Geek:
Two self-trigger-loop guards were vacuous against the namespace they most need to cover.

## The defect

Both guards assert that a bot notice carries no trigger phrase, so the comment cannot re-fire its own workflow. Under the App token these notices fire a **real** `issue_comment` event (`GITHUB_TOKEN` comments are suppressed by GitHub), so a trigger substring in the body is an actual loop, not a theoretical one.

Both spelled exactly one namespace:

| Guard | Assertion |
|---|---|
| `lib/test/run.sh` | *suppression notice contains no `/devflow:` phrase in its NOTE body* |
| `lib/test/modules/review-trigger-helpers.sh` | *duplicate notice contains no `/devflow:` phrase* |

But the workflows declare **two** namespaces. `devflow.yml` carries triggers in both (3 each of `/prflow:review`, `/prflow:pr-description`, `/prflow:implement`, plus the `/devflow:` twins) and `devflow-implement.yml` fires on `contains(…, '/prflow:implement') || contains(…, '/devflow:implement')`. A `/prflow:` phrase leaking into either notice loops exactly as a `/devflow:` one would — and neither guard would see it.

## Proof that the gap was real, not theoretical

Planted `/prflow:implement` into `devflow-implement.yml`'s duplicate-notice `NOTE=` body and ran the real module:

```
FAIL  di: duplicate notice contains no /prflow: phrase
Module review-trigger-helpers: 415 passed, 1 failed
```

Only the new row failed. **The pre-existing `/devflow:` row stayed green on that exact leak** — that is the vacuity, demonstrated rather than argued.

Same for the `run.sh` guard, driving the identical code path (the real `mint_blk` from `lib/test/module-harness.sh`, the same namespace derivation, the same assertion expression) with `/prflow:review` planted into `devflow.yml`'s suppression notice:

```
derived namespaces: prflow devflow
non-vacuity (captured notice body): 1 (expect 1)
guard  /prflow:  -> FAIL (found 1)
guard  /devflow:  -> PASS      <-- the only guard that existed before
```

and after restoring the file:

```
guard  /prflow:  -> PASS
guard  /devflow:  -> PASS
```

## The fix

Both guards now scan **every** namespace `lib/plugin_identity.py --plugin-names` declares (currently `prflow`, `devflow`). This follows the idiom the partition-invariant block above already established, and its stated reasoning applies verbatim:

> The scanned namespace set is DERIVED from the declared plugin identity, never hardcoded: a guard spelling one namespace gives zero assurance for the other, and goes fully vacuous the moment the transitional namespace is retired.

So the set follows a namespace addition automatically, and retiring the transitional namespace narrows it correctly instead of leaving a dead check.

The module builds its set with **bash builtins** (`while IFS= read -r` + `case`), not `tr`/`cut`: the value SELECTS what gets scanned, and a missing non-preflight PATH tool would silently empty it and pass by inspecting nothing — the exact failure class these guards exist to catch. It then asserts the set is non-empty before using it.

## Anti-vacuity

Each guard gains a **planted-defect control per namespace**, matching the `CONTROL A/B/C` idiom already in the trigger-gate block. Each control first asserts the planted mutation really changed the notice body, then asserts the check reports a hit on it — so a future edit that makes the expression unable to match turns the suite RED instead of passing by matching nothing. The existing `no vacuous pass` capture assertions are retained unchanged.

## Notes

- No pin literal is added, changed, or re-spelled, so the `#810` mutation-routing gate is untouched. Verified after merging the post-#968 census.
- `origin/main` merged in (not rebased) at `dca492ac`; clean, no conflicts.
- No changeset: this is test-only, which `.changeset/README.md` classifies as an internal-only change.

## Verification

Full suite via the direct leading token `lib/test/run.sh`, run with the sanctioned `preexec_fn=SIG_DFL` + `start_new_session=True` launcher (it exceeds the 600s foreground ceiling; plain backgrounding leaves `SIGINT` at `SIG_IGN` and produces spurious signal-trap failures). Both `shellcheck` invocations, `ruff check`, and the generator `--check`s. Summaries in the thread.


The01Geek:
Follow-up commit: **assertion-floor reconciliation.**

CI's `shard (monolith)` caught what my first push missed — the 6 added assertions move this module from 410 to 416, and `test_module_runner.py`'s tranche test enforces the registry value as an **equality**, not a floor:

```
AssertionError: 'Module review-trigger-helpers: 410 passed, 0 failed' not found in [...]
```

Three coupled sites carry that count, all reconciled in the same commit:

| Site | Change |
|---|---|
| `scripts/workflow-flight-recorder-registry.json` → `minimum_assertions` | 410 → 416 |
| `lib/test/run.sh` call-site floor literal | 410 → 416 |
| `lib/test/run.sh` comment transcribing the floor | reworded |

The third is the PR-#553 rot class: a comment citing an exact count of a same-file thing drifts on the very edit that changes it. Rather than just bumping the number I reworded it to describe both raises and to **name the registry as the coupled site**, so the next person changing the tally is told where the other half lives.

`test_issue_746_tranche_modules_run_green_through_the_real_runner` now passes locally. Full-suite re-run in progress.

The01Geek:
**Verification**

Full suite, direct leading token `lib/test/run.sh`, sanctioned `preexec_fn=SIG_DFL` + `start_new_session=True` launcher:

```
14249 passed, 0 failed
RC=0
```

**0 failed, 0 skipped.** `origin/main` @ `dca492ac` merged in (not rebased) before this run, so the post-#968 pin census is what was verified.

| Gate | Result |
|---|---|
| `shellcheck` (tracked `*.sh` minus `lib/test/`) | exit 0 |
| `shellcheck --extended-analysis=false lib/test/run.sh` | exit 0 |
| `shellcheck lib/test/modules/review-trigger-helpers.sh` | exit 0 |
| `ruff check` | All checks passed |
| `generate-capability-profiles.py --check` | exit 0 |
| `cloud_writer_contract.py check` | closure OK |
| CI required check `lib + python tests` | pass |

The `#810` mutation-routing gate is clean: this change adds no pin literal and re-spells none.

Issue #666, opened 2026-07-21
Make the behavioral-fix-pin mutation-check mandate mechanically enforced instead of prose-only


The01Geek:
## Problem Statement

The pin discipline's highest-leverage rule is the one rule nothing verifies. A **behavioral-fix pin** — a coverage pin added *specifically because* removing the pinned text would re-introduce a named bug — must be expressed through the mutation-taking `assert_pin_red_under`, so the pin proves it catches the *guarded regression* rather than merely its own line vanishing. That mandate is stated in prose at four sites and enforced by nothing.

An author who reaches for a plain `assert_pin_unique` instead produces a call site that is byte-indistinguishable from a legitimate structural pin. The suite stays green, the guard may be framing-only, and the regression it claims to cover remains shippable. This is the same fail-open the mandate was written to close, one level up: `assert_pin_red_under` proves a pin is non-vacuous, and nothing proves the pin reached `assert_pin_red_under`.

Who hits it: every author of pins in this repo — the autonomous `/devflow:implement` and `/devflow:review-and-fix` runs that add most of them, and the maintainer reviewing those runs. A framing-only pin that ships is invisible until the regression it was supposed to guard recurs.

## Current Behavior

`lib/test/run.sh` defines the mutation-taking assertion and the suite drives two static meta-guards over its own pin corpus, but neither guard knows what a pin is *for*.

- `assert_pin_red_under <name> <literal> <mutation> [file]` applies a `sed -E` mutation to a scratch copy, checks sed's own exit status, rejects a no-op mutation via `cmp -s`, and asserts the transition `PASS->FAIL`. A framing-only pin surviving the mutation reads `PASS->PASS` and is reported RED.
- `lib/test/pin-corpus-lint.py` implements exactly two subcommands: `lint` (pin-in-comment) and `wrapped` (wrapped-literal). Neither checks that a behavioral-fix pin carries a mutation check.
- The mandate lives in prose at four sites (see Technical Context). Nothing reads it.

Both gaps are live today:

- A behavioral-fix pin authored as `assert_pin_unique` ships silently.
- A mutation that blanks the target passes: `1,$d` deletes every line and `s/.*//` empties every line, and in both cases sed exits 0, the copy is not byte-identical, the literal is gone, and the transition reads `PASS->FAIL`. The existing rc and `cmp -s` guards cover neither.

## Desired Behavior

A declaration gate and a runtime guard, one purpose: a pin that claims to guard a named regression proves it does.

[...]

Issue #313, opened 2026-07-04
Add opt-in third-party model provider support to the cloud tier (per-workflow endpoint, auth, and model selection)


The01Geek:
## Problem Statement

DevFlow's cloud tier is hard-wired to Anthropic: every workflow authenticates with `CLAUDE_CODE_OAUTH_TOKEN` and there is no way to point a run at an Anthropic-compatible third-party endpoint. Repo owners who want to run part of the pipeline on cheaper models — the concrete driver is experimenting with GLM-5.2 for `/devflow:implement` runs at roughly 1/6th of Opus cost — cannot do so today, even though Claude Code and `anthropics/claude-code-action` both support custom endpoints via `ANTHROPIC_BASE_URL` (officially documented at code.claude.com/docs/en/llm-gateway-connect, community-confirmed with Z.ai in anthropics/claude-code-action discussion 673).

## Current Behavior

The model flows `config .claude_model` → `read-project-config` composite action → jq extract → `claude_args: --model <id>` in all three `anthropics/claude-code-action@v1` invocations. Auth is exclusively the `claude_code_oauth_token` input (and `devflow-runner.yml` additionally declares that secret `required: true` on its `workflow_call` interface). None of the three action steps has an `env:` block, so there is no seam for endpoint injection. The `--effort` flag is unconditionally present in every `claude_args` (its *value* is already config-resolved per section via `.devflow*.effort // "high"` — it is the flag's presence that cannot be dropped for a provider that rejects it). There is no per-workflow model override in practice (the schema text "individual workflows may override" is aspirational: `devflow-runner.yml` declares a `model` input that its only caller, `devflow-review.yml`, never passes — dead machinery; `devflow-review.yml` itself contains no `claude_model` handling at all, it delegates model resolution entirely to the runner).

## Desired Behavior

[...]

PR #982, opened 2026-07-31
chore: align the plugin description, de-vendor allowlist comments, ignore root scratch files

The01Geek:
Three unrelated-but-small prose alignments left over from the directory submission, grouped because each is a one-line change with a coupled site to reconcile.

## 1. `plugin.json` description now matches the text published in the directory

The manifest and the submitted directory listing had drifted apart:

| | Text |
|---|---|
| **Was** (manifest) | Turn a one-line request into a tested, reviewed, documented PR ready for your review: implement, review, docs, and a self-improving retrospective loop. |
| **Now** (submitted) | Turn feature requests into review-ready pull requests: issue, implementation, tests, a review-and-fix loop, and docs. Runs locally or on GitHub Actions, and is built for complex brownfield repos. Each week its retrospective reads what you actually merged and proposes improvements to your skill extensions. |

**Two coupled sites, reconciled in the same commit** — this is exactly the pair that would otherwise ship CI-red:

- `.claude-plugin/marketplace.json`'s `plugins[0].description` is held **byte-identical** to `plugin.json`'s by `#671 packaging` pin (2d). Updated in lockstep.
- `#671 packaging: canonical plugin description is at most 160 characters` — the submitted text is **306**. The ceiling is raised to **320** and now carries a rationale, which it previously lacked.

**On raising that ceiling — flagging the trade-off rather than burying it.** The old 160 had no recorded reason. The published directory text is the canonical user-facing string, so the manifest has to be able to hold it; the alternative (keeping the manifest short and letting it disagree with the listing) is the drift this task exists to close. But `plugin.json`'s `description` is also what renders in the **plugin picker**, and 306 characters is long there — that is a product call, and if you would rather keep the picker string short, the right fix is to change *both* the directory listing and the manifest to one shorter text, not to reintroduce the divergence. I did not assert equality against the published text: that would be a wording-only pin, which `#375`/`#666`/`#810` prohibit and which the mutation-routing gate would reject.

`CITATION.cff`'s `abstract` and `marketplace.json`'s top-level `description` are **separate** marketing prose, not pinned to this string, and are left alone.

## 2. Three workflow comments de-vendored

The model-id allowlist comment in `devflow-implement.yml`, `devflow-runner.yml`, and `devflow.yml` justified the regex by naming a third-party vendor and product:

> covers `claude-*`, provider/model, **Z.ai** bracket-suffix ids like **`glm-5.2[1m]`**, and Bedrock-style colon ids

These were left in place when the provider-routing docs were withdrawn, because editing them would have broken that PR's byte-identical capability proof. That constraint has expired. The comments now describe the *shape* the regex admits without naming a vendor — which is also what the allowlist actually cares about, and what `CLAUDE.md`'s "no product names in committed files" convention asks for.

**No executable line is touched** — the `case` pattern is byte-identical.

Residual, deliberately left and reported rather than silently swept: `lib/test/run.sh` still uses `z-ai/glm-5.2` / `glm-5.2[1m]` as **executable test-fixture inputs** for the `#313` provider-routing arms. Those are real-world id shapes the guard must accept, not branding; rewriting them would be churn against working coverage.

## 3. `/*.tmp` added to `.gitignore` — the owner's files are NOT deleted

`eo.tmp`, `p.tmp`, `p2.tmp` sit untracked **and unignored** at the repo root, so they dirty `git status --porcelain`. That is not cosmetic: the `#434` stale-prose self-scan grades committed HEAD and therefore **self-skips on a dirty tree**, taking a `blocking-gate` skip — a real gate silently disarmed by stray files.

This repo has already accepted exactly this fix for exactly this failure, and says so in `.gitignore`:

> `run-shard.sh` creates the directory and truncates its log BEFORE invoking the suite, so without this entry `git status --porcelain` reports `?? shard-tally-out/` and the `#434` stale-prose self-scan … took its `blocking-gate` skip on every CI run. That is a gate disarmed silently by a path choice.

So this follows established practice rather than setting new policy. The rule is **root-anchored** (`/*.tmp`, not `*.tmp`) so it cannot hide a nested fixture, and **no tracked file matches it** (`git ls-files | grep '\.tmp$'` is empty), so nothing is masked. The three files are left exactly where they are — they are yours, and if they are still wanted, moving them outside the repo is the cleaner end state.

## Verification

Full suite via the direct leading token `lib/test/run.sh` (sanctioned `preexec_fn=SIG_DFL` + `start_new_session=True` launcher — it exceeds the 600s foreground ceiling), both `shellcheck` invocations, `ruff check`, and the generator `--check`s. `origin/main` merged in, not rebased. Summaries in the thread.


The01Geek:
**Verification**

Full suite, direct leading token `lib/test/run.sh`, sanctioned `preexec_fn=SIG_DFL` + `start_new_session=True` launcher:

```
14238 passed, 0 failed
RC=0
```

**0 failed, 0 skipped.** Branched from `origin/main` @ `dca492ac` (post-#968), so the current pin census is what was verified.

| Gate | Result |
|---|---|
| `shellcheck` (tracked `*.sh` minus `lib/test/`) | exit 0 |
| `shellcheck --extended-analysis=false lib/test/run.sh` | exit 0 |
| `ruff check` | All checks passed |
| `generate-capability-profiles.py --check` | exit 0 |
| `cloud_writer_contract.py check` | closure OK |
| CI required check `lib + python tests` | pass |

Manifest sanity after the change: both descriptions byte-identical, length 306, ceiling 320; `marketplace.json`'s top-level description (410 chars, separate prose) untouched.

Issue #749, opened 2026-07-23
Right-size the blocking Step 1 docs-verification phase in /devflow:create-issue


The01Geek:
## Problem Statement

`/devflow:create-issue` Step 1 dispatches one fixed-cost, blocking documentation-verification pass before the run may ask its first question, and the pass's breadth is set by a whole-repository code search rather than by the question being answered. The cost is therefore the same for a one-line docs fix and for a cross-surface engine change, the phase is the only mandatory pre-draft step with no defined failure behavior, and part of what it produces has no consumer.

Measured across recorded `/devflow:create-issue` sessions, filtered to those whose first-turn user message carries the command invocation and deduplicated across duplicated project directories:

| Measurement | Value |
| --- | --- |
| Blocking Step 1 dispatches measured | 33, across 29 sessions |
| Blocking wall-clock: min / p25 / median / p75 / p90 / max | 13s / 194s / **247s** / 329s / 487s / 619s |
| Aggregate blocking wall-clock | 152 min |
| Step 1 of the run that drafted this issue | 384s, 119,263 subagent tokens, 39 tool calls |
| Step 1 share of session wall-clock | median 4.5%, p90 41%, max 88% |
| Sessions where any other agent ran concurrently with the blocking pass | 1 of 28 |
| Report size | median 12,609 chars (~3.2K tokens); max 28,456 (~7.1K tokens) |
| Files named by the report that the main thread later re-opened | median 75%, mean 72% |
| Hard dispatch failures | `Unknown skill: devflow:docs-verify` ×3 across 2 sessions; user rejection ×2 |

The 75% re-open figure is evidence **for** the phase's grounding role, not against it: three quarters of the files the report names are subsequently opened, which is the map being used exactly as Step 3 intends. The waste is not the map. It is that the survey's breadth is set by the repository rather than by the question, that the pass blocks everything while it runs, and that part of its declared output is read by nothing.

## Current Behavior

`skills/create-issue/SKILL.md` Step 1 invokes `/devflow:docs-verify --report-only` unconditionally on every run, with no predicate and no arm selection. A completion-wait discipline makes it blocking: the findings must be complete before the first Step 2 clarification question, and on a zero-question run before Step 3 drafting.

[...]

PR #998, opened 2026-07-31
feat(docs-verify): re-aim --report-only as a docs-first code explorer

The01Geek:
## Problem

`/prflow:create-issue` Step 1 dispatches `/prflow:docs-verify --report-only` to ground a new ticket in current code behavior. The report-only path inherited write mode's global identity — *"You are a Documentation Accuracy Verification Agent"* — so a pass dispatched to map a system returned a documentation-drift audit instead.

Create-issue cannot fix this from its side: Step 1 states that each leg reaches its peer as docs-verify's *search-space operand*, **"never as dispatch-prompt prose its own contract overrides."**

## What baseline testing found

Two fresh-context peers were run against the **unchanged** skill before anything was edited (`superpowers:writing-skills` RED phase). That corrected one premise and surfaced two defects prose review had missed:

- **The code map was already good** — my assumption that it came back thin was wrong.
- **The verdict had no stated boundary.** Two peers found the *same* discrepancy in `config.schema.json` and classified it differently (`DRIFT FOUND` vs `DOCS ACCURATE`), because neither could tell whether a schema file counts as "documentation". That token drives create-issue's shallow→deep escalation.
- **The duty floor was inert.** All six duties returned `discharged` in every run, so the escalate-on-`unestablished` trigger never fired.

## Changes

- **Report-only declares its own identity** — a codebase exploration agent whose deliverable is a map of current behavior; documentation is its entry point and provisional evidence, not its subject. Drift is reported briefly, as an observation.
- **Brownfield disposition**, unconditional (this plugin targets brownfield codebases), with the breadth/depth relationship stated explicitly so it cannot read as contradicting the duty floor: *the floor bounds how many things you examine, never how carefully.*
- **Three fates for a doc-derived claim** — confirmed against implementing code, contradicted by it, or returned marked `doc-sourced, unconfirmed`. Never silently promoted.
- **A stated verdict boundary** — the verdict ranges over documents inside the configured docs location only; a discrepancy outside it is reported but never moves the token. An unreadable docs location is `unestablished`, not `DOCS MISSING`.
- **A stated bar for `discharged`** — you can state the duty's answer *and* cite where you read it — plus a rule closing the `discharged, with a caveat` loophole observed mid-iteration.
- **Four techniques adopted from `agents/code-explorer.md`** — entry-point/chain/writer/failure-path reading, `file:line` citation *with* its scope rule (ephemeral analysis vs committed docs), essential-file marking, and quantitative-claim calibration. Re-expressed rather than copied, so no Apache-2.0 attribution change is required.
- **The write-mode half moved** to `references/write-mode.md` behind a **fail-closed** boundary-marker gate — deliberately stricter than create-issue's degrade-best-effort references, because a write-mode run edits files. Report-only never loads it.

## Compatibility

Every flag, verdict token, report field name, and duty name is byte-stable. `create-issue` Step 1, the issue template's Documentation Drift coupling (`producer=1 consumer=1`), and the `#749` AC7/AC13/AC26 pins are untouched. No agent, workflow, config, or `install.sh` change.

## Verification

- `create-issue-contract` module: **222 passed, 0 failed**, all pre-existing `#749` pins green. Seven new assertions cover the marker contract both directions, the routing row, its fail-closed arm, and the negative arm.
- ShellCheck 0.11.0 (tracked shell, `run.sh`, the CI-listed module) and `ruff`: clean.
- Full suite: running at the time of opening; draft until its summary is read.

**Writing-skills evidence:** RED baseline (2 runs) → GREEN (2 runs) → REFACTOR on an observed new rationalization → verification (2 runs), plus one run on a deliberately harder topic. Eight fresh-context runs total.

## Measured effect, and two honest residuals

| | Before | After |
| --- | --- | --- |
| `file:line` citations | almost none | extensive |
| Essential-file marking | absent | present |
| Failure paths / coupling | mostly missed | surfaced consistently |
| Drift section | up to two dense paragraphs | one paragraph |
| Verdict classification | same discrepancy, two answers | rule applied consistently |
| Tokens per peer | ~117k | ~112k (inside noise) |

1. **The context-saving claim is withdrawn.** A report-only peer now reads 3110 words, up from 2422 (+28%) — the split removed ~580 and the new guidance added ~1270. Prompt size was never the cost driver (<3% of a peer run); the split buys focus, not tokens.
2. **`unestablished` still never fired**, across all eight runs. The specific over-claim it targeted did not recur, but whether the token is now correctly calibrated or still unreachable is undetermined — so the escalation trigger is **not** claimed live.

This is upstream of create-issue's audit-round convergence, which is the dominant runtime cost and gets its own change.


The01Geek:
/prflow:review

Full suite green on the merged tree: 14380 passed, 0 failed, 0 skipped. All lint gates clean (tracked shell, run.sh under ShellCheck 0.11.0, the edited test module, ruff). Branch is current with main.

Issue #988, opened 2026-07-31
Rename the consumer-facing devflow_* config keys to prflow_*, with a three-part migration


The01Geek:
## Dependencies
Blocked by #987 — it changes `skills/init/SKILL.md` and `install.sh`, the same two surfaces this issue's migration and detect-and-route work touch, so landing this first would conflict with it.

## Problem Statement

The plugin is now `prflow`, but a consumer's `.devflow/config.json` still carries seven brand-named top-level keys from the old name: `devflow`, `devflow_implement`, `devflow_runner`, `devflow_review`, `devflow_review_and_fix`, `devflow_retrospective`, `devflow_version`. A fresh install on a repository with no legacy config produces a config full of the superseded brand.

The people who hit this are new adopters, who read a config that disagrees with every command they type, and the maintainer, who pays the migration cost once now instead of once per consumer later. The consumer base today is three repositories, two of them the maintainer's own — a maintainer-supplied figure, recorded as provenance rather than as a verified premise, since the installed base is not observable from this repository. After the plugin directory listing goes live, the base becomes repositories whose configs cannot be seen and cannot be fixed. This is the cheapest this change will ever be.

The rename was previously frozen for a real reason: these keys are read from consumers' committed configs, and most extractions carry a `// default` fallback, so a rename lands as a silent revert to defaults — settings that look configured but are not. This issue ships the rename together with the migration and the detection that make that revert loud instead of silent.

## Current Behavior

A consumer's committed `.devflow/config.json` carries the seven `devflow_*` keys. Nothing in the tree rewrites an already-installed consumer's config, the schema sets `additionalProperties: true` so a stale block is not a schema error, and `skills/init/SKILL.md` carries no migration capability. If the keys were renamed today, the consumer's config would move independently of the two other artifacts that read it, and five different things would happen depending on which key was left behind — four of them quiet:

1. **The shared scaffolder grafts the new blocks with template defaults.** `scripts/scaffold-config.sh` deep-merges the shipped example over the existing config, adding every key the example carries and the config lacks. Once the example carries `prflow_*`, the consumer's next `install.sh` run or `/prflow:init` creates all seven new blocks holding **example defaults**, beside the untouched `devflow_*` blocks. Readers then resolve the new key and get defaults rather than the consumer's values. This is the dominant path, and it reaches the silent revert before any detection can run.
[...]

Issue #956, opened 2026-07-30
Make concatenated in-module pin targets inspectable: unfreezes 11 of the 15 remaining untouchable pin lines


The01Geek:
Found while doing the #948 follow-through in #955. It is the larger half of a problem #948 recorded as a single site.

## The symptom

Some retained pins can never have their `# structural-pin-ok:` declaration corrected. Any edit to the line brings the site into the gate's scope, and it then fails with:

```
typed structural declaration target cannot be inspected
```

The line is therefore permanently uneditable — not just the tag, the whole logical line. One such pin already ships with a comment above it explaining that its stale tag text is deliberately left stale because no edit to it can pass.

## Measured scope

#948 recorded **one** site in this class. There are **11**, found in #955 by attempting all 35 invalid-category tags and reading what the gate refused:

- 10 in `lib/test/modules/create-issue-contract.sh` (targets `$CI_BUNDLE`, `$CI_DV`)
- 1 in `lib/test/run.sh` (`:2791`)

They are 11 of the 15 lines that remain frozen after #955, so this single cause dominates the residual.

## Why it happens

Those pins target a **concatenated in-module bundle** — a variable holding several skill/reference files joined together, built inside the module at runtime. `_typed_pin_inspection_error` cannot resolve such a target to a real file on disk, so it reports the inspection failure before the ladder ever runs. The pin itself is fine and its assertion passes; only the *declaration* path is blocked.

## What to consider

Some options, roughly in increasing ambition — the choice is a real design decision, not a foregone conclusion:

1. **Teach the inspection to resolve a concatenated bundle** to its constituent files and inspect them as a set. Closest to the existing model, and the census already resolves several bundle variables for other purposes, so there may be machinery to reuse.
2. **Treat an unresolvable-but-declared target as a distinct, non-blocking state** rather than a finding — the pin keeps its declaration, and the inspection failure becomes a recorded limitation instead of a refusal. Cheaper, but weakens a check, so it needs its own argument.
3. **Give the affected pins a resolvable target** by pinning against a single constituent file instead of the bundle. Narrow and mechanical, but it changes what each pin asserts, so each one needs its own judgment — this is a per-pin change, not a sweep.

Whichever route, the acceptance test is concrete: edit the declaration on one of the 11 and have `python3 lib/test/pin-corpus-lint.py mutation-routing-worktree .` return rc 0.

## Out of scope

The other four frozen sites have different causes and are not fixed by this:

[...]

Issue #484, opened 2026-07-14
Implement profile: grant the helpers #455 leaves out — including the review engine's, which Phase 3 runs inline — and add the head↔grant desk guard


The01Geek:
## Dependencies
**Unblocked** — #455 landed via PR #480 (merged 2026-07-14). Build on the merged tree: PR #480 rewrote the same baked `--allowed-tools` literal in `.github/workflows/devflow-implement.yml` and added the implement-tier shape lint to `lib/test/run.sh`; do not re-derive the grant form its probe table already proved.

**Supersedes #490** (closed as overlapping). #490 independently specified three of the same helper grants (`extract-doc-needed-paths.sh`, `react-to-trigger.sh`, `load-prompt-extension.sh`), the same `extract-command-heads.py` allowlist-block parse mode for `devflow-implement.yml`, and the same `lib/test/run.sh` head-guard pin over `skills/implement/**`. Its three unique acceptance criteria are folded in below (marked *(from #490)*). Where the two issues conflicted — #490 granted `mktemp` and `git rev-list`; this issue withholds them — this issue's analysis stands: both heads are invoked only from leading `VAR=$(…)` capture shapes the matcher refuses regardless of a head grant (the #401 class), and `git rev-list` is additionally a shallow-clone hazard at `fetch-depth: 50` (see Implementation Notes). `mktemp` joins the withheld list so the guard stays honest over fences that emit it inside a `$(…)` capture. #490's failed cloud run (run 29374618377) left branch `issue-490-cloud-implement-tier-grant-the-remaining` with zero commits beyond `main` — nothing to salvage; its workpad's RED observation (the head pin naming ungranted heads before any grant landed) independently corroborates this issue's Current Behavior table.

## Problem Statement

#455 fixes the two label helpers. It explicitly fixes nothing else: its AC1 reads *"grants the two label helpers … and **no other new grant is added anywhere**,"* and PR #480's diff adds exactly `Bash(.devflow/vendor/devflow/scripts/apply-labels.sh:*)` and `Bash(.devflow/vendor/devflow/scripts/ensure-label.sh:*)`.

Six more bundled helpers, two ordinary commands, and every bare `workpad.py` fence remain ungranted on the implement profile. They are silently refused on every cloud `/devflow:implement` run — the #363 class: no error, no verdict, burned budget, and a run that can end with the lifecycle unfinished.

[...]

Issue #458, opened 2026-07-13
Security: base-branch .claude/settings.json Stop hooks execute PR-head scripts inside claude-code-action, bypassing the #402 deny-floor


The01Geek:
## Problem Statement

Run `29224205805` established a fact with a security corollary that must be addressed: **base-branch `.claude/settings.json` `Stop` hooks execute inside `claude-code-action`.** The hooks are harness-level shell — they run outside the `--allowed-tools` matcher, the `#363` head extractor, the `#401` shape rules, and the `#402` tree-mutation deny-floor entirely.

The hook *configuration* comes from base (`claude-code-action` restores `.claude/` from the base branch), but the *scripts the hooks execute* do not: `.claude/settings.json`'s three Stop commands are `bash lib/efficiency-trace.sh --persist || true`, `bash lib/implement-stop-guard.sh`, and `bash "${CLAUDE_PROJECT_DIR:-.}"/scripts/stop-hook-probe.sh`. `lib/` and `scripts/` are **not** under `.claude/`, so they are supplied by `actions/checkout` at the checked-out ref — the **PR head** in the review job (`devflow-runner.yml` checks out `ref: ${{ inputs.ref }}` = the PR head SHA; `devflow-review.yml` fires on `pull_request_target`, so repo secrets resolve).

Composing those verified facts: **a PR author who edits `lib/efficiency-trace.sh` (or `lib/implement-stop-guard.sh`, or `scripts/stop-hook-probe.sh`) obtains unmediated shell execution at Stop inside the job that reviews their PR — a job holding that job's token — bypassing the `#402` deny-floor.** This is the `#404` REJECT class stated in CLAUDE.md's own words: *"the review job checks out the PR head, so a checked-out-tree copy — committed vendor dir or `scripts/` — is PR-author-editable, and a floor the PR controls is no floor."* The channel was latent and unproven until run `29224205805`; it is now confirmed open.

## Current Behavior

The `#402` deny-floor (`scripts/filter-runner-tools.sh`) strips tree-mutation *tool grants* from the review profile, and is deliberately executed only from a trusted source (base ref / fresh vendored clone), precisely because a PR-author-editable floor is no floor. **Stop hooks bypass that floor entirely:** they are not tool grants, they are shell the harness runs at session end, and they `exec` scripts from the PR-head checkout. Nothing today constrains what a `.claude/settings.json` Stop-hook target does when the reviewed PR has edited that target's source.

[...]

PR #1005, opened 2026-07-31
PRFlow rename Tier 1: atomic directory/vendored-path/config-key/workflow-body migration + /prflow:init

### LINKED ISSUE #1002: PRFlow rename Tier 1: atomic directory/vendored-path/config-key/workflow-body migration + /prflow:init (from #992)
## Dependencies
Blocked by #992 — the parent issue's /prflow:implement run (the dead-commands slice) landed first; this atomic Tier 1 migration must be built on top of it.

## Problem Statement
Follow-up to #992. The plugin is `prflow`, but everything it *ships into a consumer repository* still carries the `devflow` brand at the structural layer: the `.devflow/` state directory, the `.devflow/vendor/devflow/` vendored plugin path, the `devflow_*` config keys, the shipped workflow bodies, and the `.gitignore` rules naming the directory. #992's dead-commands slice fixed only the runtime messages; the Tier 1 structural migration was deferred because it is an atomic, high-blast-radius change (755+ vendored-path occurrences across 58 files, `.devflow/` in 231 files) whose half-application the parent issue calls "worse than no migration". It needs its own dedicated, human-supervised PR.

## Current Behavior
A consumer installs `prflow`, types `/prflow:…`, then reads a tree that disagrees with itself at every structural layer — `.devflow/` state dir, `.devflow/vendor/devflow/` path, `devflow_*` keys, `devflow`-named workflow bodies, and matching `.gitignore` rules.

## Desired Behavior
No shipped artifact carries the `devflow` brand after the migration (except the identifiers listed under the parent's Frozen section), and `/prflow:init` migrates an un-migrated consumer repo as a single atomic unit — directory move, workflow-content rewrite, marketplace-source rewrite, and version bump landing together, or none of them.

## User Impact
Every new adopter pays a first-run confusion cost that grows more expensive to reverse the longer the brand split persists; a clean migration removes it.

## Technical Context
- **Relevant Classes/Files** — `install.sh` (copy loop, marketplace.json writer), `scripts/scaffold-config.sh` (deep-merge backfill), `skills/init/SKILL.md`, the six repo-root-anchored `.devflow/` config readers, the shipped workflows (`devflow.yml`, `devflow-implement.yml`), `.claude-plugin/marketplace.json`, `.gitignore`, `lib/capability-profiles.json`, `scripts/devflow-cloud-writer-contract.json`.
- **Architecture Alignment** — the migration must respect the cloud allowlist's per-literal-path grants (a partially-moved tree is silently denied), the scaffolder's anti-graft precedent, and the trigger-time-vs-runtime config-read distinction.
- **Cross-layer Impact** — installer, config, workflows, vendored plugin, and the generated cloud-writer contract.
- Parent issue #992 was scoped to a single PR by its /prflow:implement run; see the workpad on #992 for the full scope decision. Read #988 in full first (its verified analysis of the scaffolder deep-merge, the four extraction failure behaviours, and the trigger-time channel applies unchanged), and read #987 for the migration machinery already provided.

## Acceptance Criteria
- [ ] No shipped artifact — the state directory name, the vendored plugin path, any config key, any shipped workflow body, or any `.gitignore` rule naming the directory — carries the `devflow` brand after the change, except for the identifiers listed under Frozen.
- [ ] `/prflow:init` applies the complete Tier 1 set to an un-migrated consumer repository, or applies none of it. A run that cannot complete the set leaves the repository byte-identical and reports why. There is no partial-application path.
- [ ] The directory move, the workflow-content rewrite, the marketplace-source rewrite and the version bump are observable as a **single atomic unit**: a suite assertion fails if any one of them can be applied without the others.
- [ ] Every consumer-side artifact that references the vendored plugin path — including `.claude-plugin/marketplace.json`'s local `source` — resolves to the new path after migration. A post-migration consumer tree contains no reference to the old vendored path.
- [ ] `scripts/scaffold-config.sh`'s deep-merge backfill does not graft a new-name key while its superseded counterpart is present. This holds on the migration path, on any refusal path, and on the degraded path where `jq` is unusable.
- [ ] Config-key migration satisfies #988's acceptance set. The implementing run reads #988 and treats its criteria as inherited unless this issue supersedes one explicitly.
- [ ] `/prflow:init` reports what it migrated, and separately reports every item it could not migrate, naming each. The report distinguishes a preview from an applied run.

## Implementation Notes
- **Approach** — implement the whole consumer-visible Tier 1 set as one atomic `/prflow:init` operation, with a suite assertion proving the four sub-artifacts cannot be applied independently. Do not split the Tier 1 set across PRs (parent Open-Question 6).
- **Code Patterns** — mirror the scaffolder's existing anti-graft guard; regenerate `scripts/devflow-cloud-writer-contract.json` in the same commit as any pinned-asset edit; bump the version via a `.changeset/*.md`, never by editing `plugin.json` directly.
- **Potential Gotchas** — the cloud allowlist grants are per-literal-path, so a partially-moved tree is silently denied, not merely broken; the scaffolder deep-merge is the dominant silent-revert hazard (#988 finding 1); `workflows.devflow` is frozen and invisible to a brand grep; `.devflow/learnings/*.jsonl` and `.devflow/logs/*.tsv` contents stay byte-identical.

The01Geek:
Closes #1002.

Renames everything PRFlow ships into a consumer repository from `devflow` to `prflow`, and makes `/prflow:init` migrate an existing repository as a single atomic unit.

## What moved

| | from | to |
|---|---|---|
| state directory | `.devflow/` | `.prflow/` |
| vendored plugin | `.devflow/vendor/devflow/` | `.prflow/vendor/prflow/` |
| config keys | `devflow`, `devflow_{implement,runner,review,review_and_fix,retrospective,version}` | `prflow_*` |
| shipped workflow **bodies** | the paths and keys above | the same, renamed |
| `.gitignore` rules | `/.devflow/*` and its negations | `/.prflow/*` |

The mechanical pass was rule-driven and auditable: 238 files, 5453 substitutions, validated against **13 frozen identifier classes with zero violations**. Frozen and untouched: `workflows.devflow` / `workflows.devflow-review`, `devflow-marketplace`, every `.github/workflows/` **filename**, the `DevFlow` label, the `devflow-telemetry` branch, the `<!-- devflow:* -->` markers, the `DEVFLOW_*` env vars, `devflow_module_pin_*`, the `devflow:` agent-override namespace, and the **byte contents** of `learnings/*.jsonl` + `logs/*.tsv` — those move with the directory and are never rewritten. Tier 2 is #1003; Tier 3 is #1004.

## The migration is all-or-nothing

`scripts/migrate-consumer-tier1.sh` plans, validates every precondition for all four members, stages every new byte, then commits behind a rollback journal. A single unsatisfiable member refuses the whole set and leaves the repository **byte-identical**.

That fail-closed posture deliberately inverts this repo's best-effort default, and the reason is in the helper's header: the shipped workflows invoke bundled helpers at the vendored path as repo-relative leading tokens and the cloud grants are per-literal-path, so a half-moved tree is silently **denied**, not merely broken — the run ends with no verdict.

Both entry points call the one helper: `install.sh` runs it first inside `devflow_apply_all`, where the workflow copy loop now **shares its fate**; `/prflow:init` runs it before the scaffolder.

**AC3 is discharged executably**, not in prose: one suite arm per member blocks *that member alone* and asserts the run refuses, names it, and leaves the tree byte-identical. Four RED arms if the apply is ever made non-transactional.

## Detection where a migration has not happened yet

- **The state directory has a loud transitional read-through; the config keys deliberately do not.** `/prflow:init` registers the marketplace with `autoUpdate: true`, so a consumer's *plugin* can update ahead of any migration run — without a fallback the next auto-update resolves an absent config and silently reverts every defaulted read. Readers resolve `.prflow/` first and fall back only when it is absent and `.devflow/` present, breadcrumbing every time. The end criterion is confirmation-gated, recorded in `lib/rename-map.json`.
- For the **keys**, #988's ruling stands — a silent key fallback makes the migration unobservable and therefore permanent. Detection instead: `config-get.sh`'s superseded-key probe, which distinguishes a genuinely absent key from one a consumer set to `""`/`false`/`0`/`null` (a distinction the resolver structurally could not make), and a per-family `::error::` guard in the two shipped workflows' `config` jobs, because the trigger-time channel reads config through inline `jq` and no breadcrumb can reach it.

## Defects found and fixed along the way

- **A security gap.** `config-get.sh` now sources `resolve-state-dir.sh`, which escaped the #458 hardened Stop-hook closure. On the review tier that would displace `config-get.sh` with trusted base-ref bytes and then source the **PR-head** copy of the resolver. The helper is now in the closure.
- **A guard made silently vacuous.** The mechanical sweep respelled only the first segment of a backslash-escaped ERE, so the #484 disposable-mutant fixture deleted no grant and the guard passed without testing anything.
- **A frozen key half-renamed.** The Python subscript form of `workflows.devflow` became `w['prflow']` while its own guard still read `'devflow' in w` — a `KeyError` on every config that actually carries the key.
- **`/prflow:init` RED/GREEN/REFACTOR** (`superpowers:writing-skills`) caught three more: the superseded-identifier step named a key the migration had already renamed away, so it took its degrade arm on *every* migrated repo and left the dead App slug in place; the migration reported only the version pin while renaming all seven keys; and the scaffolder asserted a pin predated the rename on a run that had just stamped it.
- **A guard-class-2 violation I introduced.** The new code derived its own location through `dirname`, which `lib/preflight.sh` does not guarantee — under `set -e` the failing substitution aborted `config-get.sh` before the caller's default was emitted. Replaced with the pure-bash derivation `lib/resolve-jq.sh` already documents.

## The pin gate needed a correctness fix, and this is the part to review hardest

The mutation-routing gate reported 43 findings. All were mechanically provable **pure respellings** — and 25 of them had **no in-policy disposition at all** (8 hit a retired-literal/prose-target combination whose message carries no remediation clause; 6 would have required minting a false `boundary` adjudication; 11 have targets the lint cannot resolve).

Root cause: `scan_changed_sources` compared the merge-base site image against HEAD, but **both** resolve target paths against the *current* repository root — so across a rename it measured path *spelling* rather than pin *identity*. The fix resolves a site to the asset it names, completing one layer up the same current-first/superseded-fallback rule this PR already added in `_revision_state_dir_path` for revision-side blob reads.

The exemption is deliberately narrow: exact equality of the whole effective tuple; the superseded→current map applied to the **merge-base side only**, so a HEAD-side superseded spelling cannot be laundered; one base site exempts at most one candidate; frozen names never mapped; an absent or malformed map withdraws the exemption entirely.

**Blast radius, measured on the final implementation:** with the rename map made unavailable, the classifier reproduces the pre-fix behaviour **byte-identically** — same 43 findings, same order, same text. The exemption is the only behavioural difference. Each safeguard was mutation-tested (disable the filter, remove the frozen alternation, drop the key qualification, make it bidirectional, remove one-for-one consumption, reorder frozen last) — every mutation turns a test RED.

## Verification

- `lib/test/run.sh` — a full run is in flight as this PR opens; **the result is posted as a
  comment below.** The previous complete run was `14570 passed, 18 failed, 0 skipped`, and
  every one of those 18 is addressed in the last three commits — do not merge on this
  paragraph, merge on the posted result and on CI.
- 15/15 focused modules green, including the new `tier1-rename-migration` (149 assertions)
- `shellcheck` (all scopes) and `ruff` clean
- all four generated artifacts reconciled (`regenerate-artifacts.py`)
- `learnings/*.jsonl` and `logs/*.tsv` byte-identical, verified by blob compare

## Writing-skills evidence

**Writing-skills evidence:** `superpowers:writing-skills` invoked before the `skills/init/SKILL.md` edit, and its RED/GREEN/REFACTOR discipline followed in full.

- **RED** — a subagent ran the *unmodified* skill against a fixture consumer on the superseded layout. It scaffolded, never migrated, produced no migrated-vs-could-not-migrate report and no preview/applied distinction, and `.prflow/` was never created. It also surfaced that the superseded-identifier step named a hardcoded `.prflow/config.json` that does not exist on an un-migrated tree, so a literal executor takes the degrade arm and leaves the dead App slug in place.
- **GREEN** — the same scenario with the edit migrated all four members, reported both halves, distinguished preview from applied, and preserved every frozen identifier (`workflows.{devflow,devflow-review}`, the workpad-marker value, the workflow filenames, the `learnings/` bytes).
- **REFACTOR** — the GREEN run was asked to be adversarial and returned nine findings. Three were real defects, now fixed: the correction step named a config key the migration step had already renamed away (so it degraded on *every* migrated repo); the migration reported only the version pin while renaming all seven keys; and the scaffolder asserted a pin predated the rename on a run that had just stamped it. Four more were wording/dedup fixes (a self-referential count, a fact relayed twice by two helpers, a duplicated preview plan).

**One deviation, recorded rather than papered over.** The three-token precision reword in `skills/review/SKILL.md` (made late, to clear a pre-existing R4 stale-prose finding my rename pulled into the branch diff) did **not** get its own `writing-skills` invocation. I judged the discipline already loaded in-session and leaned on the review engine's own contract pins — `review-and-fix-contract`, `create-issue-contract` and `review-trigger-helpers` all stay green, and the edit only narrows a backticked operator token to the shape the surrounding prose already scoped it to. That reasoning is stated so a reviewer can disagree with it; it is not a claim that the mandate was satisfied for that file.

## Follow-ups I did not fix here

1. Pin retirement is keyed by **literal** while revival authorization is keyed by **site**, so a `redundant_retire`d twin poisons the retained pin it was written about. Latent for any branch touching those lines, rename or not.
2. `Verification evidence:` and `standalone PR mode only` have no ledger row and no machine consumer; they need a re-adjudication pass, not an ad-hoc `boundary` row.
3. `$REVIEW_BUNDLE` is unresolvable to the #956 bundle-member extractor, making two pins permanently undeclarable.


The01Geek:
Requesting an independent review. CI is green (18/18, including the required `lib + python tests`) on `2455b20c`, and a full local suite run is clean — but green gates mean the checks pass, not that the judgement calls are right, and three of the four things that went wrong on this branch were caught by CI rather than by my own verification.

**Where the review is most valuable, in priority order:**

1. **`lib/test/pin-corpus-lint.py`** — the sanctioned-rename exemption in `scan_changed_sources`. This is a change to a security-adjacent guard. It drops a candidate only on exact equality of the whole effective tuple against a merge-base site, maps the **merge-base side only** (so a HEAD-side superseded spelling cannot be laundered), consumes at most one base site per candidate, never maps a name in the map's `frozen` block, and withdraws the exemption entirely when the map is absent or malformed. Measured: with the map made unavailable it reproduces the pre-fix classifier byte-identically. The claim to attack is that the exemption cannot absolve a *substantive* edit that rides along with a rename.

2. **`scripts/migrate-consumer-tier1.sh`** — specifically the commit window. Stages everything, then applies behind a rollback journal. A crash *inside* the commit is narrow but nonzero; the journal makes it detectable and refused on the next run rather than silently half-done. Worth checking the rollback ordering and whether any partial state escapes it.

3. **`scripts/scaffold-config.sh`** — the superseded-key anti-graft guard keys on the ORIGINAL config, which is what makes it hold on the refusal path and the jq-unusable path, not just the happy one.

**A caveat about this review specifically, which is not a defect in the PR.** `prflow_version` on the default branch is `main`, and that pin is resolved at **trigger time from the default branch** — so the reviewer runs the **pre-rename** plugin against a **migrated** PR head. Its `config-get.sh` will look for `.devflow/config.json`, find `.prflow/config.json` instead, and fall through to defaults; and the prompt-extension loader will not find `.devflow/prompt-extensions/review.md`. So this review runs **without this repository's own review prompt extension** — the file carrying the guard-class rules.

That is the grant-timing bootstrap the issue documents, and there is no in-PR fix: the pin is trigger-time-resolved and in-PR-inert by construction. It also shows the transitional read-through is deliberately one-directional — it lets a *new* plugin read an *old* tree, never the reverse, which is exactly why the migration must advance the pin. Practical consequence for a human reading the output: do not read silence on guard-class issues as clearance.

**Known limitations I am not asking the review to rediscover** — three are filed as #1006, #1007 and #1008, and the diff is ~294 files of which the large majority is mechanical rename churn.

/prflow:review

PR #1014, opened 2026-07-31
feat: record the frozen out-of-repo DEVFLOW_* identifiers and derive a "do not rename" advisory (#1004)

### LINKED ISSUE #1004: PRFlow rename Tier 3: /prflow:init advisory report for out-of-repo DEVFLOW_* identifiers (from #992)
## Dependencies

Blocked by **#1002** (Tier 1), merged as PR #1005. `lib/rename-map.json` now exists and its `frozen._comment` already names this issue as where the `DEVFLOW_*` environment identifiers are tracked.

## Problem Statement

Follow-up to #992. The `DEVFLOW_*` environment variables live outside the repository — in shell profiles, CI secrets and org-level variables — so nothing in the tree can migrate them. Tier 3 decided that `/prflow:init` must *report* them.

**The earlier draft of this issue specified a report that would harm the consumers it exists to protect, and its User Impact statement was backwards.** Measured against the tree:

- The two shipped workflows reference `DEVFLOW_APP_ID` on **36** lines, `DEVFLOW_APP_PRIVATE_KEY` on **13**, `DEVFLOW_PROVIDER_API_KEY` on **8**, `DEVFLOW_RUNNER` on **7**.
- **No `PRFLOW_*` read side exists anywhere in the tree.** Every `PRFLOW_` token is script-internal (state-dir constants, python env-passing wrappers, heredoc delimiters). None is a consumer input.

So an advisory that says "what to rename it to" is an instruction to break your own install — and, worse, to break it **silently**.

## Current Behavior

`/prflow:init` emits no advisory report for out-of-repo `DEVFLOW_*` identifiers; a consumer renaming a secret is on their own.

## Desired Behavior

A specific, actionable advisory that names each consumer-facing `DEVFLOW_*` identifier, says where it is likely set, and states **that the name is unchanged and what happens if you rename it** — derived from a stated criterion, and delivered on a surface its actual audience reads.

## User Impact

**Corrected from the earlier draft, which had this backwards.** The earlier text read: *"A consumer who renames a secret without the report is more likely to break their install than one who never touches it."* Measured, the inverse holds. A consumer who renames the **secret alone** hits a loud mint failure and self-corrects. A consumer who follows the earlier draft's advice and renames **both** the secret and its paired variable reaches the **silent** degradation path: every App-mint gate is `if: ${{ vars.DEVFLOW_APP_ID != '' && … }}`, so an unresolvable variable turns every gate false, every consumer falls back to `steps.app-token.outputs.token || secrets.GITHUB_TOKEN`, and the run goes **green with a degraded identity**.

Downstream, all silent at the point of cause: workflow-file pushes fail with `refusing to allow a GitHub App to create or update workflow …`, which points at App permissions that are fine; the credential refresher reports `credential refresher started (detached)` while its empty-input arm writes `cannot mint` into a redirected log nobody reads, so long implement runs die on an expired token with no trace; review attribution moves to `github-actions[bot]`, which does not satisfy a branch-protection required-approval rule.

**Highest-severity single outcome, absent from the earlier draft:** `DEVFLOW_RUNNER` is read as `runs-on: ${{ vars.DEVFLOW_RUNNER && (…) || 'ubuntu-latest' }}`. Renamed, **every job silently relocates to GitHub-hosted `ubuntu-latest`** — and that same job carries `secrets.DEVFLOW_APP_PRIVATE_KEY` and `secrets.DEVFLOW_PROVIDER_API_KEY` in its env. A consumer who self-hosts for network isolation or compliance has their App private key and provider API key executed **outside the boundary they chose self-hosting for**, with no error naming the cause.

### The criterion (this replaces the hand-list)

A `DEVFLOW_*` identifier is **consumer-facing** iff either:

- **(A1) Cloud channel** — it appears in a workflow shipped by `install.sh`'s copy loop as `secrets.<NAME>` or `vars.<NAME>`. The consumer sets it in GitHub repo/org settings; the tree can only read it.
- **(A2) Local/operator channel** — it is read as an ambient environment variable by a shipped `install.sh` / `scripts/` / `lib/` file **and** is never assigned by any non-test file in the tree. The consumer sets it in a shell profile or an install one-liner; nothing internal supplies it.

Both arms are one `git grep` each. A2's never-assigned-internally clause is what mechanically excludes `DEVFLOW_PROMPT_EXTENSION_ROOT` (assigned by `.github/workflows/devflow-runner.yml`), `DEVFLOW_TELEMETRY_PUSH`, and the ~185 harness/internal names.

**The criterion catches errors the earlier hand-list makes, in both directions:**

- **Missing:** `DEVFLOW_REVIEWER_APP_ID` — the *gating* half of the reviewer pair. The hand-list includes `DEVFLOW_REVIEWER_PRIVATE_KEY` but not the variable that decides whether the reviewer App is used at all.
- **Wrong:** `DEVFLOW_PROMPT_EXTENSION_ROOT` — workflow-written, consumer-read-never. `docs/install.md` says so in the very sentence it appears to have been drawn from.
- **Unadjudicated:** `DEVFLOW_CONFIG_FILE` satisfies A2's mechanics but is documented in none of the consumer-facing docs. Either A2 keys on the *read* (it joins, at 14) or on *doc-declaration* (it stays out, at 13). **The issue must resolve this** — it is the boundary case the stated criterion exists to force a decision on.

Criterion total 13; hand-list total 13. **The counts collide by coincidence** — one false negative cancels one false positive — which is precisely why a matching total must not be read as agreement.

### Three blast radii, not one

- **`vars.`-gated** (`DEVFLOW_APP_ID`, `DEVFLOW_REVIEWER_APP_ID`, `DEVFLOW_RUNNER`) — **silent feature loss**, largest radius. An unresolvable name is indistinguishable from "deliberately not configured."
- **`secrets.`-sourced** (`DEVFLOW_APP_PRIVATE_KEY`, `DEVFLOW_REVIEWER_PRIVATE_KEY`, `DEVFLOW_PROVIDER_API_KEY`) — **loud, but only when the paired gate still fires.** Note the ordering trap: rename the variable first and the secret's loud guard is gated off, so the "safe" rename order is the one no consumer would guess.
- **Operator escape hatches** (`DEVFLOW_BASH`, `DEVFLOW_GH`, `DEVFLOW_JQ`, `DEVFLOW_VENDOR`, `DEVFLOW_REF`, `DEVFLOW_DRY_RUN`, `DEVFLOW_APPLY`) — **silent reversion to the broken state the hatch existed to fix.** All resolve via `:-`/`:=`, so a renamed name is byte-identical to unset. `DEVFLOW_GH`/`DEVFLOW_JQ` are the worst kind: a no-op on a healthy host, so the consumer concludes the rename worked, while on the Windows/WSL host that needed the hatch it silently restores the shadowing bug #245 fixed. `DEVFLOW_DRY_RUN` is the only one that **writes** as a result of the rename: a `curl | bash` preview becomes an apply.

`DEVFLOW_REF` has a **dual role** — consumer-facing for `install.sh`, and internal for `.github/actions/vendor-plugin/action.yml` → `vendor-slice.sh`. The criterion must not imply a name is purely one or the other.

### `/prflow:init` is the wrong home — measured

- It is **local-tier only**: `disable-model-invocation: true`, **0** occurrences in `lib/capability-profiles.json`, and it is in neither cloud trigger set.
- `install.sh` mentions **none** of the six cloud identifiers; their setup lives in `docs/cloud-setup.md`'s configuration table.
- So a **cloud-only consumer — exactly the population that has these set — never runs `/prflow:init`**, while a local-only consumer, who has none of them configured, sees the whole list on every re-run. **The report is delivered to the inverse of its audience.**
- `/prflow:init` also **cannot detect anything**: the skill makes no `gh` calls and nothing reads `actions/variables` or `actions/secrets`, so the report is necessarily a static list that cannot self-suppress — the first unconditional block in a skill whose reporting ethos is "say nothing when nothing is actionable", firing on every no-op re-scaffold.

**Better homes in order:** (1) `docs/cloud-setup.md`, beside the configuration table that already names all six; (2) a `log` line in `install.sh`'s apply path, which is the cloud tier's only executable touchpoint and already carries the `devflow_report_withheld_tier` advisory precedent; (3) `/prflow:init` last, and only as an echo, gated to stay silent when nothing is actionable.

## Acceptance Criteria

- [ ] The advisory names each consumer-facing `DEVFLOW_*` identifier, states where it is likely set (repository secret, organization variable, shell profile), and states **that the name is unchanged, with the consequence of renaming it**. It names identifiers, not a prefix, and it does **not** instruct a rename.
- [ ] Each row carries its **failure mode** — silent skip / loud mint failure / silent PATH reversion / silent un-pin / silent runner relocation. Without this the advisory is more dangerous than silence: it raises salience on 13 names while withholding that 10 of them fail silently.
- [ ] The Case-A/Case-B asymmetry is stated for both App pairs: renaming the **secret** fails loud, renaming the **variable** fails silent, and renaming both is the silent path.
- [ ] The criterion above is implemented, the population is **derived** from it, and the implementing run **reports the diff against this issue's hand-list** — non-empty in both directions — as the evidence the criterion was actually run rather than transcribed.
- [ ] `DEVFLOW_REVIEWER_APP_ID` is included; `DEVFLOW_PROMPT_EXTENSION_ROOT` is excluded; `DEVFLOW_CONFIG_FILE` is adjudicated in or out with the criterion arm that decides it.
- [ ] The freeze is recorded where a tool can see it — a `frozen.env_identifiers` block in `lib/rename-map.json` — and the report's population derives from it, so a future Tier 4 has one place to unfreeze.
- [ ] The report's home is decided against the measured fact that a cloud-only consumer never runs `/prflow:init`. If `/prflow:init` is kept as a home, that choice is justified explicitly and the report is gated to stay silent when nothing is actionable.
- [ ] `lib/test/run.sh` reports zero failures and zero skips.

## Implementation Notes

- **Approach** — ship a **"do not rename" inventory with a stated reason and failure mode per name**, not a rename instruction. That is the single edit that removes the harm, and it adds the value the earlier draft lacked: it answers the question a consumer actually has, which is *why* a `DEVFLOW_*` name persists in a product called PRFlow.
- **Explicitly out of scope, and why.** Renaming the read side with dual-read acceptance was considered and rejected for this issue: `vars.X != ''` gates have no dual-read form short of duplicating `vars.PRFLOW_X != '' || vars.DEVFLOW_X != ''` across ~10 gate conditions plus every `||`-fallback consumer; each is a workflow edit that reaches a consumer only by re-running `install.sh`, giving a two-halves upgrade with a silent-skew failure mode; and `runs-on` is evaluated before any step, so its dual-read is a nested ternary with no test surface. **If the brand inconsistency must be closed, file it as a separate Tier 4** with that real scope, a retention policy, and the two-halves upgrade coupling stated.
- **Code Patterns** — `devflow_report_withheld_tier` in `install.sh` for an advisory NOTICE that names what it cannot fix; `lib/rename-map.json`'s `frozen` block for the machine-readable population.
- **Potential Gotchas** — the tree defines well over a hundred `DEVFLOW_*` names, most internal, so a raw prefix sweep over-reports; the criterion's total matching the hand-list's total conceals two errors; `DEVFLOW_REF` is both consumer-facing and internal.

The01Geek:
Closes #1004.

## What this is — and what it deliberately is not

The deliverable is a **"do not rename" inventory**, not a rename instruction. Nothing in this PR renames a `DEVFLOW_*` identifier, and nothing in it tells a consumer to.

That is the whole point. There is **no `PRFLOW_*` read side anywhere in the tree** — every `PRFLOW_` token in the repository is script-internal (state-dir constants, python env-passing wrappers, heredoc delimiters). So a consumer who renames one of these names does not move a setting; they delete it. And in 11 of the 13 cases they delete it **silently**: every App-mint gate is `if: ${{ vars.DEVFLOW_APP_ID != '' && … }}`, so an unresolvable variable turns the gate false, every consumer falls back to `steps.app-token.outputs.token || secrets.GITHUB_TOKEN`, and the run goes **green with a degraded identity**.

The highest-severity row, called out on its own in the advisory: renaming **`DEVFLOW_RUNNER`** silently relocates every job to `ubuntu-latest`, and that same job carries `secrets.DEVFLOW_APP_PRIVATE_KEY` and `secrets.DEVFLOW_PROVIDER_API_KEY` in its env — so a consumer who self-hosts for network isolation or compliance has their App private key and provider API key executed outside the boundary they chose self-hosting for, with no error naming the cause.

## AC2 evidence: derived vs. the issue's hand-list

The criterion was **run**, not transcribed. `python3 lib/generate-env-freeze-advisory.py --derive` implements both arms over the tree, and the suite reconciles its output against the record on every run. The diff against the issue's hand-list is non-empty in both directions, as the issue predicted:

| Direction | Name | Why |
|---|---|---|
| **Derived, absent from the hand-list** | `DEVFLOW_REVIEWER_APP_ID` | A1 selects it — `vars.DEVFLOW_REVIEWER_APP_ID` appears in a shipped workflow. It is the **gating** half of the reviewer pair; the hand-list carried `DEVFLOW_REVIEWER_PRIVATE_KEY` but not the variable that decides whether the reviewer App is used at all. |
| **In the hand-list, not derived** | `DEVFLOW_PROMPT_EXTENSION_ROOT` | Fails A1 (never `secrets.`/`vars.` in a shipped workflow) and fails A2's never-assigned-internally clause — `.github/workflows/devflow-runner.yml` and `matcher-probe.yml` **assign** it, and its sole shipped reader consults it only on the branch those workflows drive. Workflow-written, consumer-read-never. |

Totals collide at 13 on both sides, which is exactly why the count was not read as agreement. Both names are recorded in `lib/rename-map.json`: the first in `identifiers`, the second in `adjudicated_out` with its deciding arm and its evidence.

**`DEVFLOW_CONFIG_FILE` — adjudicated OUT, on A2's doc-declaration arm.** It satisfies A2's read mechanics (`lib/config-source.sh` reads `${DEVFLOW_CONFIG_FILE:-…}`; no non-test file assigns it) but is declared in **none** of the three consumer-facing documents (`README.md`, `docs/install.md`, `docs/cloud-setup.md`). It is a test and tooling seam that was never published as a consumer setting, so no consumer has been told to set it and none can rename it. It is recorded explicitly in `adjudicated_out` so a later reader sees a decision rather than an omission; if it is ever documented as a consumer override, that same change moves it into `identifiers`.

### A correction to the issue's own arithmetic

The issue framed A2 as "either 14 (keys on the read) or 13 (keys on doc-declaration)". Measured, **the read-only reading gives 31, not 14**: A2's read-and-never-internally-assigned mechanics alone select 25 names, almost all internal tuning knobs no consumer has been told about (`DEVFLOW_REFRESH_*`, `DEVFLOW_TELEMETRY_MAX_*`, `DEVFLOW_GH_WRAPDIR`, `DEVFLOW_RECORDER_*`, `DEVFLOW_SRC`, …). The issue's "the never-assigned clause excludes the ~185 harness/internal names" attributes to that clause work it does not do. The doc-declaration clause is what actually separates the operator surface from the harness, and that reasoning is recorded in the block's `A2_doc_declaration_rationale`.

### One residual, recorded rather than papered over

`DEVFLOW_REF` has a genuine **dual role**: `.github/actions/vendor-plugin/action.yml` assigns it for `vendor-slice.sh`'s fetch branch, while `install.sh` reads it from the operator's own environment. A blanket "never assigned by a non-test file" grep would delete it from the consumer surface on the strength of a producer that never reaches `install.sh`. So the never-assigned clause is applied as a **recorded per-candidate adjudication** rather than as a third grep, and `DEVFLOW_REF`'s row carries a `dual_role` field the advisory renders. The criterion is enforced as: every name the two arms select must be resolved into `identifiers` or `adjudicated_out` — nothing may silently appear or disappear.

## Structural decision 1 — where the report lives

**Decision: `docs/cloud-setup.md` (canonical, generated), a pointer from `docs/install.md`, and an upgrade-gated `install.sh` NOTICE. Not `/prflow:init`.**

The issue's measurements decide it, and I re-confirmed them: `/prflow:init` is local-tier only (`disable-model-invocation: true`, absent from `lib/capability-profiles.json`, in neither cloud trigger set), `install.sh` mentions none of the six cloud identifiers, and their setup lives in `docs/cloud-setup.md`. So a **cloud-only consumer — exactly the population that has these set — never runs `/prflow:init`**, while a local-only consumer would see the whole list on every no-op re-scaffold. The skill also cannot detect anything (it makes no `gh` calls; nothing reads `actions/variables`), so a report there could only be an unconditional static block inside a skill whose reporting ethos is "say nothing when nothing is actionable".

Reasoning for what I did choose:

- **`docs/cloud-setup.md` is where the audience already is.** A consumer asking "the product is called PRFlow now — should I rename `DEVFLOW_APP_ID`?" opens the page where they set it up. The new section sits between `## Required secrets` and `## Choosing the runner (DEVFLOW_RUNNER)` — between the two places these names first appear. It covers both channels, so the inventory stays single-sourced rather than split across two pages.
- **`docs/install.md` gets a pointer**, not a copy, beside the `DEVFLOW_GH` / `DEVFLOW_JQ` / `DEVFLOW_BASH` sections that are the operator channel's own home.
- **`install.sh` gets a NOTICE, gated to an upgrade.** It is the cloud tier's only executable touchpoint and already carries the `devflow_report_withheld_tier` advisory precedent. Gating it to `DEVFLOW_INSTALL_STATE = "an existing"` is the closest available approximation of "silent when nothing is actionable": a first-time installer is creating these names from the docs right now and has nothing to be warned about, while the population that has them configured *and* is now looking at a renamed product is exactly the existing-installation population. Both arms are driven in the suite, including the negative control.

A consequence worth stating: because the report is not in a skill, this PR touches **no `skills/` `SKILL.md` body**, so the `superpowers:writing-skills` mandate does not apply and no `Writing-skills evidence:` line is owed. That was a factor in the decision, not an afterthought.

## Structural decision 2 — machine-readable freeze, and the report derived from it

**Decision: `frozen.env_identifiers` in `lib/rename-map.json` is the single source; the advisory is a GENERATED region rendered from it; the criterion is re-run over the tree by the suite.**

Reasoning: the repo's strongest convention is single-source-plus-generated-output (`capability-profiles.json` → the allowlist literals, `plugin-identity.json` → the baked regions). A hand-authored table reconciled by a checker would be strictly worse — a human would have to hand-edit two places. So:

- `lib/rename-map.json` `frozen.env_identifiers` carries the policy, the criterion, the App-pair asymmetry, the 13 rows (each with `set_where`, `read_as`, `failure_visibility` and `failure_mode`), and the 2 `adjudicated_out` records.
- `lib/generate-env-freeze-advisory.py` renders the `docs/cloud-setup.md` region, banner-stamped with a sha256, and `--check` reports drift with a directional diff. It is registered as a `regenerate` row in `lib/test/regenerate-artifacts.py`, per that file's registration rule.
- `--derive` / `--audit` re-run the criterion and fail when the live selection and the record disagree **in either direction**, with deliberately different remedies: drift in the region says "re-run the generator"; drift in the population says "adjudicate this", because that one is a human judgment and a regeneration would be the wrong instruction.

A future Tier 4 that ships a dual read has exactly one place to unfreeze a row.

**A real bug this surfaced.** The advisory names the two identifiers it *excludes*, and it is rendered **into a document A2 scans for declarations** — so the generated output became an input to its own derivation and re-selected `DEVFLOW_CONFIG_FILE` purely because the sentence explaining its exclusion mentions it. The record could never have converged. The generator now strips its own complete begin/end regions from any document it scans (an unterminated banner falls through to `locate()`'s fail-closed handling rather than truncating the file), and a dedicated fixture asserts that a name declared *only* inside the region is not selected.

## Verification

- `lib/test/modules/tier1-rename-migration.sh` — **191 passed, 0 failed** (assertion floor bumped 149 → 191 in both the registry and the `run.sh` call site).
- Full `lib/test/run.sh` — reported in a comment below.
- `python3 lib/test/regenerate-artifacts.py` → all artifacts reconciled, exit 0.
- `python3 lib/test/pin-corpus-lint.py mutation-routing-worktree .` → completed.
- `shellcheck --severity=warning -e SC1091` over the tracked non-`lib/test` `.sh` set → clean; `lib/test/modules/tier1-rename-migration.sh` → clean.
- `ruff check` over every tracked `.py` → all checks passed.

Every new assertion drives the helper **file-in / exit-code-out**: the criterion re-run over the live tree, both audit failure directions with the name and the direction, fail-closed exit 2 for an incomplete row and for a malformed/absent map, a full render → check → tamper → check round trip over a synthetic tree, fail-closed handling of an absent and of a duplicated banner, and both `install.sh` gate arms. **No assertion pins the advisory's wording.** One assertion is about the deliverable's defining property: the rendered region must propose no `PRFLOW_` counterpart for a frozen name, because a rename column added here would be read as an instruction — and following it degrades a consumer's install silently.


The01Geek:
## Full local suite

```
14630 passed, 0 failed
```

No skip clause, so **0 skipped** — the summary is byte-identical to the no-skip form per the issue-#456 contract. Run detached from a clean worktree at the rebased head, log read directly rather than trusting an exit code.

Focused module: `lib/test/run-module.sh tier1-rename-migration` → **191 passed, 0 failed**.

## CI

All seven checks green on `3dfffd62`, including the required aggregate:

| Check | Result |
|---|---|
| **`lib + python tests`** (required) | pass |
| `lint (shellcheck + actionlint + ruff)` | pass |
| `shard (monolith)` | pass |
| `shard (modules-large)` | pass |
| `shard (modules-pin)` | pass |
| `shard (modules-rest)` — carries `tier1-rename-migration` | pass |
| `shard (python-pool)` | pass |

## One follow-up commit after that run

`e54a11e2` removes a **rot-prone count from my own prose**: the `frozen.env_identifiers.policy`
string said "11 of the 13 delete it SILENTLY". The measured tally is **10** — the two App
private keys fail loud while their paired variable still resolves, and
`DEVFLOW_PROVIDER_API_KEY` fails loud on the provider path. Restating a count in prose beside
a per-row `failure_visibility` field that already answers it is the checked-in-figure rot
class, so the prose now points at the field instead of racing it.

Scope of that commit: **one JSON string**. The `policy` field is not rendered into the
generated region, so the advisory's bytes and its banner sha are unchanged — verified by
`--check` reporting a match without a regeneration. Re-verified after it:
`--check`, `--audit`, `python3 lib/test/regenerate-artifacts.py` (all artifacts reconciled,
exit 0), `lib/test/run-module.sh tier1-rename-migration` (191 passed, 0 failed), and
`ruff check` over every tracked `.py`. CI is re-running the full suite on it.

PR #1019, opened 2026-07-31
fix: resolve $REVIEW_BUNDLE and comment-suffixed aliases in the pin-corpus bundle extractor (#1008)

### LINKED ISSUE #1008: $REVIEW_BUNDLE is unresolvable to the bundle-member extractor, making two pins permanently undeclarable
## Problem

Issue #956 taught the pin lint to resolve a runtime bundle variable to the member files its builder call concatenates, so a pin whose target is a bundle is editable like any other. That resolution does not work for `$REVIEW_BUNDLE`.

Measured against the extractor:

| bundle variable | members resolved |
|---|---|
| `$MAXI_BUNDLE` | 9 |
| `$CREATE_ISSUE_BUNDLE` | 10 |
| **`$REVIEW_BUNDLE`** | **0** |
| any alias (`ST_RAF`, `ST_REV`, `SP_REVIEW`, `CI_SKILL_332`) | 0 |

`$REVIEW_BUNDLE` is built through a `"${_review_members[@]}"` array rather than the shape the extractor models, so it resolves to nothing.

## Consequence

Two pins are **permanently undeclarable** — a `# structural-pin-ok:` declaration on them is rejected with `typed structural declaration target cannot be inspected`:

- `lib/test/run.sh:1257` — `/../../scripts/config-get.sh .prflow_review.verdict_severity_threshold critical`
- `lib/test/run.sh:33775` — `cat .prflow/tmp/review/<slug>/<run-id>/diff.patch | …`

Confirmed end to end, not by code reading: a declaration was added and committed in a throwaway worktree, and the gate still reported

```
MUTATION-ROUTING	lib/test/run.sh:1257	assert_pin_unique	…	typed structural declaration target cannot be inspected
```

`CONTRIBUTING.md` already records that this class *"is still unfixable"* — this issue is about removing that residual for the review bundle specifically.

## Note on the alias cases

Three further sites (`lib/test/run.sh:1256`, `:1269`, `:14120`) are blocked only because they target an **alias** of a resolvable bundle. Those could be unblocked by de-aliasing at the call site, independently of modelling the array build shape. The two above cannot.

## Desired outcome

Model the `"${_review_members[@]}"` array build shape in the bundle-member extractor so `$REVIEW_BUNDLE` resolves, or record why it cannot and what a contributor should do when they need to touch one of those two pins.

Surfaced by PR #1005 (issue #1002).

---

## Audit addendum (2026-07-31, post-rename tree)

Re-measured against `resolve_bundle_targets()` on `main` at `fda85ca7`. The defect is **live**, and it is **two distinct causes**, not one:

| bundle variable | members resolved |
|---|---|
| `$MAXI_BUNDLE` | 9 |
| `$MAXI_SKILL` (alias, no trailing comment) | 9 |
| `$CREATE_ISSUE_BUNDLE` | 10 |
| **`$REVIEW_BUNDLE`** | **unresolved — name absent from the map** |
| `$ST_REV`, `$SP_REVIEW` (aliases of `$REVIEW_BUNDLE`) | unresolved |
| `$ST_RAF`, `$CI_SKILL_332` (aliases of bundles that DO resolve) | unresolved |

- **Cause A — the stem-loop build body.** `"${_review_members[@]}"` *is* modelled by `_BUNDLE_ARRAY_EXPANSION_RE`, and the two resolving bundles are array-built too. What differs is the loop **body**: the resolving ones append the loop variable itself, while the review bundle iterates a stem list and appends a derived path.
- **Cause B — comment-suffixed aliases.** An alias whose assignment carries a trailing comment does not resolve even when its source bundle does. `$MAXI_SKILL` (no trailing comment) resolves; `$ST_RAF` and `$CI_SKILL_332` do not. This is independent of Cause A and cheaper to fix.

**The fork in "Desired outcome" is closed: model the shape.** Do not take the "record why it cannot" arm — that arm existed because the cause was unknown; it is now measured, and Cause B in particular is a small extractor fix. If Cause A proves genuinely intractable after an attempt, record that with the evidence and fix Cause B regardless.

**Implementer hazards.** Verify a fix by driving `resolve_bundle_targets()` and asserting the resolved member set — **never** by asserting source text, which is an undeclared wording-only pin (`#375`/`#666`/`#810`) that passes locally and goes RED in CI. Any new `lib/test/test_*.py` enters the pin-gate population. `pin-corpus-lint.py` is not in the cloud-writer contract, so no artifact regeneration is required.

The01Geek:
Closes #1008.

## What was wrong

`resolve_bundle_targets()` in `lib/test/pin-corpus-lint.py` maps a runtime bundle variable back to the member files its builder call concatenates (issue #956), which is what lets a `# structural-pin-ok:` declaration on a bundle-targeted pin be inspected at all. It did not work for `$REVIEW_BUNDLE`, leaving two pins permanently undeclarable — and because the gate classifies any site whose lines land in the diff, the whole logical line was frozen.

The issue's audit addendum measured **two independent causes**, and both are fixed here.

**Cause A — the stem-loop build body.** `lib/test/run.sh` builds the review bundle by iterating a word-list variable and appending an interpolated *path template* per stem:

```bash
_review_members=("$REVIEW_ROOT")
for _s in $REVIEW_PHASE_STEMS; do
  _review_members+=("$LIB/../skills/review/phases/${_s}.md")
done
```

The old grammar modeled only an append of the loop variable *itself*, and deliberately **poisoned** any other body shape rather than under-resolving it — correct, but it meant `$REVIEW_BUNDLE` resolved to nothing at all. The template shape is now modeled: the loop's word list resolves to literal stems (from literal words and/or already-known word-list variables), each stem is substituted into the template, and the result is re-resolved through the ordinary member-word grammar. Stems are restricted to `[A-Za-z0-9._-]` before substitution, so none can introduce a quote, glob, or expansion that changes how the substituted text tokenizes.

**Cause B — comment-suffixed aliases.** `ST_RAF="$MAXI_BUNDLE"   # #530: …` did not resolve even though `$MAXI_BUNDLE` did: the trailing comment is part of the right-hand side `_ASSIGNMENT_RE` captures, so the whole-token variable match failed. The comment is now stripped — gated on what precedes it being a whole-token variable reference, so a right-hand side that merely *contains* a `#` is never truncated into a false alias.

## Measured effect

Driving `resolve_bundle_targets()` over this repository's own `lib/test/run.sh`:

| bundle variable | before | after |
|---|---|---|
| `$MAXI_BUNDLE` | 9 | 9 |
| `$CREATE_ISSUE_BUNDLE` | 10 | 10 |
| **`$REVIEW_BUNDLE`** | **unresolved** | **10** |
| `$ST_REV`, `$SP_REVIEW` (aliases of the review bundle) | unresolved | 10 |
| `$ST_RAF` (comment-suffixed alias of `$MAXI_BUNDLE`) | unresolved | 9 |
| `$CI_SKILL_332` (comment-suffixed alias of the create-issue bundle) | unresolved | 10 |
| `$IMPL_SKILL_BUNDLE` (a second stem-loop bundle) | unresolved | 6 |

The review bundle resolves to exactly its root plus the nine `skills/review/phases/*.md` files `REVIEW_PHASE_STEMS` names — every one present on disk, and nothing else in that directory.

## End-to-end proof of the unfreeze

Reproduced the issue's own experiment in a throwaway commit — the two named pins carrying a `# structural-pin-ok:` declaration, then the gate run against each extractor:

- **pre-fix extractor** (`origin/main`'s `pin-corpus-lint.py`, same declarations): `rc=3`, with exactly the two findings the issue quotes —
  ```
  MUTATION-ROUTING	lib/test/run.sh:1257	assert_pin_unique	/../../scripts/config-get.sh .prflow_review.verdict_severity_threshold critical	typed structural declaration target cannot be inspected
  MUTATION-ROUTING	lib/test/run.sh:33813	assert_pin_unique	cat .prflow/tmp/review/<slug>/<run-id>/diff.patch | …/scripts/stale-prose-lint.py --rev HEAD	typed structural declaration target cannot be inspected
  ```
- **post-fix extractor**, same declarations: `rc=0`.

The probe commit was dropped; this PR adds no declarations. It removes the *residual* that made adding one impossible.

## Verification

Behavioural throughout — every new test drives `resolve_bundle_targets()` / `_typed_pin_inspection_error()` and asserts the resolved member set or the refusal. No test asserts source text.

`lib/test/test_pin_corpus_lint.py` gains `BundleStemLoopAndAliasResolution1008Tests` (13 tests) and the one #956 test that pinned the old limitation is re-aimed at the shape that is *still* unmodeled:

- **positive** — a stem-loop bundle resolves to root plus one member per stem, composed across two word-list variables in list order; the typed declaration on it is inspectable; a bare `$_s` reference resolves like `${_s}`; a comment-suffixed alias resolves to its source bundle and its pin is inspectable.
- **negative controls** (the widening is resolution, not amnesty) — a sibling phase file no stem names is *not* a member; an unmodeled stem list (command substitution, a path, a glob word, an unknown list variable, empty) leaves the bundle unresolved; a reassigned stem list is ambiguous; an unresolvable template stays unresolved; a basename skip is not composed with a template; an unmodeled body statement still poisons; a missing templated member fails closed with `FileNotFoundError`; a `#` that is not a trailing comment is never read as an alias.

Commands run at the desk:

- `lib/test/run-module.sh harness-python-guards` → `38 passed, 0 failed` (module floor unchanged — the sharded driver folds the Python file into one assertion by contract, so the assertion tally does not move with the test count)
- `python3 lib/test/test_pin_corpus_lint.py` → `Ran 203 tests … OK`
- `python3 lib/test/pin-corpus-lint.py mutation-routing-worktree .` → `rc=0`
- `python3 lib/test/regenerate-artifacts.py` → `all artifacts reconciled — exit 0`, nothing to commit
- `git ls-files '*.py' | xargs -r ruff check` → `All checks passed!`
- full `lib/test/run.sh` — see the comment below

No changeset: `lib/test/**` plus a contributor-guide paragraph is internal surface.


The01Geek:
**Verification evidence**

Full local suite at `1fc43b52`:

```
14663 passed, 0 failed
```

No skip clause and zero `NOTE ` emits, so the skipped tally is 0 (the `#456` renderer appends `, K skipped` plus one line per skip whenever K > 0).

CI on the same head — all seven checks green, including the required `lib + python tests`:

| check | result |
|---|---|
| `lib + python tests` (required) | pass |
| `lint (shellcheck + actionlint + ruff)` | pass |
| `shard (modules-large)` | pass |
| `shard (modules-pin)` | pass |
| `shard (modules-rest)` | pass |
| `shard (monolith)` | pass |
| `shard (python-pool)` | pass |

`shard (modules-pin)` is the shard that carries `test_pin_corpus_lint.py`; the local run logged `test_pin_corpus_lint.py: executed 203 test(s) across 4 concurrent worker(s) (203 enumerated)`, so the new class ran rather than being enumerated away.

**Cost of the added parse.** The word-list map is one extra pass over the source, memoized on the presented bytes like the two parses beside it. On this repository's own 3 MB `lib/test/run.sh`: 0.023 s cold, 0.000 s warm — against 0.056 s for the bundle resolution it feeds.

**One boundary deliberately left untested.** `_bundle_loop_var_pattern`'s negative lookahead stops `$_s` from matching the prefix of a longer name (`$_sx`). Both directions of that boundary are fail-closed — an over-greedy match would substitute into a path no file occupies, which surfaces as the existing `FileNotFoundError` refusal rather than a false pass — so it is left to the `${…}`-braced and bare-reference tests that already exercise the pattern, rather than adding a unit test on a private regex.

PR #1031, opened 2026-07-31
fix(install): gitignore preserved-artifact sidecars, and preview language detection against the real tree

### LINKED ISSUE #970: install.sh: preserved-artifact sidecars land untracked in .github/ with no ignore rule, so git add -A can commit them
## What

`install.sh`'s upgrade path writes `<path>.prflow-new` sidecars for artifacts it preserves (`modified`, `unverified`, `unreadable`). Those sidecars land **untracked inside the consumer's `.github/`** and are **not covered by any ignore rule** — `.prflow/.gitignore` carries only `/tmp/` and `/vendor/` — and, per its own comment in `install.sh`, its patterns are relative to `.prflow/`, so it can **never** cover `.github/*.prflow-new`.

## Why it matters

A later `git add -A` — including one inside a `/prflow:implement` run — sweeps them into an unrelated PR. Observed sizes on a real consumer: `devflow.yml.devflow-new` at 1448 lines and `devflow-implement.yml.devflow-new` at 1762 lines, plus a whole `vendor-plugin.devflow-new/` directory.

## Evidence

Found during the first upgrade of a real (private) consumer repo. That repo was pre-manifest, so four of six managed artifacts classified `unverified` and each got a sidecar. The upgrade itself behaved correctly — the sidecars are the intended preservation mechanism — the defect is only that nothing stops them being committed.

The case was avoided there by adopting all four sidecars, which is not the general case: a consumer who wants to keep their own version leaves the sidecar in place indefinitely.

## Suggested direction

Not prescriptive — options include an ignore rule the installer manages (it already manages the `/vendor/` line via `manage_vendor_gitignore`), writing sidecars outside the working tree, or naming them so an existing ignore pattern catches them. Whatever is chosen should survive a consumer who never resolves the sidecar.

## Related

Surfaced by the consumer upgrade path (PR #959).

---

> **Re-audited against `main` after the PRFlow rename (Tiers 1–3): the defect is unchanged and still live.** Only literals moved, and they are corrected inline above — except in **Evidence**, where the filenames are left as observed. That upgrade really did write `.devflow-new` sidecars; rewriting a past-time measurement would falsify it. Today the same run writes `.prflow-new`.

### LINKED ISSUE #971: install.sh: the upgrade preview can understate what --apply writes, because language detection runs against the sandbox rather than the real tree
## What

`install.sh`'s upgrade preview builds a sandbox containing only `.claude-plugin`, `.github`, `.prflow`, `.devflow` and `.claude/plugins`, then runs the real apply against it and diffs the result. Because the sandbox omits the rest of the repo, `detect-project-tools` sees no `composer.json` / `package.json` / `docker-compose*` and reports **"no known language markers detected"**, while the same step on the real tree detects the project's actual languages.

## Why it matters

The header promises the preview is *"a unified diff of every byte it would change."* For a consumer whose `config.json` does **not** already cover its detected languages, the dry run would **understate** what `--apply` writes — the user approves a diff that is not the whole change.

## Evidence

Observed on a real consumer upgrade:

```
dry-run : devflow-detect: no known language markers detected; config.json left unchanged.
--apply : devflow-detect: detected: docker node php — config.json already covers them; no changes.
```

Both outcomes happened to be no-ops there, because that repo's config already covered its languages — so the diff stayed faithful **by luck, not by construction**. A consumer without that coverage is not protected.

## Suggested direction

Not prescriptive — options include seeding the sandbox with the marker files `detect-project-tools` reads, running detection against the real tree and injecting the result, or narrowing the header's claim to the scopes the preview actually covers. The first two preserve the promise; the third makes it honest.

## Related

Surfaced by the consumer upgrade path (PR #959). Sibling of #970.

---

> **Re-audited against `main` after the PRFlow rename (Tiers 1–3): the defect is unchanged and still live.** Only literals moved, and they are corrected inline above — the sandbox scope list now also carries `.prflow`. The quoted `install.sh` header promise is verbatim on current `main`.

The01Geek:
Fixes two sibling defects in `install.sh`'s consumer upgrade path. They ship together because both add assertions to the **exact**-pinned `installer-wiring` module (`lib/test/test_module_runner.py` asserts `Module installer-wiring: {floor} passed, 0 failed` — equality, not a floor), so two parallel PRs would collide by construction. They also touch the same file and came from the same incident: a real consumer's first upgrade (PR #959).

Closes #970
Closes #971

---

## #970 — preserved-artifact sidecars were committable

When the upgrade path preserves a `modified` / `unverified` / `unreadable` artifact it writes DevFlow's version beside it as `<path>.prflow-new`. That sidecar is **untracked** and lands inside the consumer's own `.github/` (or `.claude-plugin/`), which `.prflow/.gitignore` cannot reach — its patterns are relative to `.prflow/`. So the next `git add -A`, including one inside a `/prflow:implement` run, staged it. On the real consumer that meant a 1448-line and a 1762-line sidecar plus a whole sidecar *directory*.

**Design.** `install.sh` gains `manage_sidecar_gitignore`, which appends one managed block to the **repository-root** `.gitignore`:

```gitignore
# PRFlow install.sh: preserved-artifact sidecars (never commit these)
*.prflow-new
*.devflow-new
```

- **Root `.gitignore`, not `.prflow/.gitignore`.** The issue points at `manage_vendor_gitignore` as precedent. It is precedent for *the installer managing an ignore rule at all* — not for where this rule goes: that function's own comment records that patterns in `.prflow/.gitignore` are relative to `.prflow/`, and a sidecar never lands under it.
- **Unanchored basename patterns.** Git matches them at any depth and against a **directory** as well as a file, which the composite-action case needs — a preserved `.github/actions/vendor-plugin` is copied with `cp -R` to a whole `vendor-plugin.prflow-new/` tree.
- **A standing rule, not a cleanup.** The issue's stated constraint is that it survives a consumer who never resolves the sidecar. Keeping your own version of an artifact *means* leaving the sidecar there indefinitely, so anything shaped like "resolve it, then we tidy up" never reaches that case.
- **The superseded `*.devflow-new` spelling is covered too.** Sidecars written before the `.devflow` → `.prflow` rename are precisely the artifacts observed sitting untracked on the real consumer, and nothing rewrites a file already on disk.

**Rejected:** `.git/info/exclude`. It would survive and it needs no consumer file, but it is a write **outside every preview scope** — invisible in the dry run's "diff of every byte it would change". Fixing #970 by creating a fresh instance of #971's own defect class is not a fix.

**The consumer's bytes are never rewritten.** The presence check is pure bash, not `grep -qxF`: it *decides whether a line is appended*, and grep is not preflight-guaranteed, so a host without it would append the block on every run (CLAUDE.md guard-class 2 — and the rule applies to code being written, not only code being reviewed). The append also leads with a newline when the file is non-empty, which repairs a `.gitignore` whose last line carries no trailing newline; without it the append would silently rewrite that pattern into `dist*.prflow-new`. A rule the consumer already carries is not duplicated, and re-runs are byte-identical.

**A `.gitignore` that is a symlink is refused outright, and a non-regular path after it.** `>>` follows a link and writes to its *target*, which an absolute link can put anywhere on the filesystem — outside the repository entirely. No link shape is visible to an `-e`/`-f` pair, because those tests dereference as well: a **live** link to a regular file reads `-e` true and `-f` **true**, a **dangling** one reads both false (and the append would *create* the target), and a link to a directory reads `-e` true, `-f` false. `[ -L ]` is the one test that does not dereference, so it runs first and unconditionally. This is deliberately conservative — a link pointing *inside* the repo would have been safe — but no reachable input can now make the function write outside the tree it was pointed at, and the refusal hands the consumer both patterns and the reason. `devflow_build_preview` keeps mirroring the *shape* (`cp -P`), which is what makes the sandbox refuse for the same reason the real tree does; dereferencing there would make the sandbox copy writable and have the preview advertise a MODIFY the apply refuses. A failed append surfaces its captured stderr, the way `scaffold-config.sh`'s `rewrite_config_if_changed` reports a failed `mv`.

## #971 — the preview could understate what `--apply` writes

The dry run builds a sandbox holding only `.claude-plugin`, `.github`, `.prflow` and `.claude/plugins`, then runs the real apply against it. Language auto-detection therefore scanned a tree with no `package.json` / `composer.json` / `docker-compose*` and reported `no known language markers detected`, while the same step under `--apply` saw the real tree and merged the project's toolchain into `config.json`. Both were no-ops on the observed consumer only because its config already covered its languages — a consumer without that coverage approves a diff that is not the whole change, against the header's promise of "a unified diff of every byte it would change".

**Design.** `scripts/detect-project-tools.sh` and `scripts/scaffold-config.sh` take an optional second argument — the tree to **scan**, separate from the tree they **write**. It defaults to the target root (an empty value also selects the default), so the `/prflow:init` path and `install.sh --apply` are byte-for-byte unchanged; only `install.sh`'s dry run passes a different value, the real repository root, threaded as `devflow_apply_all`'s 4th positional.

**Chosen over seeding the sandbox with marker files.** Seeding is a *reconstruction* of what the detector reads, and it would have to mirror both `tool-presets.json`'s marker globs and the lockfile names hardcoded in `find_node_lockfile`. Any later reader added to the detector that the seeder does not mirror re-opens this exact bug, silently. The scan root has no mirror to drift. Option 3 (narrowing the header claim) was unnecessary once the promise could be made true.

**Evidence that the preview still writes nothing to the real tree.** The scan root reaches exactly three read sites — `marker_present`'s `find`, `find_node_lockfile`, and the `composer.json` probe. `CONFIG`, the one path the detector writes, is still derived from `TARGET_ROOT`; `prflow_state_dir` is likewise given `TARGET_ROOT`. Asserted two ways: a whole-tree byte snapshot of the previewed consumer is unchanged across the dry run, and — at the detector's own executable surface — running it with a separate scan root merges the *scanned* tree's toolchain into the *target* repo's config while leaving the scanned tree byte-identical (no config, no state directory, no byte).

The repository-root `.gitignore` also joins `DEVFLOW_PREVIEW_SCOPES` and `devflow_build_preview`, so #970's new write is previewed like every other one.

---

## Coverage

`installer-wiring` floor **239 → 264** (bumped in both coupled sites: `scripts/workflow-flight-recorder-registry.json` and the `lib/test/run.sh` call-site operand).

**Scenario 22 (#970)** is driven over fixture consumers with a **real git repository** rather than the `.git`-shaped directory the other scenarios use, because the defect is about git's view of the tree. The load-bearing arm runs the actual failing operation: after an upgrade that leaves a file sidecar as well as a directory sidecar, `git add -A` stages neither — with a negative control that strips the managed block and requires the same recipe to stage them, so the arm is not vacuous. Around it, the append's adversarial shape matrix: a `.gitignore` with no trailing newline, one the consumer already wrote the rule into, a re-apply, one that is a **directory**, one that is a **symlink** in the live and dangling shape alike (each asserted over the bytes of the file it points at outside the repository, on the apply path and again on the dry-run path), the preview's matching refusal for a non-regular shape, the dry run that must render a normal append as a diff while performing none of it, the **append-failure** branch, and the unterminated-final-line rule.

**Scenario 23 (#971)** runs twin fixtures that gain their language markers *after* the install, so the interesting path is reachable: the previewed and applied `devflow-detect:` verdicts must be the same line; that shared verdict must actually name the detected languages (so two empty reports cannot satisfy the comparison); the `config.json` must appear in the dry-run diff with the merged tool entries in its body; and the tree must stay byte-identical. A negative control drops the preview's scan-root argument and reproduces the understated `no known language markers detected`. Two further arms pin the read-only direction at `detect-project-tools.sh`'s own surface, plus the single-argument back-compat form.

## Verification

- `lib/test/run-module.sh installer-wiring` → `Module installer-wiring: 259 passed, 0 failed`
- `python3 lib/test/test_module_runner.py` — the exact-floor coupling check
- `python3 lib/test/regenerate-artifacts.py` → `all artifacts reconciled — exit 0` (the `#591` coverage-map ratchet needed a row for label `971`)
- `git diff origin/main...HEAD | python3 scripts/stale-prose-lint.py --rev HEAD` → clean
- `shellcheck --severity=warning -e SC1091` over the tracked non-`lib/test` shell set, and `--extended-analysis=false` over `lib/test/run.sh`
- `git ls-files '*.py' | xargs ruff check`
- the full `lib/test/run.sh` suite

---

## Review round 1 — three Important findings, all closed

**1 — a LIVE symlink `.gitignore` wrote outside the repository.** The first guard was `{ -e && ! -f } || { -L && ! -e }`, and for a link to a regular file **both** `-e` and `-f` are true because both dereference, so neither arm fired and the append went to the link's target. Reproduced before fixing:

```
ln -s "$PWD/outside/target" repo/.gitignore
-e=T  -f=T  -L=T   -> guard does NOT fire -> printf >> appended to outside/target
```

Worse in the preview path, which is why it was rated Important: `devflow_build_preview` copies with `cp -P`, so an absolute-target link gives the sandbox the **same real file** — the sandbox apply mutated a file outside the throwaway copy, and `walk()` rendered no diff at all because both sides resolve to one file. The guard is now `[ -L ] || { [ -e ] && [ ! -f ]; }`, with the symlink arm first and unconditional; the refusal keeps its actionable breadcrumb (now naming both patterns and the reason) and still exits 0. `cp -P` stays deliberately — the refusal makes it inert, and it is what keeps preview and apply refusing for the same reason.

**2 — the append-failure `else` branch was entirely uncovered.** Now driven at function level with a **deleted working directory**, so the relative `>>` fails with ENOENT: deterministic and **root-immune**, the same requirement the manifest-write-failure arm satisfies by making its target a directory (a directory is unavailable here — the guard now refuses one before the append is reached). Asserts exit 0, the warning branch rather than the success line, the captured cause, and the by-hand remedy.

**3 — `devflow_gitignore_carries`' `|| [ -n "$line" ]` arm was not discriminatingly covered.** The new fixture writes `printf 'node_modules/\n*.prflow-new'` with **no** final newline, so the rule *is* the unterminated last line — the one shape whose detection depends on that arm — and asserts the count stays 1.

### Mutation evidence

Each new arm proven load-bearing by planting the defect in the real `install.sh`, running the module, then restoring:

| Planted defect | Result |
| --- | --- |
| revert to the pre-fix `-e`/`-f` guard | `262 passed, 2 failed` — both live-symlink arms |
| invert the append-result branch | `261 passed, 3 failed` — both new arms plus the success-line arm |
| drop the `\|\| [ -n "$line" ]` arm | `263 passed, 1 failed` — the unterminated-rule arm |
| restored | `264 passed, 0 failed` |

**Suggestion taken:** the absolute "and nothing else" / "holding only [four subtrees]" phrasing is reconciled at the three named sites, plus the matching sentence in `docs/install.md`, to the generic form the sibling `scaffold-config.sh` comment already uses — this PR also copies `.gitignore` into the sandbox.

## Note for the reviewer, not fixed here

`lib/test/run.sh` already carries ~20 assertions labelled `#970` for the **conditional GitHub Release publication** work (`publish-release.sh --release minor-major` / `--emit-bump-to`). Issue 970 is the installer one this PR fixes, and 970 is not a PR number; that commit landed *after* the issue existed, so the label is a genuine pre-existing mislabel. It is left as found — sweeping it would touch `CHANGELOG.md` and every one of those assertion names. The shared coverage-map row for label `970` now records both blocks and the mislabel explicitly, so a reader is not misled about what the label covers.


The01Geek:
Requesting a review before merge. This is the highest consumer-blast-radius change in the current batch: it appends a managed block to the **repository-root `.gitignore`** — a file the consumer owns — and it changes what the upgrade preview reports.

Two things worth the reviewer's attention specifically:

- The `.gitignore` write is guarded against a **dangling symlink** at that path (neither `-e` nor `-f`, and an append would create the link's target outside the repo). Please check that refusal holds on every shape, including a symlink to a path inside `.git/`, and a `.gitignore` that is a directory.
- The preview fix separates the tree that is **scanned** from the tree that is **written** (`detect-project-tools.sh` gained a second positional). The claim is that no write path derives from the scan root. Please verify that independently rather than from the tests.

Also note a pre-existing mislabel recorded but deliberately not swept here: `lib/test/run.sh` already carries ~20 assertions labelled `#970` belonging to unrelated Release-publication work.

/prflow:review

The01Geek:
All three 🟠 Important findings verified closed on `285af6bd`.

I re-ran the original reproduction against the new guard: with a **live** symlink to a regular file it now refuses (`-L` first and unconditional — the only test that does not dereference), and the outside target is byte-unchanged. Findings 2 and 3 are backed by a mutation table where each planted defect turns **only** its own arms red.

Keeping `cp -P` in `devflow_build_preview` is the right call and the reasoning is recorded in-code: dereferencing there would make the sandbox copy writable, so the preview would advertise a `MODIFY` the real apply refuses — breaking the sandbox/real agreement in the dangerous direction.

Floor 259→264 in both coupled sites. Merging.

Issue #142, opened 2026-06-26
Internalize the two runtime-dispatched superpowers skills + remove all companion-plugin dependencies (seam 3 of #139)


The01Geek:
## Problem Statement

Follow-up to #139 (seam 3 of 3, the final seam). The parent issue internalizes every external Anthropic asset DevFlow's engine dispatches as a first-party DevFlow file so the retrospective-weekly self-improvement loop can evaluate and improve it. The first PR internalized the `feature-dev` agents; the second the `pr-review-toolkit` review roster. This issue covers the **`superpowers`** seam **and the final dependency/marketplace cleanup**: the **two** `superpowers` skills DevFlow's engine dispatches at runtime, plus removing the last companion-plugin dependency so DevFlow has zero companion-plugin install dependencies. (`superpowers:writing-skills` is **not** internalized — see below.)

## Current Behavior

`skills/review/SKILL.md` dispatches `superpowers:requesting-code-review` (the final-pass reviewer, via a `general-purpose` Task), `skills/review-and-fix/SKILL.md` references `superpowers:receiving-code-review` (fix-loop principles), and `CLAUDE.md` plus several `SKILL.md` bodies invoke the `superpowers:writing-skills` contributor convention. `skills/retrospective-weekly/SKILL.md` carries a single optional `superpowers:using-git-worktrees` reference. After this seam lands, `pr-review-toolkit` and `feature-dev` are already internalized (seams 1 and 2), so removing `superpowers` empties the dependency set entirely.

## Desired Behavior

`requesting-code-review` and `receiving-code-review` are vendored in as first-party DevFlow skills under `skills/`, the engine dispatches the internalized identifiers, no engine file references a `superpowers:` identifier for those two skills, and **all** companion-plugin dependencies are removed from `.claude-plugin/plugin.json` with `.claude-plugin/marketplace.json`'s `allowCrossMarketplaceDependenciesOn` emptied — so DevFlow has zero companion-plugin install dependencies and the documented `dependency-unsatisfied` Errors-tab gotcha disappears.

[...]

PR #1037, opened 2026-07-31
feat: optimize skill descriptions for natural-language auto-triggering

The01Geek:
## Why

Users report forgetting to type the slash commands. The skills should be picked up automatically when someone just describes what they want.

The clearest instance was `implement`, whose description stated its trigger as **the literal presence of `/prflow:implement` in the message** — a condition that is false for any natural request like "implement issue #123", so the description was actively arguing against the behaviour we want. It also carried a workflow summary ("runs the full 4-phase lifecycle"), which the skill-authoring guidance warns against: agents follow the summary instead of reading the skill body.

## What changed

Eleven `description` fields — `implement`, `review`, `review-and-fix`, `create-issue`, and the seven docs skills. Each now leads with concrete triggering phrasings in the user's own vocabulary, and the overlapping skills name the sibling to use instead.

- **`review` / `review-and-fix`** split on an *observable* predicate rather than an inferred one: an unqualified "review my PR" defaults to `review`; `review-and-fix` requires explicit fix intent. The previous wording ("whenever the request implies edits should be applied") asked the agent to infer intent that users rarely state.
- **`create-issue`** also triggers on writing up an implementation plan, keyed on the work being *tracked rather than built*, and routes designing the work itself to a brainstorming or planning skill.
- **`docs-verify`** surfaces its `--report-only` mode, where the skill acts as a codebase exploration agent producing a read-only map from the internal docs. Gated on user-stated docs framing, so plain "how does X work" stays ordinary code exploration.
- **docs family** gained the reciprocal "narrower than `prflow:docs`" boundaries that were previously one-way, so the parent no longer silently absorbs its children's prompts. `docs-release-notes` picked up the `changeset` keyword — the mechanism this repo actually uses.

Slash-command dispatch is unaffected: the cloud tier matches the comment body in the workflow trigger and `resolve-implement-trigger.sh`, and both the synthesised cloud prompt and the local command resolve by skill **name**. No dispatch path reads the description; verified before removing the clause.

Regenerates `scripts/devflow-cloud-writer-contract.json`, which SHA-256-pins 8 of the edited files.

## Verification

Routing was checked with fresh-context subagents over should-trigger and near-miss should-not-trigger prompts, 3 reps each, all converging on identical output. Near-misses hold: "plan out this feature" → a brainstorming/planning skill, "map out how caching works" / "how does the retry logic work?" → Explore, "what does config-get.sh do?" → no skill.

Two real defects were caught this way and fixed in-branch: a ` #` in an unquoted YAML scalar silently truncated two descriptions mid-sentence (YAML comment syntax), and an intermediate `docs-verify` wording regressed plain "how does X work" into the docs skill.

**Honest limit on the evidence.** The baseline did not fail — under the *old* descriptions these harnesses also routed correctly, so they measure *discrimination between descriptions*, not the attentional failure users actually hit in a long session with a crowded skill roster. These results show the boundaries are clean and nothing regressed; they are not evidence that auto-triggering fires more often in practice. That only shows up in real use.

## Follow-up fix: foreign-plugin ids in a shipped description

`create-issue`'s new routing sentence originally read *"use `superpowers:brainstorming` or `superpowers:writing-plans`"*. That turned the `#142` guard RED, correctly: these `SKILL.md` files ship into third-party consumer repos that may not have the `superpowers` plugin installed at all, so a shipped description named ids that resolve to nothing there.

Fixed by rewording to name the *class* of skill rather than two specific foreign ids:

> before: `… For exploring or designing the work itself, use superpowers:brainstorming or superpowers:writing-plans; this skill records the outcome as an issue.`
> after: `… For exploring or designing the work itself, reach for a brainstorming or planning skill first; this skill records the outcome as an issue.`

The design/explore-vs-record-as-issue distinction is preserved in substance; only the hand-off target is generalised. **The guard was neither weakened nor exempted** — no assertion edited, no exemption added, no path added to its exception list; the scan it runs is now simply empty. Every remaining cross-referenced id in the eleven descriptions is a first-party `prflow:` skill that exists in this repo (checked mechanically, 11/11 resolve).

Frontmatter stays well inside the 1,536-char cap (`create-issue` is 578 chars).

## Not in scope

`retrospective-weekly` is still a pure workflow summary and `pr-description` was not rewritten, so the shipped set is stylistically inconsistent. Worth a follow-up.

Writing-skills evidence: `superpowers:writing-skills` invoked before editing; RED/GREEN loop run with subagent pressure scenarios as described above.

Writing-skills evidence (follow-up fix): `superpowers:writing-skills` invoked via the Skill tool before touching `skills/create-issue/SKILL.md`. RED — the `#142` tracked scan returned `skills/create-issue/SKILL.md` with the two offending literals. GREEN — after the reword the same scan returns empty. Wording micro-tested per the skill's "micro-test before full scenarios" step with fresh-context subagents on a consumer-style roster carrying **no** brainstorming/planning skill: no-guidance control → NONE 5/5 (soft design prompt) and 5/5 (hard "spec out … land on a design" prompt), so the control does not exhibit the failure and the sentence is not load-bearing for suppression — recorded rather than papered over. New wording → record-intent ("we should track that") selects `create-issue` 5/5, design-intent selects NONE 5/5, i.e. no regression in either direction.


The01Geek:
Requesting review. Batch 2 of a sequenced merge — #1024 and #1036 are already in; #1042 and #1043 follow.

Rewrites 11 skill frontmatter `description` fields for natural-language auto-triggering, so a user's ordinary phrasing selects the right skill.

Worth the reviewer's attention:

- **Descriptions are the discovery surface, so a regression here is silent.** A description that stops matching does not error — the skill simply never triggers. Please assess whether each rewrite widens triggering without pulling in requests that belong to a *different* skill; over-triggering is as costly as under-triggering and harder to notice.
- **`#142` is load-bearing and was this PR's original failure.** An earlier revision named `superpowers:brainstorming` / `superpowers:writing-plans` as bare namespaced ids in a shipped description. These `SKILL.md` files ship into consumer repos that may not have that plugin installed, so the ids would dangle. The fix reworded to name the *class* of skill instead. Confirm no exemption was added and the guard is untouched.
- The frontmatter has a 1,536-character cap and there is a recorded history of a ` #`-in-unquoted-scalar truncation bug in this file set — worth a parse check rather than a read.

Merge history note: this branch absorbed `origin/main` after #1024/#1036 landed. The only conflict was the generated `scripts/devflow-cloud-writer-contract.json`, resolved by **regenerating**; the result is verifiably identical to neither parent (7 entries differ from main, 1 from the pre-merge copy), so neither side's work was dropped.

/prflow:review

The01Geek:
Review verdict on `09ec6c11` (current head): **APPROVE**, no findings. Merging as Batch 2.

PR #1091, opened 2026-08-01
Add a parallel full-suite verifier for agent runs

### LINKED ISSUE #1086: Add a parallel full-suite verifier for agent runs
## Problem Statement

PRFlow's autonomous implement and sweep agents need an in-environment final verification result before they claim completion. This repository gives those agents a serial complete-suite command even though CI already partitions the same test population into concurrent shards. Repeated final and recovery passes therefore spend long intervals on work whose tested partitioning already exists, then send the complete assertion stream back into model context.

Issue #1064 exposed the cost clearly. Its implement run reported complete-suite results at 18:12, 18:28, and 18:54 during one 114-minute agent invocation. The run spent 233 turns and read about 68 million cached input tokens.

## Current Behavior

CI runs the suite through a concurrent shard matrix. Agent runs invoke `lib/test/run.sh` as one serial process.

- Verified: `lib/test/run-shard.sh` says, “Shard dispatcher for the concurrent CI job matrix (issue #877).” It defines the shard population in `SHARD_NAMES`, captures shard output, and sends the capture to `lib/test/shard-tally.py`. Baseline `claim=shard-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=44c813fb90bb30d3627586428076e730486439b2`.
- Verified: `.github/workflows/ci.yml` contains “The suite is split across concurrent SHARD jobs (issue #877) so wall-clock is”. Green run [30715088867](https://github.com/The01Geek/prflow/actions/runs/30715088867) completed its shard steps in 4:00, 4:10, 3:51, 2:31, and 1:19 on separate runners.
- Verified: `.prflow/prompt-extensions/implement.md` says, “The final gate is preserved, and on the local/interactive tier it is parallelized.” Its cloud clause assigns `bash lib/test/run.sh` to the in-environment final obligation. The same section records that caller-side redirection and environment-assignment shapes are refused on cloud. Baseline `claim=cloud-command-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=c0ec4f8eace907303ba2732c03c648bd5ce053b9`.
- Verified: `.prflow/config.json` contains `"Bash(lib/test/run.sh:*)"` and `"Bash(lib/test/run-module.sh:*)"` in `.prflow.allowed_tools` and `.prflow_implement.allowed_tools`. It contains no direct parallel-suite command. Baseline `claim=cloud-command-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=c0ec4f8eace907303ba2732c03c648bd5ce053b9`.
- Run evidence: issue [#1064](https://github.com/The01Geek/prflow/issues/1064#issuecomment-5152453627) records suite milestones at 18:12, 18:28, and 18:54. The executed command `git show origin/prflow-telemetry:.prflow/logs/efficiency/pr-1069-30709454793-1.json` reported `duration_ms=6859173`, `num_turns=233`, `cost_usd=44.687623249999966`, `output_tokens=149125`, and `cache_read_input_tokens=68062792` for [run 30709454793](https://github.com/The01Geek/prflow/actions/runs/30709454793).

The CI timing is evidence for the partition. It does not establish CI's isolation, resource budget, and wall time for concurrent shards sharing one checkout.

## Desired Behavior

Agent final verification uses one direct executable, `lib/test/run-parallel.sh`. The executable derives the suite population from the existing shard dispatcher, runs that population concurrently inside the current checkout, recombines the existing tally protocol, retains complete logs under ignored local state, and emits a compact caller-facing result.

The helper owns environment assignments, redirects, background processes, capacity accounting, and aggregation internally. Cloud agents invoke one allowlisted leading token. Local agents invoke the coordinator through the existing `DEVFLOW_BASH` selection boundary.

Focused tests remain the iteration default. `lib/test/run.sh` remains the serial primitive used by the monolith shard and by the documented uncovered-surface fallback.

## User Impact

Autonomous implement and sweep runs receive the same tested partition used by CI without spelling cloud-denied composite commands. Maintainers get shorter final-gate wall time on capable runners, smaller model context, retained failure logs, and one consistent verification command across local and cloud agent tiers.

## Technical Context

- **Relevant files:** `lib/test/run-shard.sh` owns shard membership and dispatch. `lib/test/shard-tally.py` owns extraction and aggregate accounting. `.github/workflows/ci.yml` owns the multi-runner matrix. `.prflow/config.json` owns repository cloud grants. The final-gate contracts live in `.prflow/prompt-extensions/implement.md`, `.prflow/prompt-extensions/review-and-fix.md`, `.prflow/prompt-extensions/receiving-code-review.md`, and `CLAUDE.md`.
- **Verification-flight coexistence:** `scripts/verification-flight.py` is non-executing. It coordinates ownership and terminal evidence while the agent still launches the allowed test command. The coordinator command identity, compact summary, exit status, skip population, and retained-log root become the full-suite flight evidence.
- **Capacity and invocation boundaries:** `lib/test/module-harness.sh` already resolves `DEVFLOW_POOL_WIDTH` and carries a launch-window pending-signal replay guard. `CLAUDE.md` assigns local shell selection to `DEVFLOW_BASH`. The coordinator must compose with both contracts instead of creating an independent process-width policy and bypassing the local Bash selector.
- **Verified architecture:** `lib/test/run-shard.sh` contains “This script is what one shard job runs: it maps a shard name to its work,”. `lib/test/shard-tally.py` contains “Fail closed: any counted failure, any shard problem (missing/malformed tally,”. Baseline `claim=shard-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=44c813fb90bb30d3627586428076e730486439b2`.
- **Executed command-contract inventory:** a repo-wide whitespace-normalized search for `Bash(lib/test/run.sh:*)`, `bash lib/test/run.sh`, and the cloud final-gate sentence found active agent-policy consumers in `.prflow/config.json`, the three prompt extensions, `CLAUDE.md`, `docs/implement-skill.md`, and `docs/DEVFLOW_SYSTEM_OVERVIEW.md`. `README.md` and `CONTRIBUTING.md` describe the human contributor command and retain the serial primitive. `docs/install.md` uses the literal only as grammar evidence. `lib/test/run-shard.sh` and `lib/test/run.sh` remain implementation primitives. Immutable learning, retrospective, changeset, changelog, and pin-retirement records are evidence, not edit targets. Re-derive with `git ls-files -z | python3 -c 'import pathlib,re,sys; ns=["Bash(lib/test/run.sh:*)","bash lib/test/run.sh","The final full-suite obligation binds the cloud tier too"]; ps=sys.stdin.buffer.read().split(b"\\0"); [(print(n,p)) for n in ns for b in ps if b for p in [pathlib.Path(b.decode())] if p.is_file() and re.sub(r"\\s+"," ",n).strip() in re.sub(r"\\s+"," ",p.read_text(errors="replace"))]'`. Baseline `claim=cloud-command-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=c0ec4f8eace907303ba2732c03c648bd5ce053b9`.
- **Documentation drift:** `docs/DEVFLOW_SYSTEM_OVERVIEW.md` and `docs/implement-skill.md` accurately describe the existing shard architecture and serial in-run command. They contain no parallel in-run coordinator contract. Baseline `claim=documentation-contract revision=b1b86d34cb7199d3e1ea2d327673fbf8e7f4e23d identity=3ec3fdc75afd3c04386c974eef86e2d91737ae10`.
- **Grant timing:** `devflow-implement.yml` resolves repository grants from the default branch before it checks out the implementation branch. The grant added by this change is inert during its own implementing run. In-PR verification must use existing granted commands and code reads. A live cloud invocation is a post-merge proof.
- **Deployment scope:** this is repository-internal test infrastructure. The generic consumer-repository test command and shipped workflow set stay unchanged. The config schema, data model, GitHub required-check name, and review-tier CI policy stay unchanged.

## Acceptance Criteria

- [ ] `lib/test/run-parallel.sh` is executable and accepts the bare invocation `lib/test/run-parallel.sh` as its complete agent-facing command shape.
- [ ] The helper obtains its launch population from `lib/test/run-shard.sh --list-shards` and starts the returned population without a copied shard list.
- [ ] One invocation chooses a fresh writable run root under `.prflow/tmp`, falling back to `${TMPDIR:-/tmp}` when the checkout is read-only, and gives a private `TMPDIR`, tally directory, and process log to the shard process identified by a returned shard name. Unwritable and exhausted roots fail with a named diagnostic before a completion claim.
- [ ] A preflight-guaranteed `python3` probe establishes one positive total process budget, capped at eight, with `DEVFLOW_SUITE_PROCESS_BUDGET` as the positive-integer test and operator override. An absent, nonnumeric, and nonpositive probe fails closed. The scheduler reserves one slot for a normal shard and one through four slots for `python-pool`, exports that reservation as `DEVFLOW_POOL_WIDTH`, and never exceeds the total budget across top-level shards and nested pool workers.
- [ ] The returned shard processes overlap when the established budget is at least two, width one remains serial and complete, and the parent waits for the complete launched population before aggregation.
- [ ] HUP, INT, and TERM handling carries a pending-signal replay guard across each fork-to-PID-registration window, then terminates and reaps the launched population before the parent exits.
- [ ] An empty shard population, malformed shard name, launch failure, nonzero child, absent tally, malformed tally, and skip-detail disagreement produce a nonzero aggregate with a named diagnostic.
- [ ] A clean run emits lifecycle breadcrumbs, the combined pass/fail/skip result, the shard roster, and the retained-log root without replaying assertion logs.
- [ ] Aggregate failure and skip detail uses an enforcement cap of 20 entries per detail class. `lib/test/run-module.sh parallel-suite-runner` drives 25 synthetic entries and asserts 20 rendered detail lines plus a `5 omitted` line while the complete synthetic log remains readable under the retained run root.
- [ ] The helper derives its expected tally population from the returned shard list and passes only the current run's explicit tally directories to aggregation, so a stale sibling directory does not satisfy the current invocation.
- [ ] `.prflow.allowed_tools` and `.prflow_implement.allowed_tools` grant `Bash(lib/test/run-parallel.sh:*)` while preserving the existing serial and focused grants.
- [ ] The cloud final full-suite contract in the three repository prompt extensions and its `CLAUDE.md` mirror invokes `lib/test/run-parallel.sh` as a direct leading token. The local contract invokes the same coordinator through the documented `DEVFLOW_BASH` boundary. Focused iteration and the serial uncovered-surface fallback retain their current behavior.
- [ ] Final full-suite ownership still flows through `scripts/verification-flight.py`; its terminal evidence records the coordinator command identity, compact aggregate, exit status, skip population, and retained-log root.
- [ ] The local `Verification evidence:` marker records the coordinator's retained-log root and compact aggregate without a caller-side redirect.
- [ ] CI keeps its current shard matrix, aggregate required-check name, and multi-runner behavior. The new helper is added to CI's explicit `lib/test` ShellCheck list.
- [ ] `docs/DEVFLOW_SYSTEM_OVERVIEW.md` and `docs/implement-skill.md` document the coordinator, retained-log contract, same-checkout isolation, and the distinction between single-runner agent timing and multi-runner CI timing.
- [ ] The agent-policy migration remains on the serial command until a same-revision cloud-class host records `/usr/bin/time -p lib/test/run.sh` and `/usr/bin/time -p lib/test/run-parallel.sh`, exact pass/fail/skip and itemized-skip equality, the expected shard roster, and host capacity in the issue workpad. Migration requires zero failures, zero skips, and parallel `real` time no greater than 75 percent of serial `real` time.
- [ ] *(maintainer, post-merge)* One cloud implement run records that the bare `lib/test/run-parallel.sh` command was granted, started, and produced its compact aggregate. Its invocation contains no caller-side assignment, redirect, pipeline, and interpreter prefix.

## Implementation Notes

- **Approach:** add `lib/test/run-parallel.sh` as the single in-run coordinator. Read shard names from `run-shard.sh --list-shards`. Select a checked writable run root. Resolve one bounded process budget with `python3`, reserve nested pool width from that budget, and schedule shards without exceeding it. Launch one `run-shard.sh` child for a returned shard name with private temporary and tally paths. Capture child output inside the coordinator. Use monitored background subshells as shard process-group leaders with pending-signal replay through PID registration. Wait PID-by-PID for portability. Feed the current invocation's tally paths to `shard-tally.py combine` with detail cap 20. Print aggregate output and retained paths. Preserve `scripts/verification-flight.py` as the non-executing single-flight owner.
- **Code patterns:** reuse `run-shard.sh`'s shard map, `shard-tally.py`'s fail-closed records, `module-harness.sh`'s process-group cleanup discipline, and `run-module.sh`'s explicit `${TMPDIR:-/tmp}` allocation pattern. Use preflight-guaranteed `python3` when a decisive value needs computation. Use Bash builtins for command selection.
- **Focused coverage:** add and register `parallel-suite-runner` as the focused module for synthetic coordinator fixtures. Change its module file, registry entry, matching `lib/test/run.sh` driver operand and floor, and exactly-one existing shard-map membership together. No registered module invokes the coordinator against the repository's real shard population.
- **Testing strategy:** this is an automated command-boundary integration test with process-supervision fixtures. The assertion set below is exhaustive for the new coordinator contract.
  - `parallel-suite-runner.population` plants a synthetic dispatcher and proves the returned population launches with temporal overlap at budgets two and eight, stays complete at budget one, propagates the nested-pool reservation, and never exceeds the total budget. It fails RED before the coordinator exists.
  - `parallel-suite-runner.isolation` runs consecutive coordinators with planted stale tallies and proves only current explicit paths reach aggregation.
  - `parallel-suite-runner.failure-contract` drives empty, malformed, launch-failed, crashed, missing-tally, malformed-tally, nonzero, and skip-inconsistent fixtures. The unit under test and `shard-tally.py` are not mocked.
  - `parallel-suite-runner.signal-cleanup` delivers HUP, INT, and TERM before PID registration and after registration to resistant child fixtures, then proves the pending signal is replayed and those child PIDs are reaped.
  - `parallel-suite-runner.output-contract` plants assertion-like log lines plus 25 failure and skip entries, then proves clean output suppression, the cap, the omitted count, and full retained logs.
  - `parallel-suite-runner.matcher-shape` invokes the bare helper token for cloud and the `DEVFLOW_BASH` boundary for local execution, covering WSL, Git Bash, MSYS2, unset, empty, and invalid overrides. It proves the cloud actor supplies no cooperative assignment, redirect, pipeline, interpreter prefix, and background syntax.
  - `parallel-suite-runner.run-root` covers a read-only checkout with writable `TMPDIR`, an unwritable root, an exhausted root, and log and tally write failures.
  - The real-suite comparison runs outside the registered module population. It compares exact passed, failed, skipped, itemized skip population, and expected shard roster values between serial and parallel runs, emits both value sets on a named success-path record, and has a RED fixture that drops a passing assertion while both commands exit zero.
- **Documentation needed:** `docs/DEVFLOW_SYSTEM_OVERVIEW.md`
- **Documentation needed:** `docs/implement-skill.md`
- **Potential gotchas:** CI isolates shards on separate runners. The in-run coordinator shares CPU, memory, checkout state, and process namespace. The tests must prove shared-checkout safety rather than infer it from CI.
- **Potential gotchas:** the Python pool already creates concurrent workers. Running the shard population together increases process pressure on one host. The same-host timing record must include failures, skips, and host details so a fast but incomplete run is not accepted.
- **Potential gotchas:** a real-suite comparison inside `parallel-suite-runner` recursively enters the module shard. Keep that comparison outside the registered population and make nested real-population invocation fail with a named reentrancy diagnostic.
- **Potential gotchas:** do not scan a reusable tally parent with a literal expected count. Derive the population and pass current explicit paths so stale data stays outside the aggregate.
- **Potential gotchas:** the new grant is trigger-time state. That grant is unavailable as in-PR evidence during its own implementation run.

The01Geek:
## Summary
- Agent runs verified with a serial `lib/test/run.sh` even though CI already partitions the same suite into concurrent shards. `lib/test/run-parallel.sh` runs that same tested partition inside one checkout and prints one compact aggregate instead of the whole assertion stream.
- The coordinator owns every environment assignment, redirect, background process and aggregation internally, so an agent invokes one allowlisted leading token — the shapes the cloud matcher refuses caller-side never appear.
- `lib/test/run.sh` is unchanged and stays the serial primitive; focused modules stay the iteration default.

## Changes

**Coordinator**: `lib/test/run-parallel.sh` derives its launch population from `run-shard.sh --list-shards`, allocates a fresh run root under `.prflow/tmp/parallel-suite/` (falling back to `${TMPDIR:-/tmp}`, refusing by name when neither is usable), resolves one `python3` process budget capped at eight with a `DEVFLOW_SUITE_PROCESS_BUDGET` override, reserves the nested Python pool's width out of that same budget as `DEVFLOW_POOL_WIDTH`, launches each shard as a process-group leader behind a pending-signal replay guard, waits for the complete launched population, and aggregates through explicit per-run tally paths. A non-zero shard exit fails the aggregate even when that shard's tally reads clean.

**Aggregation**: `lib/test/shard-tally.py combine` gains `--detail-cap`, bounding rendered entries per detail class and announcing the omitted count. It defaults to uncapped, so CI's aggregator output is byte-unchanged, and it never bounds the counts, the pass/fail decision, or the issue-#456 skip-disagreement check.

**Coverage**: `lib/test/modules/parallel-suite-runner.sh` (97 assertions) drives the coordinator against synthetic dispatchers with `shard-tally.py` unmocked — population and overlap, budget and pool reservation, run-root fallback and refusals, stale-tally isolation, signal handling, the failure contract, reentrancy, the detail cap, and the command shapes. Registered in the flight-recorder registry, `run.sh`, the `modules-rest` shard group, the audited pin population, and CI's ShellCheck list.

**Agent policy**: the final full-suite contract in the three `.prflow/prompt-extensions/*.md` files and its `CLAUDE.md` mirror now names the coordinator — a bare leading token on cloud, the `DEVFLOW_BASH` boundary locally — with the grant-timing caveat that the new grant resolves from the default branch and is inert on this PR.

**Docs**: `docs/DEVFLOW_SYSTEM_OVERVIEW.md` and `docs/implement-skill.md` document the coordinator, the retained-log contract, same-checkout isolation, and why single-runner agent timing is not multi-runner CI timing.

## Resolves
Resolves #1086

## Test Plan
- [ ] `lib/test/run-module.sh parallel-suite-runner` reports 97 passed, 0 failed
- [ ] `lib/test/run-parallel.sh` over the real population completes and prints one compact aggregate, the shard roster, and a retained-log root
- [ ] `python3 lib/test/shard-tally.py combine <dir> --expect 1` (no `--detail-cap`) still renders every entry, so CI's aggregator is unchanged
- [ ] CI's `lib + python tests` aggregate passes with its shard matrix and required-check name unchanged
- [ ] `shellcheck --severity=warning -e SC1091 lib/test/run-parallel.sh lib/test/modules/parallel-suite-runner.sh` and `ruff check lib/test/shard-tally.py` are clean

## Post-Merge Verification
- [ ] *(maintainer)* The agent-policy migration record: a same-revision cloud-class host records `/usr/bin/time -p lib/test/run.sh` and `/usr/bin/time -p lib/test/run-parallel.sh`, exact pass/fail/skip and itemized-skip equality, the expected shard roster, and host capacity in the issue workpad. Migration requires zero failures, zero skips, and parallel `real` time no greater than 75 percent of serial `real` time.
- [ ] *(maintainer)* One cloud implement run records that the bare `lib/test/run-parallel.sh` command was granted, started, and produced its compact aggregate. Its invocation contains no caller-side assignment, redirect, pipeline, and interpreter prefix.

## Visual Changes
N/A

## Breaking Changes
None. `lib/test/run.sh`, the CI shard matrix, the `lib + python tests` required-check name, the config schema, and the shipped consumer workflow set are all unchanged.


The01Geek:
### Round-2 review findings: disposition (recorded post-merge)

This PR merged at `6d251567` (merge commit `d5bcf6a4`) while the round-2 findings were still being worked. Both were Suggestion-severity and the verdict was APPROVE, so nothing here blocked the merge — but the fixes exist and are **not** in `main`. Recording the disposition so a later pass does not re-derive it.

**Where the work is:** branch `worktree-issue-1086`, commit `955d0fac` (rebuilt on current `main`). It touches `lib/test/run-parallel.sh`, `lib/test/modules/parallel-suite-runner.sh`, `lib/test/run.sh` and `scripts/workflow-flight-recorder-registry.json`. Focused module: 104 passed, 0 failed. Not opened as a PR — that is a call for the maintainer.

#### Finding 1 — empty bash-array expansion under `set -u` — **valid, fixed**

Verified. `TALLY_ARGS` is empty on a reachable path (no shard produces a tally: every shard fails to launch, or every one dies before writing one), and the aggregation step expanded it bare.

Two corrections to the finding's own framing, neither of which changes the disposition:

- The reachable window for **this file** is bash **4.0–4.3**, not "bash < 4.4" generally. `lib/test/module-harness.sh` declares associative arrays at top level, which needs bash 4.0 of every shard — so on the 3.2 that stock macOS ships, no shard runs at all and this expansion is never the first failure. `declare -A` establishes 4.0, not 4.4; the guard's window is the gap between them.
- The severity grade is right as annotated. The fail direction is closed either way — the abort was already non-zero — so what the guard changes is *which diagnostic prints*, not whether the run passes.

Fixed with the `${arr[@]+"${arr[@]}"}` guarded form. That idiom is already this repository's convention for the case, with its rationale written out at `scripts/build-denial-record.sh` and in use at `lib/implement-stop-guard.sh` and `lib/telemetry-branch.sh` — so this is matching an existing uniform pattern, not importing an outside one. Added a `notally-all` dispatcher mode plus assertions over the all-shards-no-tally shape, which was untested (the existing case has only `beta` missing, leaving the array non-empty). Stated in the test comment rather than left implicit: on a 4.4+ host the guarded and bare forms are identical, so those assertions cover the aggregation path and do not discriminate the guard — no host available here can run the pre-4.4 arm.

#### Finding 2 — coverage gaps — **all three valid, all three fixed**

- **`_shard_cost`'s `python-pool` special case never reconciled with the real dispatcher.** Confirmed: every case drives a synthetic dispatcher that hardcodes the name. Added an assertion that `run-shard.sh --list-shards` still emits `python-pool`. Paired with the existing reservation-routing assertion, that closes the coupling from both ends without a source-text pin. Mutation-checked: renaming the shard in `run-shard.sh` turns it RED.
- **At-cap boundary asserted only for the SKIP class.** Added the failure-recap class. The two are separate call sites passing their own prefix and population, so this is a per-call-site assertion rather than a re-test of `_render_detail`. Mutation-checked against a hardcoded cap at the failure call site.
- **Signal delivery while a shard is queued behind the budget.** Confirmed a coverage gap and **not** a bug: `LAUNCHING` is 0 in the slot-wait loop, so the signal takes the already-covered handler path. What was genuinely uncovered is the queue itself — that a coordinator tearing down must not go on to launch the shard still waiting for a slot. Added that case at budget 1, which makes the state deterministic rather than timed. Both new assertions were discriminated by separate mutations: `exit 1` → `exit 0` in the handler falsifies the exit-status assertion, and a coherent non-fatal handler (return plus consistent slot bookkeeping) launches the queued shard and falsifies the other.

The module's assertion floor moved 97 → 104 in both coupled sites (the registry and the `run.sh` boundary call); `test_module_runner.py` reconciles them and passes.

Issue #1105, opened 2026-08-02
create-issue Step 3.6: a scoped audit round re-checks resolved claims, and records the draft-line span that makes the trade measurable


The01Geek:
## Dependencies
Blocked by #1104 — **merged**. #1104 does not repair any historical byte history; it makes a recorded staged write *mandatory* at every fresh file-arm dispatch (the `record-dispatch` refusal `file-arm-requires-staged-write`), so condition 3 stops failing for runs recorded after it landed. Without that enforcement this change's widened claim set has no observable effect: condition 3 still fails first. The 64 states recorded before it remain `dispatch-bytes-unrecoverable` and are not recovered by this change.

## Problem Statement

Issue #793 shipped scoped ("targeted") audit rounds for `/prflow:create-issue` Step 3.6, and gated
them on five conditions. Condition 4 requires a non-empty enumerated claim set, and a claim is a
ledger entry whose status is still `unresolved`.

The shipped Step 3.6 revision discipline records a resolution for every finding the per-finding
verification confirmed fixed, and it does so before the next round's kind is selected. So a run that
fixes what it was told about and confirms the fixes empties the very set the scoped round requires,
and the next round is dispatched cold. **The better a run behaves, the more certainly it is
ineligible.** Across the runs recorded on this machine the ledger is almost entirely resolved, and
runs that recorded a complete byte history and followed the discipline exactly still dispatched every
round cold, terminating at `empty-claim-set`.

The filter also skips the thing most worth auditing. A resolution is the drafter's own attestation
that its own fix worked, produced by the same context that wrote the defect. Excluding resolved
claims means a later round re-checks only the claims the drafter already admitted it had not fixed,
and never re-checks the ones it believes it fixed — which is precisely where a self-attested fix is
wrong.

## Current Behavior

`_enumerated_claims` walks every ledger entry in the run and yields only those whose status is still
`unresolved`; a resolved, invalidated or superseded entry is skipped. `select_round_kind`'s condition
4 then selects the cold kind with the reason `empty-claim-set` when nothing remains.

Across the recorded corpus, ledger entries are 1424 `resolved`, 46 `unresolved` and 2 `invalidated`,
and 61 of the 68 runs carrying a ledger have **every** entry resolved. Replaying the selector over
each persisted state answers `empty-claim-set` for 5 runs — the runs that got far enough for
condition 4 to be the binding one — and `targeted-eligible` for none of the 83.

[...]

Issue #1104, opened 2026-08-02
create-issue Step 3.6: refuse a file-arm audit dispatch whose draft bytes are absent from the recorded byte history


The01Geek:
## Problem Statement

Issue #793 shipped scoped ("targeted") audit rounds for `/prflow:create-issue` Step 3.6, and gave
them a durable byte history to compute their delta against: `stage` lands each canonical-draft byte
state at a content-addressed path, and `record-staged-write` records the resolved path durably so the
name survives the turn that produced it. `select_round_kind` reconstructs a round's dispatch bytes
from that history, and selects the cold whole-draft kind when it cannot.

Across the recorded runs on this machine, most carry no byte history at all. The staging transport
exists, is correct, and is simply not on the path every run actually takes — so the delta has no
"before" side, condition 3 fails, and every round of those runs is dispatched cold. No scoped round
has ever been dispatched.

The failure is silent by construction and correctly so: a missing operand must degrade the selection
rather than abort a run. What is missing is anything that makes the operand present in the first
place. The canonical write is mandated by prose in the Step 3.6 reference, and prose is exactly what
a context compaction evicts.

## Current Behavior

`select_round_kind`'s condition 3 recovers the round's dispatch bytes by locating the recorded staged
artifact whose digest equals the round's recorded dispatch digest, then re-hashing that artifact and
comparing before returning the bytes. With no recorded artifact the lookup returns nothing and the
selection answers `discovery` with the reason `dispatch-bytes-unrecoverable`.

Replaying `select_round_kind` against each persisted state file, `dispatch-bytes-unrecoverable` is
the terminal reason for 64 of 83 runs, and 71 of the 83 record no staged-write history at all — so on
those 71 no round satisfied condition 3 whatever else the run did. `targeted-eligible` is
answered for none of the 83.

The tool already enforces a staged-write guarantee at one site: `record-revision` refuses a revision
recorded while the latest round dispatched on the file arm unless it carries the intended-bytes
digest. That refusal exists because a post-revision write-failure closure needs a durable comparand.
The same reasoning applies one step earlier — a file-arm *dispatch* whose bytes are absent from the
history leaves every later round without the operand its delta needs — and there is no refusal there.

## Desired Behavior

A `/prflow:create-issue` run that dispatches an audit round on the file arm has that round's draft
bytes present in its recorded byte history, because the state owner refuses a file-arm dispatch whose
draft digest it cannot find there. A run whose canonical write did not land is unaffected: it
[...]

Issue #1103, opened 2026-08-02
create-issue Step 3.6: name the carriage cause on record-return and persist the round-kind selecting reason


The01Geek:
## Problem Statement

`/prflow:create-issue` Step 3.6 runs a per-round lifecycle through the state owner: dispatch the
round, record the auditor's return, adjudicate the findings, revise, record the revision and the
resolutions, then select the next round's kind. Issue #793 made that last step derive a scoped
("targeted") round when five recorded conditions hold, and fall back to the expensive whole-draft
round when any fails — naming the failing condition in a reason token.

Two things go wrong at that seam, and both are quiet.

First, a round that falls off the lifecycle satisfies none of the later conditions. `record-return`
treats absent carriage evidence exactly like mismatched evidence: it classifies the return
`no-parseable-verdict`, prints its ordinary contract line, exits 0, and leaves the round open. The
orchestrator that omitted an operand sees a successful-looking call. `record-adjudication` then
refuses a REVISE verdict paired with a zero unresolved count — loudly, but by then the round has
already been closed on a path nobody meant to take. The run ends with a round that completed but was
never adjudicated and never revised, so the next selection answers `no-completed-round` or
`no-revision-after-round` and the run pays for another cold round.

Second, and worse for anyone trying to diagnose the first: **the reason is never persisted.** The
round record stores the kind and the scope. The reason token that says *why* the kind is what it is
is printed on a query's stdout and discarded. So a cold round taken because it is the run's first is,
in the run's own durable state, indistinguishable from a cold round taken because a precondition
failed. The measurement instrument that joins rounds to kinds reads that same record, and can make
the same distinction: none.

The asymmetry compounds it. `record-dispatch` cross-checks the caller's declared kind against the
tool's selection and refuses a mismatch — but only in one direction that matters. A caller declaring
`targeted` over a fallen-back selection is refused loudly; a caller declaring `discovery` is always
accepted, whatever reason produced it.

## Current Behavior

`_carriage_ok` answers false when the file-arm carriage object id is absent, and false when it
disagrees with the recorded dispatch digest. `classify_return` maps both to `no-parseable-verdict`,
which sets a retry-pending state and returns through the ordinary exit-0 path. No breadcrumb names
the carriage cause, so an absent operand and a genuinely unparseable auditor return are
indistinguishable to the caller.

`cmd_record_dispatch` composes the round record with `kind` and `scope`, and no reason field. Across
[...]

Issue #1090, opened 2026-08-01
Thin-consumer local tier has no reachable engine: /prflow:review-and-fix cannot run outside Actions without DEVFLOW_VENDOR=1


The01Geek:
## Problem Statement

On a **thin install** — the default — a consumer's local checkout contains no copy of the PRFlow
engine, and nothing on the local/interactive tier knows where else to look. `/prflow:review-and-fix`
run locally in such a repository therefore has no engine to load.

This is distinct from #1085. That issue reports three defects in *how* engine resolution is written;
this one reports that on this tier there is **nothing on disk for any of those mechanisms to find**.
Fixing #1085 in full leaves this tier exactly as it is, which is why #1085 scopes it out explicitly.

## Current Behavior

- `install.sh` copies the plugin tree into `.prflow/vendor/prflow/` **only** under `DEVFLOW_VENDOR=1`.
  Its own header records the default: the vendored tree "is never committed".
- The runtime materialization that fills that path is `.github/actions/vendor-plugin` — a **composite
  GitHub Action**. It runs inside Actions and nowhere else, so it does not help a local run.
- The engine does exist locally, in the Claude Code plugin cache. On a developer machine that resolves
  to `~/.claude/plugins/cache/devflow-marketplace/prflow/<version>/skills/review/SKILL.md`.
- Nothing in `skills/`, `lib/`, `scripts/`, or `docs/` references that cache. Verified:
  `grep -rln "plugins/cache" skills/ lib/ scripts/ docs/` returns nothing.

So the local/interactive tier of a thin consumer has an engine on the machine and no documented way to
reach it.

## Why this is not simply "add the cache to the lookup"

The cache holds **many versions side by side**. On the machine this issue was written from,
`ls -d ~/.claude/plugins/cache/devflow-marketplace/prflow/*/` returns 2.28.1, 2.28.12, 2.28.18 and
several more, each with a complete `skills/review/SKILL.md`. Any lookup that reaches the cache must
also decide **which** version is the right one, and a wrong choice is worse than a miss: the run would
silently execute a stale engine against current sources rather than failing.

The obvious candidate operand is the consumer's pinned `prflow_version`, but whether the cache
directory name is guaranteed to correspond to it — and whether it is even correct to run the pinned
version rather than the loaded one — is not established here.

## Desired Behavior

A thin consumer's local `/prflow:review-and-fix` either resolves the engine it is actually running
under, or fails with a message that names the remedy. Whichever it does, it does so deliberately rather
[...]

PR #1110, opened 2026-08-02
fix: state that agent-authored prose is not hard-wrapped

The01Geek:
## Problem

Drafted issue bodies have been arriving hard-wrapped at a fixed column. Issue #1104's prose lines sit at 96–101 characters; #1103 and #1105 are the same. GitHub reflows markdown prose to the reader's width, so a hand-inserted break renders as a ragged short line, and any later edit rewraps the whole paragraph.

## Root cause: authoring habit, not a tool

Nothing in the write or staging path reflows the draft bytes, and no rule existed anywhere in either direction.

- `git grep` for `textwrap` / `fold` / `fmt -w` / `COLUMNS` across `scripts/`, `lib/`, `skills/` and `.github/` returns no hit on any drafting or posting path — the only matches are test helpers and unrelated TSV column tuples.
- `issue-audit-state.py`'s `emit-body` reads the canonical draft with `Path(source).read_bytes()` and writes it with `sys.stdout.buffer.write(body)`. It is byte-exact by construction, which is what the recorded body digest attests.
- The wrap **width varies**: ~84–90 in issue #950, ~96–101 in #1103/#1104/#1105, ~104 in #1090. A tool wrapping at a fixed column would produce one width.
- Wrapped and unwrapped issues are **interleaved within minutes**: #1073 (18:53) is wrapped, #1076 (18:55) is not; #1049 (04:23) is wrapped, #1051 (04:24) is not. No commit lands between them.
- It is also not new. #950 (2026-07-30) is wrapped at ~85, which predates the shared writing standard entirely. What changed recently is only how often it happens.

So the cause is (a) from the brief: a drafting habit with nothing prohibiting it. The repo's own reference files — `skills/create-issue/references/issue-template.md`, the skill bodies — are themselves hard-wrapped at ~85 columns, which is the house style a drafter sees while composing.

## Fix

One rule added to `lib/writing-standard.md`, the shared prose standard that `/prflow:create-issue` already reads at its compose point ("Before composing the draft prose, read the shared writing standard … and follow it"). Placing it there rather than in the skill keeps it single-sourced and fixes the same class for the other surfaces that take the same read — PR descriptions, review verdict reports, the docs family, retrospective issues.

The rule scopes itself to prose and explicitly exempts line breaks inside fenced code blocks, so a drafter does not unwrap a shell snippet. `lib/writing-standard.md` is `non_code_exempt` in `coverage-map.json` and carries no test pin, so nothing else needed reconciling.

No test is added: this is agent-executed prompt prose whose only reader is the runtime agent, which under the recorded decision in `CLAUDE.md` (#843, generalised by #876) carries no automated regression coverage by design, and a wording-only pin is prohibited by #375/#666/#810.

## Deliberately left alone

- The ~85-column hard wrapping of `issue-template.md` and the skill bodies themselves. Reflowing them is a large, purely cosmetic diff across the create-issue bundle and would collide with concurrent work; it is a separate change if wanted.
- `skills/create-issue/SKILL.md` and the `references/` routing table are untouched, so the marker contract needs no reconciliation.

## Concurrency note

Two other agents are working right now — one on CI for PR #1101 (`prflow/issue-1050`), one implementing issue **#1104, which is itself about `/prflow:create-issue` Step 3.6 staged writes** and may touch `skills/create-issue/`. This PR deliberately touches neither `skills/create-issue/` nor `scripts/issue-audit-state.py`, so it should not conflict with either.

## Verification

- `lib/test/run-module.sh create-issue-contract` — 221 passed, 0 failed
- `git ls-files '*.sh' | grep -v '^lib/test/' | xargs -r shellcheck --severity=warning -e SC1091` — clean
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` (ShellCheck 0.11.0) — clean
- `git ls-files '*.py' | xargs -r ruff check` — All checks passed
- Full suite via `lib/test/run-parallel.sh` — see the follow-up comment

### DIFF
--- a/.changeset/no-hard-wrap-agent-authored-prose.md
+++ b/.changeset/no-hard-wrap-agent-authored-prose.md
@@ -1,0 +1,7 @@
+---
+bump: patch
+---
+
+### Fixed
+
+- The shared writing standard now states that agent-authored prose is not hard-wrapped: each paragraph and bullet is one line and the renderer wraps it. Drafted issue bodies were arriving with prose broken at a fixed column, which GitHub renders as ragged short lines and which makes every later edit rewrap the whole paragraph. Nothing in the write path reflowed the bytes — no rule existed in either direction, so the drafting agent's habit decided it.
--- a/lib/writing-standard.md
+++ b/lib/writing-standard.md
@@ -23,6 +23,8 @@ Write for a competent developer who does not know this codebase. They understand
 
 5. **Define a coined term at first use, and prefer a standard term.** When a word already exists for the thing, use it rather than inventing one. When no word exists and you must coin one, define it where it first appears. Do not invent ALL-CAPS taxonomies, and do not attach serial tags such as `R7`, `META`, or `CORRECTION` — the surrounding structure already classifies the content.
 
+6. **Do not hard-wrap.** Write each paragraph and each bullet as one line and let the renderer wrap it. GitHub and every other markdown renderer reflows prose to the reader's width, so a hand-inserted fixed-column break survives into the rendered output as a ragged short line, and it makes every later edit rewrap the whole paragraph. Line breaks inside a fenced code block are content rather than wrapping, so leave those alone.
+
 ## Machine-consumed structure comes first
 
 Some of this content is parsed by tools, not only read by people. An `## Acceptance Criteria` heading, a `- [ ]` checkbox row, an HTML marker block, and a literal cross-reference token such as `PR #<N>` are read by downstream code that matches them exactly. These structures are exempt from the rules above and survive verbatim. When a rule here would push you to reword one of them, the structure wins.


The01Geek:
## Verification

**CI on this commit — all six checks green**, including the required `lib + python tests`:

| check | result |
| --- | --- |
| `lib + python tests` (required) | pass |
| `lint (shellcheck + actionlint + ruff)` | pass, 1m36s |
| `shard (monolith)` | pass, 4m38s |
| `shard (python-pool)` | pass, 5m23s |
| `shard (modules-large)` | pass, 2m8s |
| `shard (modules-pin)` | pass, 4m33s |
| `shard (modules-rest)` | pass, 3m42s |

**Local `lib/test/run-parallel.sh` — four of five shards completed clean, one could not complete on this host.** Recording it as an auditable skip rather than reporting a clean pass:

- `monolith` — 7885 passed, 0 failed, 0 skipped
- `python-pool` — 3426 passed, 0 failed, 0 skipped
- `modules-large` — 2548 passed, 0 failed, 0 skipped
- `modules-pin` — 39 passed, 0 failed, 0 skipped
- `modules-rest` — **did not complete.** Its last module, `parallel-suite-runner`, fails `psr signal: INT after registration → exits 1, acknowledges the signal, and every launched shard is reaped` with `expected 1|0|yes, actual 1|0|no` — the exit code and the acknowledgement are correct, only the reaping is not — and the module then blocks indefinitely on the next assertion.

That failure reproduced twice, including under a launcher that restores `SIG_DFL` for INT/HUP/TERM/QUIT/PIPE and sets `start_new_session=True`, so it is not the backgrounded-`SIG_IGN` artifact. Both attempts ran while nine to ten sibling `run-parallel.sh` / `run-shard.sh` processes from concurrent agents were saturating the machine (one of them wedged for over 3.5 hours), which is what the reaping assertion's wait budget is sensitive to. The sibling `psr signal: HUP after registration …` assertion — same code path, same reaping check — passed in the clean run.

I am not claiming this as a flake dismissal. The evidence that it is not this PR is that CI ran `shard (modules-rest)` on this exact commit and passed it in 3m42s, and the diff is two files, both markdown:

```
 .changeset/no-hard-wrap-agent-authored-prose.md | 7 +++++++
 lib/writing-standard.md                         | 2 ++
```

Neither file is read by `lib/test/run-parallel.sh`, `lib/test/run-shard.sh`, or any signal-handling code. If a reviewer wants the local shard green before merge, it needs a host that is not running several suites at once.

**Lints, all clean locally:**

- `git ls-files '*.sh' | grep -v '^lib/test/' | xargs -r shellcheck --severity=warning -e SC1091` — exit 0
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` (ShellCheck 0.11.0) — exit 0
- `git ls-files '*.py' | xargs -r ruff check` — All checks passed
- `lib/test/run-module.sh create-issue-contract` — 221 passed, 0 failed

Issue #593, opened 2026-07-19
Document the grant-timing bootstrap (in-PR tool grants are post-merge-only) and make the create-issue mirror-sweep scope repo-wide


The01Geek:
## Problem Statement

Two institutional facts had to be re-discovered by fresh-context audit rounds while dogfooding `/devflow:create-issue` on issue #591, because nothing committed documents them — the knowledge lives only in workflow YAML and prior issue bodies (#591, #551):

1. **Grant-timing bootstrap.** `devflow-implement.yml` resolves `allowed_tools_extra` from the default-branch checkout at workflow start, so a PR that adds a tool grant to `.devflow/config.json` cannot use that grant during its own implementing run — it is post-merge-only. The #591 draft proposed in-run verification obligations on grants it itself shipped, twice, before audit rounds caught it.
2. **Mirror-sweep scope.** The create-issue prompt extension mandates whitespace-normalized enumeration of coupled mirror sites at three sites, but never states sweep breadth. The #591 draft's sweep was directory-scoped to `lib/test/` and missed the `docs/DEVFLOW_SYSTEM_OVERVIEW.md` mirror of the extension's cloud-posture sentence.

Every future create-issue run in this repo (and every human drafter here) pays the same audit-round tax until both facts live in the shared docs and the extension's own machinery. **Deliberate narrowing:** the consumer-generic skill/template layer (`skills/create-issue/SKILL.md`, `references/issue-template.md`) also carries the enumeration mandate with no breadth statement, and this issue deliberately leaves that vendored layer untouched — the fix ships in this repo's prompt extension plus the two consumer-facing docs sentences only, keeping the change off the engine-skill surface (no writing-skills routing, no cloud-writer-manifest membership, no fenced-command changes); hoisting the breadth rule into the vendored skill/template layer is follow-up work, not this issue.

## Current Behavior

- `.github/workflows/devflow-implement.yml` triggers `on: issue_comment`; its `config` job runs a bare `- uses: actions/checkout@v6` (no `ref:`), which on an `issue_comment` event checks out the **default-branch tip**. The job then reads `.devflow/config.json` from that tree and emits `allowed_tools_extra` via `jq -r '.devflow_implement.allowed_tools // [] | …'`, spliced into the baked `--allowed-tools` string. Consequence: a grant shipped inside a PR is inert for that PR's own implementing run.
[...]

PR #1114, opened 2026-08-02
The mandated batched artifact regeneration has no invocable form on the cloud implement tier, so implement runs ship PRs with red CI

### LINKED ISSUE #1055: The mandated batched artifact regeneration has no invocable form on the cloud implement tier, so implement runs ship PRs with red CI
> **Rewritten 2026-08-02.** The original filing said no pre-push reconciliation step existed and asked for one to be authored. **That premise was false** — the step has been mandated since 2026-07-19 — and a run acting on it would have re-authored an existing mandate while leaving the real defect untouched. The observed harm is unchanged; the diagnosis and the fix are not. **Revised again 2026-08-02** after investigating the #1078 interaction: the grant belongs in `.prflow/config.json`, not the capability manifest — see *Grant channel* below.

## Problem Statement

A cloud `/prflow:implement` run reliably gets the substantive work right and then ships the PR with **red CI from a mechanical bookkeeping step** — a coverage-map entry, an exact-pinned module floor, a diff-scoped lint. Each has a self-repairing command.

The engine already knows this. `.prflow/prompt-extensions/implement.md`'s **Batched artifact regeneration** section (issue #619, merged **2026-07-19**) mandates running `lib/test/regenerate-artifacts.py` once after applying edits and before each full-suite re-verify, with a `batched-regeneration: run|refused|skipped` workpad discharge line and a two-denial stop rule.

**The defect is that the mandate has no invocable form on the cloud implement tier.** It cannot fire, and the run records `refused`/`skipped` — or nothing — rather than failing loudly. The three red PRs below post-date the mandate by twelve days.

## Current Behavior — verified at `origin/main`

**1. The mandated helper is not invocable.** `docs/cloud-allowlist.md` records `python3 <script>` — the **interpreter-head shape** — as denied even though `python3` is a granted head (probe row R4); the remedy is invoking the helper by a granted path as the command's **leading token**, which requires the executable bit.

`Verified:` `git ls-tree origin/main` —

| helper | mode | direct-token grant in `implement` profile | invocable on the cloud tier |
| --- | --- | --- | --- |
| `lib/test/regenerate-artifacts.py` | **100644** | **absent** | **no** |
| `lib/test/cloud_writer_contract.py` | **100644** | **absent** | **no** |
| `lib/test/coverage_map_guard.py` | 100755 | present | yes |
| `lib/test/test_module_runner.py` | 100755 | absent | no |

Neither the interpreter form nor the leading-token fallback resolves for the batched entry point.

**2. The precedent for the fix already exists — in the config channel.** `Verified:` this repository's own `.prflow/config.json` already grants repo-root-relative `lib/test` helpers as direct tokens: `Bash(lib/test/run.sh:*)`, `Bash(lib/test/run-parallel.sh:*)`, `Bash(lib/test/run-module.sh:*)`. `regenerate-artifacts.py` was simply never added there.

`lib/capability-profiles.json`'s `implement` profile *also* carries seven such tokens (`coverage_map_guard.py`, `test_python_scripts.py`, `test_module_harness.py` and others). **Those are the anomaly, not the model** — that profile is compiled into a workflow shipped to every consumer. They are the subject of #1078; do not follow them as precedent.

**3. The helper's internal calls are NOT a second allowlist problem.** `Verified:` every row in `regenerate-artifacts.py`'s table invokes its child as `argv = ("python3", "<path>", …)` — but those are **subprocess spawns from inside Python**, not Bash tool calls, so the permission matcher never sees them. Granting the top-level entry point as a direct leading token is sufficient; the child `python3 <path>` shapes need no grant.

**4. Three observed reds, one session (2026-07-31).** PR #1036 (issue #1025) — the `#434` diff-scoped stale-prose lint reported a gating `STALE` row. PR #1042 (issue #1032) — the `#591` coverage-map guard, both `#619` ratchet arms, **and** `review-trigger-helpers` grown with neither coupled floor site moved. PR #1045 (issue #1040) — `issue-audit-state` grown with neither coupled floor site moved. *Scoped honestly:* a fourth PR that session (#1037) also shipped red, but on a content rule (`#142`) that no self-repairing command covers; it is not evidence for this issue.

## What the batched pass does NOT cover

Two surfaces are genuinely uncovered and are the only *new* scope here:

- **`scripts/stale-prose-lint.py` is not a `regenerate-artifacts.py` row.** `Verified:` the row set is exactly `cloud-writer-manifest`, `capability-profile-literals`, `plugin-identity-regions`, `coverage-map-ratchet`, `env-freeze-advisory-region`. The `#434` lint appears in none of them. It is also not self-repairing in the same sense — it reports, it does not rewrite.
- **The exact-pinned module floors.** `Verified:` **eleven** modules carry exact-tally equality, not eight — seven asserted individually (`review-and-fix-contract`, `issue-audit-state`, `installer-wiring`, `capability-profiles`, `create-issue-contract`, `efficiency-trace-telemetry`, `harness-python-guards`) plus a four-member loop in `lib/test/test_module_runner.py` (`prompt-extension-reader`, `review-trigger-helpers`, `review-stall-backstop`, `experiment-records`). A separate loop asserts registry-to-`run.sh` **literal** equality for *every* module on disk; only the exact run tally is restricted to the eleven.

**Do not cite a floor literal in this issue or in the fix.** `review-trigger-helpers` has moved 482 → 509 → 511 → 552 → 605 in about a week. Read floors live; a transcribed number is stale before the run starts.

## Non-goals

- **Do not author a new pre-push reconciliation step.** One exists. If its wording needs a change, that is a small delta to the existing section — not a new mandate.
- **Do not gate the push on a local suite run.** `.prflow/prompt-extensions/implement.md` states the push is *not* gated on the local run finishing; the *completion claim* is. Any new obligation attaches to the claim.
- Not a change to what CI checks, nor to the `#456` skip-accounting rule.

## Technical Context

- **Files this change edits:** `.prflow/config.json` (the grant), `lib/test/regenerate-artifacts.py` (mode bit, and possibly a new row), `.prflow/prompt-extensions/implement.md` (only if the existing *Batched artifact regeneration* wording needs a delta), plus tests and a changeset.
- **Files this change must NOT edit** — context only: `lib/capability-profiles.json`, `lib/generate-capability-profiles.py`, `.github/workflows/devflow-implement.yml`, `.github/workflows/matcher-probe.yml`, `lib/review-profile.tokens`. Touching any of these means the wrong channel was used.
- **Also relevant, read-only:** `lib/test/cloud_writer_contract.py`, `docs/cloud-allowlist.md` (the shape rules and probe rows).
- **Grant channel — settled 2026-08-02, and it is NOT the capability manifest.** The grant goes in **`.prflow/config.json`'s `prflow_implement.allowed_tools`**. No `manifest_version` bump, no regenerated workflow literal, and `lib/review-profile.tokens` is untouched by construction.

  `Verified:` `.prflow/config.example.json` ships `allowed_tools: []` for all three families, so **no consumer inherits a config grant** — the channel is self-repo-only by construction. `Verified:` this repository's own `.prflow/config.json` already carries `Bash(lib/test/run.sh:*)`, `Bash(lib/test/run-parallel.sh:*)` and `Bash(lib/test/run-module.sh:*)` there. That is the established pattern for exactly this class of grant.

- **Why the manifest is the wrong channel — and why this issue must not widen the existing problem.** `install.sh` ships `.github/workflows/devflow-implement.yml` to every consumer (`for w in devflow devflow-implement`), and the generated `--allowed-tools` baseline is baked into it, so a token added to the `implement` profile reaches every consumer repository. `.github/actions/vendor-plugin/vendor-slice.sh` prunes `lib/test` from the vendored copy and no install channel puts this repository's `lib/test` into a consumer tree, so such a grant can never match a PRFlow-owned file there. `lib/capability-profiles.json` has **no self-repo-versus-consumer scoping mechanism** — one generated literal serves both tiers — which is *why* the existing seven tokens ship, not a justification for them.

  The grant is repo-relative and evaluated against the **consumer's** checkout, so a consumer whose own repository happens to carry a file at one of those paths gets it pre-authorized for execution with no permission prompt, against filenames PRFlow chose. The benefit is zero, so any residual risk is unjustified. This is the substance of **#1078**, whose direction follows from the same finding: remove the seven from the shipped literal and move any this repository needs into the config channel. **This issue adds no shipped token, so it is consistent with #1078 whichever way that issue is resolved** — the former reconciliation blocker is discharged.

- **`#593` grant-timing bootstrap — expected, not a defect.** A config-channel grant is resolved at trigger time from the default branch, so it is **in-PR-inert**: it takes effect only after merge. Do not place an in-run verification obligation on it; record it as post-merge.
- **Post-#1091 layout:** `lib/test/run.sh` is now sharded (`run-parallel.sh`, `run-shard.sh`, `shard-tally.py`). New assertions must target that layout and register in `lib/test/modules/coverage-map.json`.

## Acceptance Criteria

- [ ] `lib/test/regenerate-artifacts.py` is invocable on the cloud implement tier as a granted direct leading token: its executable bit is set, and its grant is added to **`.prflow/config.json`'s `prflow_implement.allowed_tools`**, alongside the existing `run.sh` / `run-parallel.sh` / `run-module.sh` entries.
- [ ] **No token is added to `lib/capability-profiles.json`**, no workflow literal is regenerated, `manifest_version` is unchanged, and `lib/review-profile.tokens` is byte-identical. A PR that touches any of those has used the wrong channel.
- [ ] The grant's post-merge-only effect (`#593`) is recorded rather than verified in-run **(post-merge)**
- [ ] The same determination is made and recorded for `lib/test/cloud_writer_contract.py` — either granted the same way, or explicitly recorded as reached only as a `regenerate-artifacts.py` subprocess and therefore needing no grant.
- [ ] A test asserts the batched entry point is reachable in the granted shape — that its head resolves under the generated implement baseline, driven by the existing `lib/test/extract-command-heads.py` / `extract-command-shapes.py` machinery rather than a new bespoke check.
- [ ] The `#434` stale-prose lint's non-coverage by the batched pass is resolved: either it becomes a `regenerate-artifacts.py` row, or the prompt-extension section names it as a separate obligation. **State which, and why.**
- [ ] The exact-pinned module floors are addressed for **eleven** modules, not eight. The chosen behavior is stated explicitly in the PR: **recommended — report a mismatch and never write a decrease; write a raise only from a measured tally read from an actual module run.**
- [ ] No floor literal is transcribed into prose, a test name, or a comment; floors are read live.
- [ ] Any new source-presence pin declares a `# structural-pin-ok: <category>` with a non-empty rationale. No wording-only pin is added (`CLAUDE.md` #375/#666/#810).
- [ ] A uniquely-named `.changeset/*.md` with a `bump:` frontmatter key is added and cited in the PR (engine surface).
- [ ] `lib/test/run.sh` reports zero failed checks over the committed tree, and its skip population is recorded per the `#456` rule rather than laundered into a clean pass.

## Unestablished — stated as unestablished

- **The causal link is inferred, not measured.** That the three red PRs are explained by the batched pass being unrunnable is the best available explanation, not an observed fact: no execution transcript from those runs was read for a denial of the batched entry point. A run implementing this should look for that evidence and record what it finds — including if it contradicts this diagnosis.
- Whether the runs recorded `batched-regeneration: refused`, `skipped`, or omitted the line entirely is not established here.

The01Geek:
## Summary
- The batched artifact-regeneration pass the implement loop mandates had no runnable form on the cloud tier: the helper carried no executable bit, no allowlist grant, and was named in prompt prose only as `python3 <path>` — the interpreter head that matcher silently denies. Runs therefore skipped the pass and shipped PRs with red CI from unregenerated artifacts.
- The helper is now a granted direct leading token, and every prompt surface fences the invocation so the command-head scanner and the matcher can both see it.
- The pass gained a second writing row that measures the eleven exact-policy module floors through the real focused runner and raises them, never lowering: a decrease is a non-writing judgment item, and any untrustworthy measurement is a non-writing infrastructure failure.

## Changes

**Cloud invocability**: `lib/test/regenerate-artifacts.py` is mode `100755` and `Bash(lib/test/regenerate-artifacts.py:*)` is granted in both `prflow_implement.allowed_tools` and `prflow.allowed_tools` — the latter because two of the three prompt mirrors are loaded by the command path. No token was added to `lib/capability-profiles.json`, no workflow literal was regenerated, `manifest_version` is unchanged, and `lib/review-profile.tokens` is byte-identical.

**Prompt surfaces**: the batched invocation and the `--list` conflict oracle are fenced as direct leading tokens in `implement.md`, `review-and-fix.md` and `receiving-code-review.md`. `lib/test/cloud_writer_contract.py` is deliberately *not* granted — it is reached only as a `regenerate-artifacts.py` subprocess, and a matcher regression asserts its head stays ungranted in the resolved profile.

**Measured floor reconciliation**: a new `lib/test/reconcile-module-floors.py` selects its population from `assertion_floor_policy: "exact"` registry metadata rather than a second hand-maintained list, lowers floors in a temporary registry to measure, and stages both coupled sites through one `git apply` so a raise and its `run.sh` call site move together or not at all. The registry is rewritten by surgical token replacement, not re-serialization, so a one-digit raise stays a one-digit diff.

**Conflict routing**: the row's conflict class is `by-hand`, not `reconcile-source`. Both declared outputs are hand-authored files in which the row owns a single numeric token per module, so classing them as generated would tell an agent never to hand-merge conflicts in `lib/test/run.sh` — the repository's largest hand-authored file — and to regenerate a file no generator produces.

**Stale-prose lint (AC 6)**: it stays a **separate obligation**, named as such in the prompt extension, rather than becoming a row. `scripts/stale-prose-lint.py` consumes a caller-selected diff on stdin and requires a correct post-image mode; it has no checked-in artifact it could regenerate, so it does not fit the row contract.

**Tests**: coverage is executable rather than prose-pinned — the real `extract-command-heads.py` / `extract-command-shapes.py` run over every (extension, section) pair, so an inline-backtick or interpreter-head regression fails while a rewording does not. Portability fix: a row-kind mutation used GNU sed's `0,/re/s//repl/`, a silent no-op under BSD sed that was reading an unmutated helper.

## Resolves
Resolves #1055

## Test Plan
- [x] `lib/test/run-module.sh regenerate-artifacts` — 276 passed, 0 failed
- [x] `lib/test/test_reconcile_module_floors.py` — 13 passed, including a mutation-verified guard (neutering the span scanner's string-awareness turns it RED and leaves the other twelve green)
- [x] `lib/test/run-shard.sh monolith` in-env — 7951 passed, 0 failed, **0 skipped** (issue #456 accounting: the skip population is empty, not omitted)
- [x] `lib/test/regenerate-artifacts.py` — exit 0, all six rows clean, including a live measurement of all eleven exact modules
- [x] `lib/test/pin-corpus-lint.py mutation-routing-worktree .` — rc 0
- [x] `ruff` over all tracked Python; `shellcheck` over non-`lib/test/` shell and over `run.sh` with `--extended-analysis=false`
- [x] CI green on this exact head sha across every shard plus the required `lib + python tests` check

## Post-Merge Verification
- [ ] The grant's post-merge-only effect (`#593`): `.prflow/config.json` is resolved at trigger time from the default branch, so the new grant is inert for this PR's own runs and takes effect only once merged. Confirm a post-merge implement run invokes `lib/test/regenerate-artifacts.py` without a denial.

## Visual Changes
N/A

## Breaking Changes
None. The grant is additive and the reconciliation is raise-only, so no existing floor can be lowered by this change.

## Notes
- Versioning: `.changeset/issue-1055-cloud-batched-regeneration.md` (`bump: patch`, `type: Fixed`) — the engine-surface changeset this PR is required to carry.
- Writing-skills evidence: every prompt-surface edit was routed through a context-isolated subagent invoking `superpowers:writing-skills` (mode `subagent`; the availability check reported the skill present on each pass).
- Follow-ups deferred under operator direction to fix only Critical/Important this round are filed as #1126, including two Important items deferred with a stated reason rather than as minor: bounding the floor measurement's heavy units (unsafe until tally mode-invariance is established) and settling two exact-policy modules whose assertion counts are host-gated.
- Review coverage is stated honestly: the inline pass ran engine Phases 0/1/3 with five reviewers, but no Step 2.6 shadow fan-out, so shadow agreement is **not verified** and this is not reported as a clean approve.

Generated via `/prflow:implement 1055`


The01Geek:
# PRFlow Review — PR #1114



**Status:** 🚀 Reviewing
**Diff profile:** `engine_self_modifying + detect_all_audit` — full checklist; completeness critic required
**Run:** _(local run)_
**Last updated:** 2026-08-02T04:28:49Z

## Blueprint
- [ ] Review agents (Phase 3)

## Findings (live)
Phase 0.6 stale-prose lint: clean (exit 0; no STALE or UNRESOLVABLE rows).

## Verdict
_(pending)_

Issue #1179, opened 2026-08-03
Raise the implement tier's per-command Bash ceiling in devflow-implement.yml, and scope the "not escapable" prose to in-run


The01Geek:
## Problem Statement

On the cloud implement tier, every whole-suite verification command is killed at 10 minutes before it prints anything, and the run throws the work away.

`.github/workflows/devflow-implement.yml` passes **no `settings:` input** to its `anthropics/claude-code-action@v1` step and sets no timeout override anywhere, so Claude Code's default `BASH_MAX_TIMEOUT_MS` (600000 ms) and `BASH_DEFAULT_TIMEOUT_MS` (120000 ms) are both live. `lib/test/run-parallel.sh` — the parallel coordinator CLAUDE.md names as the run's final gate (#1086) — structurally cannot finish inside that window on the runner the workflow actually uses. The invocation is terminated with no summary, no tally, no partial output: the shards' work is done and then discarded wholesale.

The run then does the right thing (decomposes the partition through `lib/test/run-shard.sh` and recombines it, per #1132) and gets its evidence anyway. That makes the coordinator invocation **pure loss on every run** — not a failed attempt at something useful, but ~10 minutes of runner time and model wall-clock spent to produce nothing, immediately before the same work is redone shard by shard.

A second, quieter cost: the prose that describes this ceiling reads as though it were immovable, so nobody has looked at the workflow. `.prflow/prompt-extensions/implement.md` states the ceiling "is not escapable in-run (… no `BASH_MAX_TIMEOUT_MS` override exists …)". That sentence is **accurate about the run** — an agent mid-run genuinely cannot raise it — but it has been read as "immovable", and CLAUDE.md's tier-2/3 bullets document shard decomposition as the sanctioned route on that premise. The ceiling is entirely movable at *authoring* time, by the workflow that launches the run.

## Evidence

All figures measured **2026-08-02** from `claude-execution-scrubbed.json` transcript artifacts plus `gh api .../jobs` timings, unless marked otherwise.

**The discarded commands — 5 of 5 implement runs analysed:**

| run | command | duration | outcome |
| --- | --- | --- | --- |
| 30747308684 | `run-parallel.sh` | 600.0 s | exit 143, no output |
| 30750396866 | `run-parallel.sh` | 580.0 s | exit 143, no output (plus `run.sh` 590 s, also killed) |
[...]

PR #1306, opened 2026-08-05
test: extend the #908 settings-input hook guard to devflow.yml

The01Geek:
## Problem

`lib/test/run.sh`'s `#908` workflow-settings block defines a hook extractor (`_wfg_hook`, backed by the `hook-check.py` file it writes out) that refuses a `hooks` key — or any `PreToolUse` mention reachable through a `with:` value — in a `claude-code-action` step's `settings:` input. It enforces the `#908` tier-scope decision: the PreToolUse shape guard is registered on the **review** tier (`devflow-runner.yml`) only.

**It was invoked against `.github/workflows/devflow-implement.yml` alone.** `.github/workflows/devflow.yml` — the shipped `command` tier (manual `/prflow:review`, `/prflow:review-and-fix`) — was referenced in the block only as `_908_COMMAND_YML`, and only by the `#1179` *ceiling* extractor. Nothing ran the hook extractor over it, so adding a `hooks` key to its action step passed the suite silently.

That matters because `devflow.yml` is a live, shipped tier: `install.sh` copies it into every consumer repo, and its own `settings:` comment records that `/prflow:review-and-fix` Step 0.5's branch sync moves the working tree onto the **PR head** before the agent runs. A hook registered there would be PR-author-editable code executing ahead of every tool call. Whether such a hook should ever exist on that tier is a separate open decision — this change makes either answer a *visible* one instead of a silent default.

### Gap verified before changing anything

- `_wfg_hook` is defined once (`lib/test/run.sh`) and invoked in this block only. Pre-change, its only live-workflow call site was `"$(_wfg_hook "$_908_IMPLEMENT_YML")"`; the other calls are synthetic fixtures.
- `_wfg_grep_fallback` (the degraded-host arm, matching `PreToolUse|pretooluse-shape-guard` on non-comment lines) was likewise driven over `$_908_IMPLEMENT_YML` and the fixtures, never over `$_908_COMMAND_YML`.
- Neither `_wfg_hook`/`_wfg_grep_fallback` nor `_908_COMMAND_YML`/`_908_IMPLEMENT_YML` appear anywhere outside `lib/test/run.sh` (repo-wide search) — the guard's target list is stated in exactly one place, so there is no mirrored constant or coverage-map entry to keep in lockstep.
- Correction to the audit note that prompted this: `devflow.yml` **does** carry a `settings:` key today. It is the `#1179` env-only block scalar (`BASH_MAX_TIMEOUT_MS: "1200000"`) at the `command` job's `claude-code-action` step — exactly the shape the hook guard is designed to *allow*. So the clean file passes the newly added row rather than needing a workflow edit; no `.github/workflows/**` file is touched by this PR.

**Executable demonstration of the gap.** Driving the *pre-change* block (`git show HEAD~:lib/test/run.sh`, lines of the `#908` block) against a shadow tree whose `devflow.yml` carries `{"hooks": {"PreToolUse": [{"matcher": "Bash", "hooks": []}]}, "env": {…}}`:

```
32 passed, 0 failed
```

Green on a hook-bearing `devflow.yml`. That is the gap.

## What changed

`lib/test/run.sh` only — three additions inside the existing `#908` block:

1. **Armed arm:** `devflow.yml` is now driven through `_wfg_hook`, alongside the existing `devflow-implement.yml` row, with the same allowance (an env-only `settings` block passes).
2. **Both fallback arms:** `_wfg_grep_fallback` is driven over `devflow.yml` in the armed matrix *and* in the `else` (no python3/PyYAML, or no scratch space) branch, matching how `devflow-implement.yml` is treated on both hosts.
3. **Negative control.** A new fixture composer (`compose-hook-negative.py`, written out as a file like the block's other extractors) splices a `hooks` key into a copy of the **live** `devflow.yml`'s existing `settings: |` block scalar — every other byte, authoring comments included, intact — and both the armed extractor and the degraded-host grep must refuse it. The composer fails **loudly** (its own assertion row) on zero matches, more than one match, or a splice that changed nothing, so it can never hand the refusal rows an unmutated copy that would let them pass vacuously.

### Fixture-matrix semantics for the second target

The matrix already carries `hooks-key-quiet` (`{"hooks": {"Stop": []}}`), asserting that **any** `hooks` key is refused, not merely `PreToolUse`. That semantics is correct for `devflow.yml` for the same reason it is for `devflow-implement.yml`: the property under guard is "this tier registers no hook **through the action's settings input**", and the `#908` tier-scope decision sanctions such a registration on the review tier only. It is deliberately *not* a claim about the base-restored project `.claude/settings.json` Stop hooks, which arrive by a different mechanism and are untouched by this input. The full matrix (10 refusal fixtures + 4 allow fixtures) is unchanged and still passes with the new target present.

The fallback's `PreToolUse|pretooluse-shape-guard` pattern is still correct for the new file: `devflow.yml` names neither token anywhere today, in code or in comment. That means the clean row alone could not distinguish a working pattern from one that matches nothing — which is precisely why the negative-control grep row was added.

## Verification evidence

Focused verification drove the shipped `#908` block text itself (extracted verbatim from `lib/test/run.sh` into a scratch driver supplying `assert_eq`/`skip`), against four host/input combinations. No suite behavior was paraphrased.

| # | block | tree | result |
|---|---|---|---|
| 1 | post-change | live (clean) | `37 passed, 0 failed` |
| 2 | **pre-change** | shadow, hook-bearing `devflow.yml` | `32 passed, 0 failed` — **the gap** |
| 3 | post-change | shadow, hook-bearing `devflow.yml` | `35 passed, 2 failed` — **the guard fires** |
| 4 | post-change, degraded host (`python3 -c 'import yaml'` stubbed to fail) | clean, then hook-bearing | `6 passed, 0 failed` + 3 host-capability skips; then `5 passed, 1 failed` |

Row 3's two failures are exactly the two new live-workflow rows, in both arms:

```
FAIL   #908 AC1: devflow.yml (command tier) registers no PreToolUse guard (env-only settings allowed)
       want=[yes]  got=[no]
FAIL   #908 matrix: the degraded-host grep fallback clears the command tier's live workflow
       want=[yes]  got=[no]
```

Row 4 confirms the degraded arm is not a silent pass: on a host without PyYAML the new `devflow.yml` row still runs through the grep fallback and still flips to `no` on a hook-bearing file, while the python-dependent rows self-skip as `host-capability` per `#456` rather than vanishing.

Composer fail-loud check: pointed at a workflow with no block-scalar `settings:` input, it exits nonzero (→ `composer-failed`, which satisfies neither assertion) instead of writing an unmutated copy.

Lints, all green on the rebased tree:

```
shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh   → rc 0
python3 lib/test/pin-corpus-lint.py mutation-routing-worktree .  → MUTATION-ROUTING-STATIC-SCAN-COMPLETED, rc 0
python3 lib/test/coverage_map_guard.py .                         → rc 0
python3 lib/test/lint-tree-enumeration.py --root .               → audited 128 of 128 files, rc 0
python3 lib/test/lint-carveout-guard.py --repo-root .            → OK, rc 0
```

The aggregate signal is CI's `lib + python tests`; a green local full-suite aggregate was deliberately not chased.

## Pins and coupled sites reconciled

- **No count or ordinal pin moved.** The block carries no comment citing "the Nth call", an arm count, or a fixture tally; searched the block and the file for such wording before editing.
- **No new pin authored.** The added assertions compare *process output* (`$(_wfg_hook …)`, `$(_wfg_grep_fallback …)`, `$(_wfg_compose_neg …)`), not file text, so they are behavioral rows and need no `# structural-pin-ok:` declaration. Confirmed by the `#810` gate above, which runs post-commit against the merge base.
- **Coverage map:** the new assertion names carry the existing `#908` label, which already has a `run_sh_blocks` entry (`owner: unmodularized`). `coverage_map_guard.py` is green, so no entry was added or changed.
- **Prose kept in step:** two same-file comments that said the guarded property holds "on the implement tier" / that `devflow.yml` "is driven through the same ceiling extractor" were updated in the same commit to name both tiers and both extractors.

## Changeset call: none

`CLAUDE.md` and `.prflow/prompt-extensions/implement.md` scope changesets to changes that "reach consumer repos as an update — a fix, feature, or breaking change to the engine surface", and state that "internal-only changes (tests, CI, dev-only docs) add **no** changeset". This PR touches exactly one file, `lib/test/run.sh`. It is test surface: the vendor slice prunes `lib/test/` entirely, `install.sh` ships none of it, and nothing here changes any byte a consumer receives or any runtime behavior — the shipped `devflow.yml` is not modified. So no changeset. (Deliberately not hedged into adding one: a `.changeset/*.md` that consolidates into a user-facing CHANGELOG entry for a change with zero consumer-visible effect is a false release note.)

## Out of scope

- No `hooks` key was added to any workflow; no file under `.github/workflows/` is touched.
- `scripts/pretooluse-shape-guard.py` and the denial-logging helpers are untouched.
- `CLAUDE.md` is untouched.
- Whether the `command` tier *should* register a hook remains an open decision this PR does not take a position on — it only ensures the suite stops being silent about it.

No closing keyword; no issue exists for this yet.

### DIFF
--- a/lib/test/run.sh
+++ b/lib/test/run.sh
@@ -46365,7 +46365,10 @@ _908_IMPLEMENT_YML="$LIB/../.github/workflows/devflow-implement.yml"
 # The `command` tier (manual /prflow:review, /prflow:review-and-fix). It carries the same
 # #1179 ceiling for the same reason the implement tier does — the whole-suite coordinator
 # its prompt extension names as the final gate exceeds Claude Code's 600000 ms default —
-# and is driven through the same ceiling extractor below.
+# and is driven through the same ceiling extractor below. It is driven through the HOOK
+# extractor below for the same reason the implement tier is: the #908 tier-scope decision
+# registers the shape guard on the review tier ONLY, so a hooks key appearing in this
+# tier's settings input is out of scope by that decision and not merely undeclared.
 _908_COMMAND_YML="$LIB/../.github/workflows/devflow.yml"
 assert_eq "#908 AC1: devflow-runner.yml's claude-code-action step carries a settings: input" "yes" \
   "$(grep -qF 'settings: |' "$_908_RUNNER_YML" && echo yes || echo no)"  # structural-pin-ok: routing-dispatch-contract -- pins the effectiveness channel AC1 requires: the claude-code-action settings input is what registers the PreToolUse hook with the review-tier runner
@@ -46378,9 +46381,9 @@ assert_eq "#908 AC1: the settings input registers a PreToolUse hook naming the g
 # registration would evade it entirely. Parse the YAML with PyYAML instead and check,
 # structurally, that no step registers a PreToolUse hook — the same precision the #460
 # errexit fixture above uses to extract a step's `run:` body, applied here to a `with:`
-# key check instead. The property #908 protects on the implement tier is "no PreToolUse
-# guard" (that shape guard is review-tier-only), NOT "no settings input at all": issue
-# #1179 legitimately adds an env-only `settings` block here to raise BASH_MAX_TIMEOUT_MS,
+# key check instead. The property #908 protects on the implement and command tiers is "no
+# PreToolUse guard" (that shape guard is review-tier-only), NOT "no settings input at all":
+# issue #1179 legitimately adds an env-only `settings` block to each to raise BASH_MAX_TIMEOUT_MS,
 # which registers no hook. So a `settings` block is allowed IFF it declares no `hooks`
 # key and no `with` value mentions "PreToolUse". Falls back to a PreToolUse-only grep
 # when python3/PyYAML is unavailable, so the assertion still runs somewhere.
@@ -46528,11 +46531,43 @@ def bounded(val):
 
 print("ok" if vals and all(bounded(v) for v in vals) else "no")
 PY
-  # Both runners fail CLOSED: a YAML parse error or an otherwise unhandled shape exits
+  # NEGATIVE-CONTROL COMPOSER for the live-workflow hook rows. A guard shown only to PASS
+  # on a clean file is the "passes on the very inputs it was added to catch" class: such a
+  # row reads identically whether the guard works, is hardwired to `yes`, or is pointed at
+  # a workflow carrying no claude-code-action step for it to inspect. This writes a copy of
+  # a LIVE workflow with a hooks key spliced into its existing `settings: |` block scalar —
+  # every other byte, authoring comments included, intact — so the refusal is demonstrated
+  # against the real file rather than against a synthetic stand-in for it.
+  #
+  # It fails LOUDLY rather than emitting an unmutated copy: zero matches (the settings input
+  # reshaped away) or more than one (an ambiguous splice target) exits nonzero, and so does
+  # a splice that changed nothing. An unmutated copy would make the refusal rows below pass
+  # vacuously — the same fail-open the composer exists to rule out.
+  cat > "$_WFG_D/compose-hook-negative.py" <<'PY'
+import re, sys
+
+src = open(sys.argv[1], encoding="utf-8").read()
+matches = list(re.finditer(r"^(?P<ind>[ \t]*)settings: \|\n(?P<body>[ \t]*)\{\n", src, re.M))
+if len(matches) != 1:
+    sys.exit(1)
+m = matches[0]
+out = (
+    src[: m.end()]
+    + m.group("body")
+    + '  "hooks": {"PreToolUse": [{"matcher": "Bash", "hooks": []}]},\n'
+    + src[m.end() :]
+)
+if out == src:
+    sys.exit(1)
+open(sys.argv[2], "w", encoding="utf-8").write(out)
+print("composed")
+PY
+  # All three runners fail CLOSED: a YAML parse error or an otherwise unhandled shape exits
   # nonzero, and the caller substitutes a value that satisfies no assertion below (issue
   # #908 confirmatory review; mirrors the _908_PROBE_JOB extractor's discipline).
   _wfg_hook() { local o; o="$(python3 "$_WFG_D/hook-check.py" "$1" 2>/dev/null)" || o="extractor-failed"; printf '%s\n' "$o"; }
   _wfg_ceiling() { local o; o="$(python3 "$_WFG_D/ceiling-check.py" "$1" 2>/dev/null)" || o="extractor-failed"; printf '%s\n' "$o"; }
+  _wfg_compose_neg() { local o; o="$(python3 "$_WFG_D/compose-hook-negative.py" "$1" "$2" 2>/dev/null)" || o="composer-failed"; printf '%s\n' "$o"; }
   # Compose a synthetic workflow fixture named $1 whose single claude-code-action step
   # carries the `with:` body supplied on stdin (indented to its position under `with:`).
   # The result is a real YAML document reaching the extractors through the same entry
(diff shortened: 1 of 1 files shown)


The01Geek:
Follow-up on the three review Suggestions, applied in `11203e887` (`lib/test/run.sh` only — it landed on the branch just before the merge).

**Applied all three.**

**1 — negative control now proves refusal *for the right reason*.** The finding was correct: `hook-check.py` returns `no` both for a present hooks key and for a settings block it cannot parse as a JSON object, so a composer whose hooks literal was ever reworded into invalid JSON would keep the refusal row green while exercising the fail-closed parse arm and never touching hook detection. Closed it with a **settings-shape read-back** — a separate extractor, deliberately *not* a call into `hook-check.py`, since a read-back sharing the program under test could not contradict it. It reports one token per `claude-code-action` step (`object-with-hooks` / `object-no-hooks` / `not-an-object` / `absent`), and both directions are asserted: the clean `devflow.yml` settings input is `object-no-hooks`, the spliced copy `object-with-hooks`. So the refusal is attributable to a hooks key inside still-valid JSON.

Demonstrated the assertion discriminates, rather than assuming it: I reworded the composer's hooks literal into invalid JSON (dropped one `}`) and drove both extractors over the result — `hook-check.py` still answered `no` (the vacuity the finding described, invisible to the refusal row), while the shape read-back answered `not-an-object` and would take the row RED.

**2 — the unreachability claim holds; the check is gone.** Verified before acting: the insertion is a fixed non-empty 63-byte literal plus optional indentation, so once a single match is found `len(out) == len(src) + 75` on the live file and `out` is strictly longer than `src` for any input — `if out == src` can never fire. Dropped the arm and the "and so does a splice that changed nothing" clause. The comment now names the read-back as what actually establishes the copy is mutated, and records in one parenthetical why the old arm could not, so nobody re-adds it.

**3 — applied; it really was a one-liner.** It needed no synthetic fixture at all: `devflow-runner.yml` is a live, tracked workflow that genuinely registers the guard inside its `settings:` block scalar (AC1 above already pins that registration), so driving the grep fallback over it and requiring `no` is one assertion needing neither python3/PyYAML nor scratch space. It sits with the other always-on fallback rows, outside the armed/degraded split, so it closes the degraded-arm vacuity for **both** tiers. Previously that arm asserted only `yes` verdicts and could not tell a working pattern from one that had stopped matching anything. Had it needed scratch space or a composed fixture I would have skipped it per the brief — it did not.

**Stale-count reconciliation** (same commit, PR #553 class): adding a runner and rows rotted three self-referential counts in this block — "All three runners", "these two rows", and "Both directions of that guard". All three are now count-free. Also updated the degraded arm's comment, which previously said the command tier's negative control lived in the armed arm only; it now points at the always-on refusal row as what discriminates for it.

**Verification.** `monolith` shard (the covering test — the coverage map has this block as `run_sh_blocks: 908`, owner `unmodularized`, so no focused module exists): **8500 passed, 0 failed, 0 skipped**, with every row named:

```
PASS  #908 fallback: it REFUSES the review tier's live workflow, which really does register the guard
PASS  #908 negative control: a hook-bearing copy of devflow.yml is composed from the live workflow
PASS  #908 negative control: the clean devflow.yml settings input parses as a JSON object carrying no hooks key
PASS  #908 negative control: the spliced copy still parses as a JSON object, now carrying a hooks key
PASS  #908 negative control: the hook guard REFUSES a devflow.yml carrying a settings-input hook
PASS  #908 negative control: the degraded-host grep fallback also refuses that same devflow.yml
```

`shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` clean (0.11.0). CI green on `11203e887`, `lib + python tests` included.

No changeset, consistent with the prior commit: `lib/test/` is pruned from the vendor slice, so this ships no consumer-visible bytes. No assertion floor to reconcile — the block is unmodularized in `run.sh`, and no registry carries a count for it.

Issue #1064, opened 2026-08-01
Denied permission attempts are not durably recorded with their command, and the durable channels that exist are asymmetric across the engine-invoking workflows


The01Geek:
> **Scope boundary, stated up front.** This issue is about **post-mortem observability** — recording what the harness refused, with enough fidelity to diagnose it later and to query it across runs. It is **not** a runtime control. It does not propose, depend on, or decide anything about the `PreToolUse` shape guard. Its relationship to #1047, #919 and #1051 is spelled out in a dedicated section at the bottom, and in each case the relationship is "independent".

All verification below was run at `045617f4` (`main`, after `git fetch`). Every `Verified:` line names the command actually run and reflects output actually read.

---

## Problem Statement

When a cloud run emits a Bash command in a shape the `claude-code-action` permission matcher does not grant, the command is refused silently — no error, no output, no exit code. The repository's whole diagnosis story for that class rests on the execution file's `permission_denials` data. Today that data is surfaced in a form that is **lossy at the point that matters, and durable nowhere on any tier that can actually run**.

Three concrete losses, each verified below:

1. **The command text is cut at 200 characters.** Denied commands are typically long pipelines, and the ungranted head is frequently in the tail that gets cut. The surfaced line can therefore contain none of the information needed to identify what was refused.
2. **The surfacing is ephemeral.** On both live tiers it goes to stdout and `$GITHUB_STEP_SUMMARY` only. Nothing lands on the telemetry branch, in the workpad, or in the learnings store. It cannot be queried across runs, so "which shapes does the matcher keep refusing, and how often" is not answerable from this repository's own data.
3. **The one durable *count* channel that ever existed has lost its producer.** `scripts/build-experiment-records.py` reads `permission_denials_count` out of the `Devflow Review` check-run summary — which was written by `devflow-review.yml`'s `finalize_check`, deleted under #936. Its own docstring records this. So on every live-tier run today, the experiment store's `permission_denials_count` is structurally unestablished.

A second, orthogonal problem the same change should fix: the one channel that *does* preserve full fidelity — the opt-in scrubbed transcript artifact — exists in exactly one workflow, and it is the one workflow that cannot run. That asymmetry is not a deliberate tiering decision recorded anywhere; it is an artifact of where the feature happened to be built. The tier that spends the most budget on the longest autonomous runs, `devflow-implement.yml`, has no forensic channel at all.

---

### The 200-character truncation

[...]

Issue #475, opened 2026-07-14
Harness-side cost floor: merge the claude-code-action execution_file's cost into per-run efficiency records through the --persist telemetry-branch path


The01Geek:
## Problem Statement

Every recovery layer in the efficiency-telemetry pipeline reconstructs run records from agent-authored artifacts: `--self-check` warns from the agent's own workpads, `--persist` derives records from `iter-<N>.json` emits, and the Layer-3+ synthesis floor rebuilds iteration structure from `fix: address review findings (iteration N)` commits. When the review-and-fix loop drops its emits, structure is recoverable but the COST half is not: a synthesized record carries no `tokens`/`calls`/`wall_clock_s` figures, so `_run_cost` returns `None` and `telemetry_complete` is `false`. The runs that drop telemetry are precisely the abnormal runs the retrospective and experiment analyses most need cost data for — and today they contribute none.

For every cloud run whose claude step lives long enough to emit it, claude-code-action writes an execution file — harness-side, independent of anything the agent volunteers — and that file carries cost directly (`costUSD`, `total_cost_usd`, per-model `modelUsage`, per-message `usage`). Both writable workflows already consume it for diagnostics. This issue adds the first floor in the pipeline that is NOT fed by an agent-volunteered operand: a deterministic reader plus merge path that lands the execution file's cost in the run's efficiency record through the #442 telemetry-branch persist path. Payoff of the #437/#438/#457/#458 arc.

## Current Behavior

- `.github/workflows/devflow.yml` and `.github/workflows/devflow-implement.yml` plumb `steps.claude.outputs.execution_file` into their execution-diagnostics step (`EXECUTION_FILE` env), which surfaces `is_error`, `num_turns`, `duration_ms`, `total_cost_usd`, and `permission_denials_count` into the run log and `GITHUB_OUTPUT` — then the cost figures are discarded.
- Both workflows' `Persist review-and-fix observability artifacts (backstop)` steps (`if: always()`) run `bash "$HELPER" --persist` with no execution-file input; `lib/efficiency-trace.sh` has no way to receive one (its argv parser rejects unknown flags with exit 2).
- A run that dropped every emit AND has no unrecorded fix commits produces no record at all: the synthesis floor exits with "nothing to synthesize", and the run is invisible to `_index_efficiency`.
- A synthesized record carries iteration structure but no numeric telemetry: `_run_cost(record)` is `None`.

## Desired Behavior

[...]

PR #1307, opened 2026-08-05
fix: give the denial floor a skeleton arm so denials survive a costless run

The01Geek:
## The problem

Permission denials on the cloud tiers are recorded durably by pushing a denial record to
the `prflow-telemetry` orphan branch, via `lib/efficiency-trace.sh --persist`. That works —
except the denial record is a **rider on the efficiency record**.

`apply_denial_floor` had three arms: merge onto a record staged this pass, merge onto one
already on the branch, or **decline** with a warning and write nothing. The decline arm was
reachable, and reachable on exactly the runs that need denial forensics most:

1. `scripts/prepare-harness-floor.sh`'s `_cost_has_figures` refuses to stage an all-null
   `harness_cost` when the execution file carried no cost or usage figures.
2. That empties `DEVFLOW_EXECUTION_COST`.
3. `apply_harness_floor` returns at its **first guard** on an empty cost — before its own
   skeleton arm — so no efficiency record for the run-id exists anywhere.
4. `apply_denial_floor` therefore found nothing to merge onto and declined.

Result: a run that produced denials but died before yielding usable cost figures had its
fully-built denial record discarded, leaving only a `::warning::` in a job log that expires.
That is the stall / crash / execution-ceiling class.

The denial record itself was still being built on that path: `scripts/build-denial-record.sh`
only declines when the execution file is *absent or empty*; a present-but-figureless file
yields a real record, and an unparseable one yields `count: "unavailable"`.

## The fix

`apply_denial_floor` gains a **skeleton arm mirroring `apply_harness_floor`'s** — same
run-id identity targeting, same class gate, same empty-PR gate, same skeleton-overwrite
guard, same minimal `{schema_version, slug, generated_at, source: null, synthesized: true,
iterations: 0, per_iteration: [], telemetry: []}` shape — carrying `permission_denials`
where the cost skeleton carries `harness_cost`. The `synthesized: true, source: null` marker
is the existing sanctioned precedent for this situation (verified against
`apply_harness_floor`'s own skeleton, issue #475 AC6).

The decline arm's rationale — a denial record without a cost record *"has no analysis join"*
— is **right about analysis and wrong about forensics**. The record's primary job is telling
an operator, after the fact, what the harness refused on a run that produced no verdict; an
unjoinable denial record still beats a vanished one. That reasoning now sits in a code
comment where the decline arm's comment was.

### Second change: the PR operand was coupled to the cost operand

Verifying the inherited gates surfaced a coupling that would have made the skeleton **inert
on the very path it targets**. `prepare-harness-floor.sh` short-circuited its PR resolution
on every cost-inert branch (`_emit "" "$CLASS"` — empty PR), from back when
`DEVFLOW_EXECUTION_PR` was the cost floor's operand alone. So on the all-null-cost path
*both* `DEVFLOW_EXECUTION_COST` **and** `DEVFLOW_EXECUTION_PR` were empty, and a faithfully
mirrored skeleton would have declined at its PR gate every time.

The operands fail independently — a run can die before writing cost figures and still have
produced denials worth persisting — so the cost-inert arms now record the cost as inert and
**fall through** to the PR resolution instead of exiting. The cost side is unchanged:
`apply_harness_floor` returns at its first guard on an empty cost, so no all-null
`harness_cost` and no cost skeleton is ever staged. That is asserted as an explicit negative
control, not just reasoned about.

## What this does and does not close

It **closes the all-null-cost drop path**. It is **not** an unconditional guarantee that
every denied permission request reaches durable storage. The mirrored skeleton inherits
`apply_harness_floor`'s two gates, and each remains a real drop path — both verified in the
source, both named in a code comment beside the arm, both taking a named breadcrumb rather
than a silent drop, and both driven by tests:

1. **`DEVFLOW_COMMAND_CLASS` is `pr-description` or does not classify.** Those classes derive
   no record at all, so there is no keyed store to write into.
2. **`DEVFLOW_EXECUTION_PR` is empty.** The record is keyed `<slug>-<run-id>.json` with slug
   `pr-<N>`; with no PR number there is no identity to file it under.

**How often the decline arm was actually reached is unestablished** and cannot be established
from this tree. No estimate is offered, and the absence of these breadcrumbs is not evidence
of zero.

## Downstream consumer claim — verified, holds

`scripts/build-experiment-records.py`'s `_efficiency_entry` requires only a top-level string
`slug`, so a denial-only record ingests cleanly: `_run_cost` reads `record["telemetry"]` and
returns `None` for an empty list, `_telemetry_complete` returns `False`, and
`permission_denials` flows through verbatim to `_denials_from_eff`, which reads only
`.count`. This is asserted by driving the real module functions over a real skeleton, not
by inspection — a record that broke ingestion would be worse than none.

## `efficiency_telemetry_enabled` also gates denial persistence

`apply_denial_floor` is gated on `prflow_review_and_fix.efficiency_telemetry_enabled` exactly
as record derivation is (verified in `lib/efficiency-trace.sh`). Its schema description and
the `docs/efficiency-trace.md` config table mentioned only trace rendering and
`.prflow/logs/efficiency/` writes, so a consumer disabling it for cost reasons had nothing in
the key's name or description warning them they were also giving up denial forensics. Both
now say so.

## Tests

All behavioral, in `lib/test/modules/efficiency-trace-telemetry.sh`'s existing `#1064 AC3/W1`
block (which already drove the merge/decline arms). The decline-arm assertions are replaced
by, and the glue assertions extended with:

- the skeleton is written for a record-deriving class with a PR, with the asserted record
  shape and the denial record carried verbatim (and no fabricated `harness_cost` beside it);
- that skeleton ingests downstream and yields its count through `_denials_from_eff`;
- re-running the backstop over the persisted skeleton is a tree-equality no-op;
- the overwrite guard declines when a real, populated record occupies the skeleton's own
  filename and `list_blobs` silently returned empty (the patched-lib fixture pattern the cost
  floor's own guard test uses);
- both residual gates: no skeleton written, named breadcrumb, for `pr-description`, for an
  unclassified class, and for an empty PR;
- the glue now resolves the PR on the absent-file, unparseable and figureless branches while
  the cost handoff stays empty, plus the negative control that a resolved PR with an empty
  cost writes no cost skeleton.

`lib/test/run-module.sh efficiency-trace-telemetry` → **920 passed, 0 failed** (post-commit).
`shellcheck --severity=warning -e SC1091` over the tracked non-`lib/test` shell surface and
`ruff check` over the tracked Python surface are both clean.
`scripts/devflow-cloud-writer-contract.json` regenerated in the same commit (it SHA-256-pins
`lib/efficiency-trace.sh`).

No issue exists for this work; the problem and its evidence are described above.

### DIFF
--- a/.changeset/denial-record-skeleton-arm.md
+++ b/.changeset/denial-record-skeleton-arm.md
@@ -1,0 +1,33 @@
+---
+bump: patch
+type: Fixed
+---
+
+- **A permission-denial record is no longer discarded when the run produced no usable
+  cost figures.** Denial forensics ride the per-run efficiency record as its
+  `permission_denials` key, and `apply_denial_floor` could only ever *merge* onto a host
+  record — so a run that died before yielding cost figures lost its fully-built denial
+  record entirely. `scripts/prepare-harness-floor.sh` refuses to stage an all-null
+  `harness_cost`, which empties the cost handoff, which returns `apply_harness_floor`
+  before its own skeleton arm, so no host record existed to merge onto and the denial
+  record was dropped with only an expiring job-log warning — precisely the stall / crash /
+  execution-ceiling class where denial forensics matter most. The denial floor now has a
+  skeleton arm mirroring the cost floor's (same run-id targeting, same overwrite guard,
+  same minimal `synthesized: true, source: null` shape), so the record is written rather
+  than declined. An unjoinable denial record still beats a vanished one. This **closes the
+  all-null-cost drop path**; it is not an unconditional guarantee — the mirrored skeleton
+  inherits the cost skeleton's two gates, and each remains a named, breadcrumbed drop
+  path: a `pr-description` or unclassified command class, and an unresolvable PR number
+  (the record has no slug to be keyed by). How often either is reached is unestablished.
+- **The PR number is now resolved even when the cost is inert.**
+  `scripts/prepare-harness-floor.sh` short-circuited its PR resolution on every
+  cost-inert branch, back when `DEVFLOW_EXECUTION_PR` was the cost floor's operand alone.
+  It has a second consumer now, and the two operands fail independently — so the
+  short-circuit made the denial skeleton unreachable on exactly the path it exists for.
+  The cost side is unchanged: `apply_harness_floor` still returns at its first guard on an
+  empty cost, so no all-null `harness_cost` and no cost skeleton is ever staged.
+- **`efficiency_telemetry_enabled` now documents that it also gates denial forensics.**
+  Its schema description and the `docs/efficiency-trace.md` config table mentioned only
+  trace rendering and record writes, so a consumer disabling it for cost had nothing
+  warning them that they were also giving up the durable record of what the harness
+  refused.
--- a/.prflow/config.schema.json
+++ b/.prflow/config.schema.json
@@ -250,7 +250,7 @@
         },
         "efficiency_telemetry_enabled": {
           "type": "boolean",
-          "description": "When true, /prflow:review-and-fix derives a per-subagent effectiveness trace at Loop Exit, prints it to chat, and persists one per-run record to .prflow/logs/efficiency/. When false, the loop neither renders the trace nor writes any file under .prflow/logs/. Derivation is best-effort: a failure never aborts the fix loop.",
+          "description": "When true, /prflow:review-and-fix derives a per-subagent effectiveness trace at Loop Exit, prints it to chat, and persists one per-run record to .prflow/logs/efficiency/. When false, the loop neither renders the trace nor writes any file under .prflow/logs/. Derivation is best-effort: a failure never aborts the fix loop. This key ALSO gates permission-denial forensics: the durable record of what the harness refused on a run (its count, denied tool names and scrubbed command text) is persisted as the permission_denials key on that same per-run record, so setting this to false to save telemetry cost silently gives up denial forensics too -- a run that ends with no verdict then leaves only an expiring job log.",
           "default": true
         },
         "efficiency_cut_candidate_min_dispatch": {
--- a/docs/efficiency-trace.md
+++ b/docs/efficiency-trace.md
@@ -295,7 +295,7 @@ Both live under `prflow_review_and_fix` in `.prflow/config.json`
 
 | Key | Type | Default | Effect |
 |---|---|---|---|
-| `efficiency_telemetry_enabled` | boolean | `true` | Master gate. When `false`, the loop renders no trace and persists no effectiveness record to the telemetry branch. |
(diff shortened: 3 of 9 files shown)


The01Geek:
Review fixes applied in `445181a`.

## Applied — the Suggestion (executable regression on the denial gate)

This PR documented in `.prflow/config.schema.json` and `docs/efficiency-trace.md` that `prflow_review_and_fix.efficiency_telemetry_enabled` gates **denial** persistence as well as cost, and nothing executable held that claim: the only flag-off arm (`A8`) drives `apply_harness_floor`, so `apply_denial_floor`'s own gate was unexercised. Since this PR created the gap, it is the right place to close it.

Added a `df-gate` block to `lib/test/modules/efficiency-trace-telemetry.sh`, mirroring `A8`'s structure (same `DEVFLOW_CONFIG_FILE` fixture with `efficiency_telemetry_enabled: false`). Every input the skeleton arm needs is supplied — operand, `DEVFLOW_EXECUTION_PR=42`, `DEVFLOW_COMMAND_CLASS=review-and-fix`, `GITHUB_RUN_ID` — and the neighbouring `df-skeleton` block proves that exact shape *does* write a skeleton with the flag on, so the gate is the only thing standing between the fixture and a write.

Both halves asserted:

- **no write** — nothing under `.prflow/logs/efficiency/` reaches the telemetry branch (the broader `ls-tree` form the `df-residual` arms use, so a skeleton under any filename would fail it, not just `pr-42-646-1.json`);
- **the breadcrumb** — the denial floor's own `denial floor: efficiency telemetry is disabled` emit, matched against the helper's real stderr rather than file text, so this is a behavioral assertion and not a wording-only pin.

### Coupled floor literals

`efficiency-trace-telemetry` carries `assertion_floor_policy: "exact"`, so both equality-enforced literals moved **920 → 922** in the same commit: the `lib/test/run.sh` call site and `scripts/workflow-flight-recorder-registry.json`. Confirmed by running the module directly — `Module efficiency-trace-telemetry: 922 passed, 0 failed`.

`scripts/devflow-cloud-writer-contract.json` needed **no** regeneration: none of the three touched files is a pinned asset, and `python3 lib/test/cloud_writer_contract.py generate` is a byte no-op on this tree. No changeset amendment either — this is a test-only addition to a PR that already carries one.

Verified: `lib/test/run-module.sh efficiency-trace-telemetry` (922/0), `shellcheck --severity=warning -e SC1091` on the module and (with `--extended-analysis=false`) on `run.sh`, and the `#810` gate `pin-corpus-lint.py mutation-routing-worktree` clean.

## Declined

- **Minor — the skeleton-write failure `else` branch (jq/`printf` failure → warning + `rm -f`) is unexercised.** Declined as low value: the operand is already validated as a JSON object before control reaches that point, so the branch guards a residual failure mode (a broken `jq` binary, a write error) rather than any input shape a caller can produce. Exercising it needs a stubbed-`jq`/read-only-filesystem fixture whose cost is out of proportion to the risk it retires.
- **Trivial — only `review-and-fix` exercises the record-deriving class arm; `review`/`implement` are the same no-op fall-through.** Declined: those two classes take a byte-identical path through the same `case`, so additional rows would assert the enumeration rather than any distinct behavior, and the arm's *negative* side (`pr-description` and the unclassified empty string) is already covered by the `df-residual` loop.

Neither is load-bearing on inspection; if either reading is wrong I would rather be told than have guessed.

PR #1308, opened 2026-08-05
test: measure the harness hook surface (PermissionRequest / PreToolUse deny / defer probe arms)

The01Geek:
Adds three `matcher-probe.yml` jobs that measure three currently-unestablished facts about the cloud harness's hook surface, so a downstream design decision can rest on evidence instead of inference. **Measurement only — no production behavior changes.** No issue exists yet; the rationale is here.

### 1. `permissionrequest-probe` — does `PermissionRequest` fire, and *which calls reach it*?

The Agent SDK type contract the action depends on (`@anthropic-ai/claude-agent-sdk@0.3.222`) publishes a `PermissionRequest` hook returning a decision object (`{behavior:"allow"…}` / `{behavior:"deny",message?…}`). Two things are unknown, and the second is why the job exists:

- **(a)** Does it fire at all under the CLI `claude-code-action@v1` installs?
- **(b)** *Which* calls reach it. The published six-step permission order — hooks, deny rules, ask rules, permission mode, allow rules, `canUseTool` — does not place `PermissionRequest` anywhere. If it sits at the `canUseTool` slot it sees only calls the allowlist did **not** resolve, and an unconditional-deny hook is harmless to granted work. If it resolves earlier it also sees calls the allowlist **would have approved**, and such a hook silently blocks granted work.

So the job is built around a **granted control command**: the session attempts `mkdir -p …` (the one granted head) and then `basename …` (in no allowlist), and the hook appends its raw stdin payload — which carries the tool input — to a JSONL breadcrumb before denying. The breadcrumb's *content* attributes each firing. A hook that fires on the granted control answers (b) negatively and kills the downstream design.

It also measures **(c)**: does the deny `message` reach the engine transcript at all?

### 2. `pretooluse-deny-probe` — does a `PreToolUse` **deny** deliver its reason?

The existing `pretooluse-probe` emits `permissionDecision: "allow"`, for which `permissionDecisionReason` is specified to be **ignored** — so its accumulated observations say nothing about deny-path reason delivery. This job emits a real `deny` with a distinctive sentinel reason on a sacrificial command.

It is a **separate job** on purpose. `PreToolUse` runs before the permission system: sharing a session, a `PreToolUse` deny would resolve the call so `PermissionRequest` never fires (recording a false "not fired"), and a `PreToolUse` allow would rescue the ungranted control and destroy that arm. Within its own job the hook is **command-scoped** — a `case` on its own stdin payload — so it can only touch the command carrying the `ptudprobe-sacrificial` token, and it writes a separate deny-arm breadcrumb so "the hook ran" stays distinguishable from "the hook denied". The granted control's survival is itself an asserted axis.

### 3. `defer-probe` — does `defer` fall through?

`defer` is the token `scripts/pretooluse-shape-guard.py` emits on every fail-open path, on the premise that it means "fall through to the default permission flow" — that file's own header records the premise as an **unestablished assumption**. The published hooks reference says the opposite: the tool does not execute and the process exits with `stop_reason: "tool_deferred"`. If the reference is right, every one of that guard's fail-open paths is in fact fail-*closed*.

This job settles it and **does not touch the guard** — that fix is separate work in which `defer` is load-bearing at several sites.

## Cross-cutting properties

- **A third verdict axis on jobs 1 and 2: is a *hook-issued* deny still visible in `permission_denials`?** If hook denies are invisible to that array, wiring such a hook in production would make the published denial count fall while the measurement silently stopped — an improvement-shaped regression. Measured now rather than discovered later. Both renderers also distinguish "the hook denied and it is not in the array" from "the hook never denied, so this arm measured nothing".
- **Production mode parity — a deliberate deviation from this file's convention, stated in the job comment with its reason.** Every older probe job passes `--permission-mode acceptEdits`; these three pass **none**, matching both live tiers (`devflow.yml` and `devflow-implement.yml` pass none, so they run in `default`). Under `acceptEdits` the permission-mode step resolves *before* allow rules and auto-approves a whole class of calls, so a verdict about which calls reach a hook could be an artifact of the probe's own mode and would not transfer to the tiers the decision is about.
- **Breadcrumb-first verdicts.** Marker files are the primary signal; the execution file is secondary. On the `defer` arm this is essential rather than merely robust: a honored defer may destroy the execution file, so a file-derived verdict would read `unavailable` in exactly the case the job exists to detect. That renderer reports a missing execution file as an *expected* reading that neither confirms nor contradicts the breadcrumb verdict.
- **Deterministic derivation only.** The model's prose is never the measurement — verdicts come from the execution file and on-disk side effects.
- **An established negative is never collapsed onto an unestablished one.** "Marker absent" and "execution file unreadable" render differently everywhere; an unparseable file, a zero-byte file and an unrunnable `jq` all read `unavailable`, and a denial count of zero renders as the digit `0`, never as `unavailable` (and vice versa).
- **"Event absent" vs "nothing reached it" is stated, not left to the reader.** If the ungranted command was denied (visible in `permission_denials`) but no breadcrumb exists, job 1's renderer says so explicitly: that is evidence the installed CLI does not deliver the event, not merely that nothing reached it. With no denials and no trace of the control command it says the opposite — nothing reached the permission system at all (the session may have failed before issuing any tool call, e.g. if the `settings` input was rejected) — and calls the result **unestablished**, not negative.
- **The observed CLI version is recorded beside the run id** in every verdict. The action ref floats (`@v1`), so a verdict taken without it expires silently on upgrade with nothing noticing.

## What's here

| File | Role |
|---|---|
| `.github/workflows/matcher-probe.yml` | the three jobs |
| `scripts/describe-permissionrequest-probe.sh` | job 1's verdict renderer |
| `scripts/describe-pretooluse-deny-probe.sh` | job 2's verdict renderer |
| `scripts/describe-defer-probe.sh` | job 3's verdict renderer |
| `lib/probe-observation.sh` | shared execution-file readers (exec state, CLI version, transcript search, `permission_denials`) |

The renderers follow `scripts/describe-pretooluse-probe.sh`: positional args, source `lib/resolve-jq.sh`, always exit 0, three-state outputs, and `contains` rather than `startswith` when searching the execution file (the issue #908 review's wrapped-envelope regression, regression-locked again here).

## Coupled sites reconciled in this commit

- `lib/test/run.sh` — 48 assertions driving each renderer's full arm matrix: marker present/absent, the granted-vs-ungranted firing discrimination and its two opposite inferences, sentinel present / **wrapped** / absent, unparseable file, zero-byte file, unrunnable `jq`, no-arg → stderr breadcrumb + rc 0 + empty stdout. Plus two YAML-parse assertions over the workflow: all three jobs declared with a `settings`-registered hook, **no `--permission-mode`**, and their own verdict helper invoked; and each `settings` blob parsing as JSON with a non-empty `Bash`-matched hook for its own event.
- `lib/test/modules/coverage-map.json` — `files` entries for the four new helpers. No new `run_sh_blocks` label: the assertion names carry no issue number, so they land in the existing `unlabeled` aggregate.
- `.changeset/hook-arm-matcher-probe-jobs.md` — `bump: patch` (workflows are engine surface). `.claude-plugin/plugin.json` and `CHANGELOG.md` untouched.
- SPDX headers on all four new files.

Verified as **not** coupled, rather than assumed: `lib/capability-profiles.json` (a `settings:` block is in no generated region, and these jobs use inline `--allowed-tools` rather than a composed profile literal — `--check` clean); the workflow flight-recorder registry (lists only the two live tiers); `lib/test/modules/regenerate-artifacts.sh`'s conflict-covered path list (already names `matcher-probe.yml`).

## Local verification before the first push

`actionlint` (no new findings — the single `SC2129` it reports pre-exists on `main`), `shellcheck --severity=warning -e SC1091` over every tracked `.sh` plus the `--extended-analysis=false` pass on `lib/test/run.sh`, `ruff` over every tracked `.py`, YAML parse + JSON parse of each `settings` blob, an execution of all three hook commands against synthetic hook payloads (each emits valid decision JSON and writes the expected breadcrumbs; job 2's hook correctly declines the control command), the new `run.sh` block in isolation (48 passed, 0 failed), the `regenerate-artifacts` / `capability-profiles` / `workflow-flight-recorder` focused modules, `coverage_map_guard.py`, `coverage-map-retention-check.py`, `cloud_writer_contract.py check`, `generate-capability-profiles.py --check`, the `#810` mutation-routing worktree gate, and the `lint-tree-enumeration` / `lint-carveout-guard` / `lint-shipped-pruned-path` / `lint-gh-api-repo-path` desk lints.

## Cost note

A `pull_request` touching this file re-fires **every** probe job, each a paid Claude session. The branch was therefore pushed first without a PR, and the PR opened once.


The01Geek:
## First live run — two of three questions settled

Run [30966800385](https://github.com/The01Geek/prflow/actions/runs/30966800385), observed CLI version **`2.1.222`**, at `de92754`.

### `pretooluse-deny-probe` — decisive

```
- hook firing: FIRED-AND-DENIED
- deny honored: DENY-HONORED       (sacrificial side effect absent)
- granted control: CONTROL-RAN     (command scoping held)
- deny reason in transcript: REASON-DELIVERED
- hook deny in permission_denials: HOOK-DENY-NOT-RECORDED (count: 1)
```

A `PreToolUse` **deny is honored** and its `permissionDecisionReason` **does** reach the execution transcript — the axis the allow-only `pretooluse-probe` structurally cannot answer. The command-scoped hook left the granted control untouched.

The denials row is the interesting nuance: **count 1, but no entry carrying the reason sentinel.** The only refusable call in that session was the sacrificial one, so the hook deny almost certainly *did* produce a `permission_denials` entry — just one that does not carry the reason text. The renderer could not say that outright, which is a real gap in it, so the follow-up commit adds a second denials axis that looks for the denied **command** as well as the sentinel.

### `defer-probe` — decisive, and it contradicts a shipped assumption

```
- hook firing: FIRED
- defer behavior: DEFER-BLOCKED    (granted command's side effect absent)
- stop reason (secondary): STOP-REASON-DEFERRED
- command issued (secondary): yes
- permission_denials entries: 0
```

**`defer` does not fall through.** The hook fired, the granted command was issued, and it did not execute; the transcript carries `tool_deferred`. The published reference is right and the premise recorded in `scripts/pretooluse-shape-guard.py`'s header is wrong — its fail-*open* paths are in fact fail-*closed*. (Settling that is all this PR does; repairing the guard is separate work in which `defer` is load-bearing at several sites.)

The execution file survived the honored defer on this run, so both axes agreed — but the breadcrumb-first design is what made the verdict independent of that.

### `permissionrequest-probe` — not settled, and the renderer said so

```
- hook firing: NOT-FIRED
- granted control: CONTROL-RAN
- deny message in transcript: SENTINEL-ABSENT
- hook deny in permission_denials: NOT-APPLICABLE (count: 0)
- ungranted arm in permission_denials: no
- inference: no breadcrumb and zero denials, though the transcript is readable —
  no call was refused, so nothing was ever offered to a PermissionRequest hook.
  The event's availability is UNESTABLISHED, not negative.
```

The session ran the granted control and answered `DONE` **without ever issuing the ungranted command** (`num_turns: 3`, `permission_denials_count: 0` — one tool call). Nothing reached the permission system, so the arm measured nothing. The renderer refused to launder that into "the event does not exist", which is the behavior it was built for, but the question stands.

Two things *are* established from it: the `settings` input carrying a `PermissionRequest` hook was **parsed and merged without error** (`Parsed settings input as JSON` / `Merged settings with input settings`), so the registration itself is not rejected; and the hook did **not** fire on a granted call — which is the safe direction, but confounded with "the event is absent" while no ungranted call was ever offered.

## What the follow-up commit changes

1. The prompt now brackets the ungranted command between **two** granted controls, and says plainly that the model must not adjudicate permissibility for itself. The post-arm control makes a skip *attributable*: a run in which it executed while nothing was refused is a session skip, not an absent event. `--effort medium` for this job alone — the failure mode is instruction-following, and it voids the arm silently.
2. A new `devflow_probe_tooluse_has` reader answers "was this command issued" from recorded **tool-call inputs**, not the transcript at large, and every attempt axis in all three renderers now uses it. The transcript carries the run's own prompt, which necessarily quotes the very command tokens those axes search for — the old any-string search would have reported a confident `ATTEMPTED` for a command nothing ever issued. It returns `unavailable`, never `no`, when no tool-call inputs were recorded at all.
3. The deny probe gains the denied-**command** axis described above.

That is push 2 of 2. Every push to this file re-fires all fifteen probe jobs, so the fixes were batched rather than iterated.

The01Geek:
## All three questions settled

Run [30967680822](https://github.com/The01Geek/prflow/actions/runs/30967680822), observed CLI version **`2.1.222`**, at `24976a8`.

### 1. `PermissionRequest` — the event is not delivered, and it did **not** fire on the granted control

```
- hook firing: NOT-FIRED
- granted control: CONTROL-RAN
- ungranted arm attempt: ATTEMPTED
- ungranted arm outcome: UNGRANTED-REFUSED
- post-arm control: AFTER-CONTROL-RAN
- deny message in transcript: SENTINEL-ABSENT
- ungranted arm in permission_denials: yes (count: 1)
- inference: a call the allowlist did not resolve DID reach the permission system
  and was REFUSED, yet no hook breadcrumb exists — so this is evidence that the
  installed CLI does not deliver a PermissionRequest event, NOT merely that
  nothing reached it.
```

Every precondition the negative needs is now positively established rather than assumed:

- the `settings` input carrying a `PermissionRequest` hook was **parsed and merged without error** (`Parsed settings input as JSON` / `Merged settings with input settings`), so this is not a rejected registration;
- a granted call **executed**, so the session was live;
- the ungranted call was **issued** (a recorded tool-call input carries it) and **refused** (its on-disk side effect is absent, and `permission_denials` names it) — so a call the allowlist did not resolve genuinely reached the permission system;
- the session **got past** the arm (the post-arm control ran), so nothing was truncated;
- and the hook wrote **nothing**, on either call.

**(a) does it fire at all — no**, on `2.1.222` under `claude-code-action@v1`. **(b) which calls reach it —** moot, but the specific thing the downstream design feared is directly observed not to happen: `PermissionRequest` did **not** fire on the granted control. The dangerous configuration is not present; there is simply nothing to configure.

**Incidental observation, worth a caveat rather than a conclusion.** Push 2 used `basename <token>` as the ungranted command; that run recorded **zero** `permission_denials` even though the command was issued and the session got past it. Push 3 swapped in `touch <path>` — equally ungranted — and the refusal was both effective and recorded. The probe cannot say what happened to the `basename` call (it has no side effect to check), so this is an open thread, not a finding. It is exactly why the arm no longer infers "refused" from an absent denials entry.

### 2. `PreToolUse` deny — reason delivered, deny honored, and the deny **is** visible to a denial count

```
- hook firing: FIRED-AND-DENIED
- deny honored: DENY-HONORED
- granted control: CONTROL-RAN
- deny reason in transcript: REASON-DELIVERED
- hook deny in permission_denials: HOOK-DENY-NOT-RECORDED (count: 1)
- denied command in permission_denials: COMMAND-RECORDED
```

A `PreToolUse` **deny is honored** and its `permissionDecisionReason` **does** reach the execution transcript — the axis the allow-only `pretooluse-probe` structurally cannot answer, since `permissionDecisionReason` is ignored on an `allow`. The command-scoped hook left the granted control untouched.

On the cross-cutting question: the denials array carries **no entry bearing the reason sentinel**, but it *does* carry **an entry naming the denied command**. So wiring such a hook in production would **not** silently drop the published denial count — the improvement-shaped regression this axis existed to catch is not present. What is lost is the *reason text*, which never reaches the array.

### 3. `defer` — it does **not** fall through

```
- hook firing: FIRED
- defer behavior: DEFER-BLOCKED
- stop reason (secondary): STOP-REASON-DEFERRED
- command issued (secondary): yes
- permission_denials entries: 0
```

The hook fired, the granted command was issued, and it **did not execute**; the transcript carries `tool_deferred`. The published reference is right and the premise recorded in `scripts/pretooluse-shape-guard.py`'s header — that `defer` falls through to the default permission flow — is **wrong on this CLI**. Its fail-*open* paths are in fact fail-*closed*.

Settling that is all this PR does. Repairing the guard is separate work in which `defer` is load-bearing at several sites, and nothing here touches it.

(The execution file survived the honored defer on all three runs, so both axes agreed — but the verdict never depended on that, which is the point of the breadcrumb-first design.)

---

Verdicts 2 and 3 reproduced identically across three independent runs (`de92754`, `b0dadbf`, `24976a8`). All are scoped to CLI `2.1.222`; every one of them is re-derivable by dispatching this workflow, and each verdict carries the CLI version it was taken on precisely so it cannot outlive the floating `@v1` ref unnoticed.

The01Geek:
## Review fixes applied

### Fix 1 (Important) — `permission_denials` readers now gate on the array's presence

Correct finding. Both readers in `lib/probe-observation.sh` read the array's **length** without first establishing that any such array exists, so an execution file that parses but carries no `permission_denials` satisfied the exec-state gate and yielded `0` / `false` — and the renderers published `HOOK-DENY-NOT-RECORDED`, `COMMAND-NOT-RECORDED`, "the array under-reports refusals" and `count: 0` out of a quantity nothing measured. That contradicted the file's own header promise of fail-closed `unavailable` on every miss.

Both readers now gate on presence the way `devflow_probe_tooluse_has` gates on an empty tool-call population, giving **three states**:

| execution file | count | `_have` |
|---|---|---|
| `permission_denials` present, with entries | that count | `yes` / `no` |
| `permission_denials` present, empty | `0` (a real measurement) | `no` (established negative) |
| no `permission_denials` array anywhere | `unavailable` | `unavailable` |
| present but **not an array** | `unavailable` | `unavailable` |

Presence is "some object carries a `permission_denials` whose value is an *array*" — strictly stronger than `has("permission_denials")`, so a field that is present but restructured also reads unestablished rather than being skipped into a zero.

**Proof it discriminates.** Fixtures for all three states are driven through all three renderers. Reverting only `lib/probe-observation.sh` to the pre-fix version turns exactly the 12 new denials rows RED (including "an absent `permission_denials` array is never rendered as `HOOK-DENY-NOT-RECORDED`", the reviewer's named false verdict) while the other 77 stay green; with the fix, 89 passed / 0 failed. The `unavailable` prose on each affected axis now names the absent-array case.

### Fix 2 — three comment inaccuracies

- `matcher-probe.yml` Job 1 header: ungranted arm corrected from `basename` to `touch`, pointing at the Action 2 note that records why the on-disk side effect is load-bearing.
- `describe-permissionrequest-probe.sh` usage synopsis: added the missing required `<ungranted-marker>` positional.
- **Run-id attribution — both citations were already right.** A matcher-probe run id names a whole *workflow* run, not one job, and run `30966800385` contains both jobs. Verified against the job logs: its `permissionrequest-probe` job recorded `NOT-FIRED` with `permission_denials_count: 0`, and its `pretooluse-deny-probe` job recorded `HOOK-DENY-NOT-RECORDED … (count: 1)`. Nothing was mis-attributed; the citations were merely ambiguous, so both now name their job explicitly so the misreading is not repeated.

### Fix 3 — the possibly-dead branch: confirmed unreachable, removed

`describe-permissionrequest-probe.sh` Axis-5's `CONTROL_ATTEMPTED = no && AFTER_RAN = no` arm sits behind `UNGRANTED_OUTCOME = unavailable`, which requires `devflow_probe_tooluse_has` to return `unavailable` for the ungranted needle. Every condition producing that return is **needle-independent** (unreadable/unparseable file, jq not runnable, or no tool-call inputs recorded at all), and `CONTROL_ATTEMPTED` is read from the same file by the same function with a non-empty needle — so it is `unavailable` there too and can never read `no`. The arm is removed; the surviving `else` carries an authoring note stating the mechanism so it is not re-added.

### Secondary test gaps — added

- The positive `HOOK-DENY-RECORDED` direction in **both** renderers (so the now-three-state axis is covered in all three directions).
- `HOOK-SAW-BOTH`, including its inference line.
- `CONTROL-BLOCKED` in **both** renderers — the only one of the three control verdicts that would mean a hook blocked *granted* work.
- The `cli_version` garbage-version fail-closed arm, plus a negative control that the rejected string is not echoed into the step summary.
- A present-but-wrong-type `permission_denials` field.

### Skipped, with reasons

- **No probe job, prompt, hook command or measured quantity was changed** — per the constraint, the verdicts already recorded on this PR stay attributable to the shipped arms. Only renderers, comments and tests changed.
- `docs/cloud-allowlist.md` (verdict transcription is a separate PR), `scripts/pretooluse-shape-guard.py` (a concurrent PR owns it) and `CLAUDE.md` left untouched.
- The pre-existing `actionlint` SC2129 style finding at `matcher-probe.yml` line ~1433 is left as-is: it reproduces identically on `origin/main` and is outside this PR's diff.

### Verification

`origin/main` merged in (the branch was 8 behind). Locally: the hook-arm block driven under its own harness (**89 passed, 0 failed**, plus the RED-first revert above); `lib/test/test_python_scripts.py` **3675 passed, 0 failed**; `lib/test/run-module.sh harness-python-guards` **45 passed, 0 failed**; the `#810` mutation-routing worktree gate clean with its completion sentinel; `shellcheck` clean on both scopes; `ruff` clean. One push, so the probe jobs re-fire once.

PR #1311, opened 2026-08-05
fix: the PreToolUse shape guard's fall-through is exit 0 + no stdout, not `defer`

The01Geek:
## The problem

`scripts/pretooluse-shape-guard.py` is a `PreToolUse`/`Bash` hook. Its header documented
`defer` as its **fail-open majority path** — the decision returned for every command it does
not recognize — on the belief that `defer` means "fall through to the normal permission
flow". Six sites emitted it: a non-UTF-8 stdin, an empty stdin, an unparseable JSON payload,
a payload `_read_command` cannot classify (non-`Bash` or malformed), a command matching no
deny-set arm, and `main()`'s blanket exception handler.

## The measurement

Run [`30967680822`](https://github.com/The01Geek/prflow/actions/runs/30967680822) (branch
`hook-probe-arms`, PR #1308) added purpose-built probe arms and answered it:

| Arm | Job | Verdict |
| --- | --- | --- |
| `defer-probe` | `92185120496` | **`DEFER-BLOCKED`** — the hook fired and the granted command's side effect `.prflow/tmp/deferprobe-tool-ran` is **absent**, so the tool did not execute. Secondary: **`STOP-REASON-DEFERRED`** (the transcript carries `tool_deferred`). |
| `pretooluse-deny-probe` | `92185120507` | **`DENY-HONORED`** + **`CONTROL-RAN`** + **`REASON-DELIVERED`** — `deny` is honored, a non-matching command is left alone, and `permissionDecisionReason` reaches the transcript. Also **`COMMAND-RECORDED`**: a `permission_denials` entry names the denied command, though the reason text is not in that entry. |

Both arms report `observed CLI version: 2.1.222`.

So on the CLI the action installs, **every one of the guard's "fail-open" paths was in fact
fail-closed** — a wired run would have ended on the first command the guard did not
recognize. The documented true fall-through is exit 0 with no stdout: *"Exit code 0 with no
output means the hook has no decision to report, so the tool call continues through the
normal permission flow."*

## What this changes

**All six `defer` emissions become the real fall-through** — exit 0, empty stdout — routed
through a single named `_emit_no_decision()` so the guard's whole fail-open surface is one
greppable shape. `deny` is now the only token the guard ever writes, and it is the token the
run measured as honored with its reason delivered.

**The header's reasoning is rewritten, not left stale.** It argued the `defer` majority path
was load-bearing because the harness caps consecutive hook blocks. That was unsound in both
halves: `defer` *was* a block, so the majority path bounded nothing, and no evidence in this
repository establishes such a cap. The no-decision fall-through makes the bound real
*without* depending on the cap existing — a hook that reports no decision has not blocked
anything, so it cannot count toward such a cap if the harness has one. Blocking is now
confined to exactly the deny-set arms `_matched_arms` returns, which is the property the
bound was always meant to name.

**The `UNESTABLISHED … permissionDecision VOCABULARY` block is replaced by what is now
established**, citing the run id and both job ids. Its pointer at the existing
`pretooluse-probe` arm was always wrong — that arm emits `allow`, so it can observe neither
`deny` nor `defer`. The block records that the verdicts were taken on **CLI 2.1.222** and
that **`claude-code-action@v1` floats**, so a future reader knows they can expire on upgrade;
the genuinely open item (what the harness does with an *unrecognized* token) stays marked
open.

## Coupled sites reconciled in the same commit

- **`scripts/harden-stop-hooks.sh`'s `STUB_PY`** printed the same `defer` object as its
  "benign" no-op for a `.py` entry target. A stub emitting it would have terminated the very
  run it exists to leave untouched. It now prints nothing and exits 0 — what the bash `STUB`
  does in bash.
- **`lib/test/run.sh`**'s stub assertion grepped that stub's stdout for the literal
  `"permissionDecision": "defer"`. It now asserts **empty stdout and rc 0**.
- **`lib/test/test_python_scripts.py`**'s rig reported a stripped-empty stdout as
  `PARSE-FAIL`. It now reports a distinct `_NO_DECISION` sentinel — "wrote nothing" and
  "wrote something unparseable" are opposite verdicts here, and collapsing them would let a
  decision-object regression pass every fall-through assertion. Every `'defer'` assertion in
  the file reads the sentinel instead, and the assertion on the guard's stderr breadcrumb
  tracks its reworded text.

## Tests

New byte-level coverage of the fall-through contract (`raw_stdout`, no decoding, no
normalization), on **both structurally distinct routes** — `_run`'s own early return and
`main()`'s blanket handler, which a stubbed dependency drives — asserting **zero stdout bytes
and exit 0**. A negative control on the same rig proves a denied shape still emits its
decision object with `hookEventName: PreToolUse`, `permissionDecision: deny`, and a non-empty
`permissionDecisionReason`, so the emptiness is attributable to the fall-through and not to a
guard that writes nothing at all.

## Scope

No `.github/workflows/**` edits — the wiring decision stays open; this only makes the guard
correct if and when it is wired. `docs/cloud-allowlist.md` is untouched (transcribing the
probe verdicts is a separate PR), as is `CLAUDE.md`. `scripts/devflow-cloud-writer-contract.json`
pins none of the touched files, so no regeneration was needed.

## Verification

- `lib/test/test_python_scripts.py` (the guard's `focused_test` per `lib/test/modules/coverage-map.json`): **3680 passed, 0 failed**
- `ruff check` on both touched Python files: clean
- `shellcheck --severity=warning -e SC1091` on `scripts/harden-stop-hooks.sh` and, with `--extended-analysis=false`, on `lib/test/run.sh`: clean
- `pin-corpus-lint.py lint` and the `#810` `mutation-routing-worktree` gate: rc 0 (`MUTATION-ROUTING-STATIC-SCAN-COMPLETED`)
- The rewritten `run.sh` stub assertion was driven directly against a freshly installed stub: empty stdout, `rc=0`


The01Geek:
**Reception pass on the four review findings (commit `102c02a0a`)** — this repo's `fix_severity_threshold` is `suggestion`, so all four re-opened the diff and all four are addressed:

1. **Partial-deny → `main()`-handler path now tested.** The rig patches its *own copy* of the guard so `_emit` writes a partial object and then raises; the assertion is byte-exact on the partial prefix (`{"hookSpecif`), so a handler that emitted anything would append to it and fail for the reason it pins. Carries a positive control on the same patched rig.
2. **Comment corrected + real coverage added.** `{not json` is relabelled the json-parse-failure route (it never reaches `_read_command`), and a new byte-level row drives the genuine `_read_command → None` route with a valid-JSON non-`Bash` payload whose command *is* a denied shape — so removing the `tool_name` check turns it red.
3. **Probe-arm provenance stated.** The docstring now says neither `defer-probe` nor `pretooluse-deny-probe` is merged (verified: `matcher-probe.yml` at HEAD carries only `pretooluse-probe`; both arms exist on `hook-probe-arms` at `710767c`), so re-running them is not actionable from this tree alone.
4. **Count-rot removed.** "six sites" → count-free wording in both `_emit_no_decision()`'s docstring and the changeset.

**Deferred, recorded here:**
- **Mutation-check of the two new tests was not executed.** WHY: the operator directed this session to use CI rather than local suite runs; the mutation run was interrupted and the working tree was restored clean (verified `git status` empty). The tests are argued discriminating by construction above rather than demonstrated RED. **Revisit condition:** before relying on either test as a regression guard, run the two mutations (reintroduce an `_emit(...)` in `main()`'s blanket handler; drop the `tool_name != "Bash"` check) and confirm each fails on its own assertion.
- **No verification record was written**, so `check-completion-evidence.py --context-mode direct` returns `completion-check: missing-evidence — verification-record: reference not supplied`. CI is green on this head (all five shards + lint + `lib + python tests`), and the focused `lib/test/test_python_scripts.py` passed locally at `3684 passed, 0 failed` before the mutation attempt. **Revisit condition:** none if CI stays green; the gap is the recording channel, not the evidence.

Issue #936, opened 2026-07-29
Withhold the auto PR-triggered review tier from this release: remove devflow-review.yml from main and stop shipping the three-file tier


The01Geek:
## Problem Statement

DevFlow is being submitted to the Claude Code Community Directory. The auto PR-triggered review tier is the one surface in this repo that a security reviewer will flag, and #930 and #920 confirm the flag would be correct.

`.github/workflows/devflow-review.yml` triggers on `pull_request`, `pull_request_target`, `check_run`, `workflow_run`, `check_suite` and `status`, calls a reusable workflow with `secrets: inherit`, checks out the pull-request head, and carries no actor-authorization gate. The `status` arm is the widest of them: it fires for any commit status posted by any app, with no branch scoping of its own. Its own enable toggle is read from the PR's tree, so a fork PR supplies the config that decides whether it is reviewed at all.

Two open issues cover the fix and neither is close to landing:

- **#930** — `precheck` performs a bare `actions/checkout`, which under the `pull_request` trigger resolves the PR merge ref, so the config gating a review comes from the PR under review. The fix needs a trusted base-ref read plus a bootstrap "neutral check when no config exists" arm for which the issue records no established prior art across policy-bot, Allstar, Renovate, Dependabot, scorecard-action or harden-runner, and merge-queue parity for `neutral` is unverified. Its acceptance criteria are self-declared unaudited.
- **#920** — blocked on #930, and carries a live Blocked section that cannot be discharged locally: it is unknown whether `GET /repos/{owner}/{repo}/collaborators/{username}/permission` succeeds under `precheck`'s `pull-requests: read` token, and discharging it requires a probe run on a real fork-originated `pull_request` event. Even after the fix, fork `pull_request` events receive a read-only `GITHUB_TOKEN` regardless of the `permissions:` block, so `create_check` cannot POST and the required context goes unreported — #920 records that as a terminal bad outcome, not an improvement.

A public directory listing is the worst venue for that behavior, because opening fork pull requests is the first thing strangers do with a listed repository. The immediate business consequence is that the submission is likely to be flagged and denied.

[...]

PR #1315, opened 2026-08-05
docs: replace the guard header's unmerged-arms claim with a re-probe recipe

The01Geek:
## The problem

`scripts/pretooluse-shape-guard.py`'s header carries a `VERDICT PROVENANCE AND EXPIRY`
paragraph whose whole job is to tell a future maintainer **how to re-establish** the
`deny` / `defer` harness verdicts after a `claude-code-action` or CLI upgrade expires
them. It ended by saying they can't be re-established from here:

> Re-running the two arms is how they are re-established after an upgrade — but NEITHER
> ARM IS MERGED: `.github/workflows/matcher-probe.yml` at HEAD carries only
> `pretooluse-probe` [...] and `defer-probe` / `pretooluse-deny-probe` exist only on the
> unmerged `hook-probe-arms` branch. So re-running them is not actionable from this tree
> alone; that branch has to land (or the arms be re-authored) first.

PR #1308 landed that branch. The paragraph now tells the reader the answer is "you can't",
which is precisely backwards — the actionable answer is what the paragraph is for.

## What I verified on `origin/main`

Read from `git show origin/main:.github/workflows/matcher-probe.yml`:

- **Arms present:** `pretooluse-probe`, `permissionrequest-probe`, `pretooluse-deny-probe`,
  `defer-probe` — all four are jobs on `main`. Both arms the paragraph called unmerged are
  there.
- **Triggers:** `workflow_dispatch` (bare, no inputs) and `pull_request` with
  `paths: ['.github/workflows/matcher-probe.yml']`. So any same-repo pull request touching
  that file re-probes automatically, and neither trigger can select an individual job.
- **Cost:** **15 jobs**, **14** of which `uses: anthropics/claude-code-action@v1` — i.e.
  start a paid Claude session. (The fifteenth, `cancel-probe`, is a sleep job.) The
  workflow also carries a `concurrency` group with `cancel-in-progress` keyed on the ref.

## What the paragraph says now

The replacement keeps the paragraph's voice and structure and everything it already got
right — the `2.1.222` CLI provenance, that `claude-code-action@v1` is a floating tag, and
that these verdicts are a measurement of one harness build rather than a standing property.
In place of the stale clause it states the mechanism: the two triggers that reach the arms,
and the cost that governs how the recipe should be used — neither trigger selects a job, so
either fires the whole workflow, which is why a re-probe belongs in one push rather than an
iterated sequence. The `concurrency` cancellation is named honestly (it reclaims the
unfinished remainder, not what cancelled sessions already spent).

Wording follows the repo's non-decay preference: the arms' presence is anchored to the
#1308 merge as a dated event rather than asserted present-tense about a moving `main`, the
job count is likewise dated to that merge, and the trigger description is a property of the
workflow file a reader can check in one look.

## Verification

- `ruff check scripts/pretooluse-shape-guard.py` — clean.
- `lib/test/test_python_scripts.py` (the `focused_test` this file's
  `lib/test/modules/coverage-map.json` entry names) — **3684 passed, 0 failed**.
- **Behavior-inertness proved mechanically, not asserted:** parsed both this revision and
  `origin/main`'s, dropped the module docstring from each AST, and compared the dumps —
  identical. The only bytes that changed are the docstring.

## Coupled sites checked

- `scripts/devflow-cloud-writer-contract.json` does **not** list this file (checked
  case-insensitively), so no regeneration is owed.
- No test asserts the changed text. `lib/test/run.sh`'s one reader of this file
  (`RGV_ARMS_JSON`) extracts `REMEDIATION` keys via `ast`, not source text; the sole
  `lib/test/` hit for a phrase from the paragraph is an unrelated comment in
  `lib/test/test_python_scripts.py` citing CLI `2.1.222`. No new pin was added — new
  prose-presence pins are prohibited.

## Changeset call

**Added one, `bump: patch`.** `CLAUDE.md` names `scripts/` as engine surface and the
Phase 3 review gate FAILs an engine-surface change carrying no changeset; the
internal-only exemption is scoped to tests, CI, and dev-only docs, and this file ships to
consumers through the vendor slice. The change is comment-only with no behavioral effect,
which is the argument for omitting — but that makes it exactly the "genuinely torn" case,
and a changeset is the cheap side of that trade. The entry says plainly that the executable
body is byte-identical, so it does not overclaim in the CHANGELOG.

## Sibling staleness found

**Fixed (in scope):** the one paragraph above. The sweep for arms-unmerged /
`matcher-probe.yml`-at-HEAD claims found no other. The header's other mention of
`hook-probe-arms` — "both in run 30967680822 on branch `hook-probe-arms`" — is a correctly
dated historical attribution of where the run happened and stays true; likewise the
`defer-probe` references in the fall-through section, which #1308 only strengthens.

**Found but NOT fixed — reporting for a separate decision.** The `REGISTRATION IS NOT YET
WIRED` paragraph near the top of the same file says:

> Nothing in the tree registers it: `.claude/settings.json` carries no `PreToolUse` key,
> and `.github/workflows/devflow-runner.yml` passes no `settings` input to the action.

The first half still holds (`origin/main`'s `.claude/settings.json` has no `PreToolUse`
key). **The second half is false at `origin/main`:** `devflow-runner.yml` passes a
`settings:` input registering a `PreToolUse`/`Bash` hook that execs this very guard, and
`lib/test/run.sh` pins that registration. The paragraph's *conclusion* may still hold for a
different reason — per `CLAUDE.md`, `devflow-runner.yml` is the #936-withheld tier with no
reachable entry point — but its stated evidence is stale, and the `TRUST BOUNDARY`
paragraph's "The two registration channels are not [shipped]" inherits the same problem.

I did not touch either. That correction is not what #1308 staled, it needs a determination
about the withheld tier's reachability that I'd otherwise be guessing at, and guessing is
how the defect being fixed here got written in the first place. It deserves its own change.


The01Geek:
## Added: the registration / trust-boundary correction

Folded into this PR (commit `b5d7418`) rather than a follow-up, now that the withheld-tier
question I declined to guess at is settled.

**What was wrong.** `REGISTRATION IS NOT YET WIRED` and `TRUST BOUNDARY` both rested on
"nothing in the tree registers it … `devflow-runner.yml` passes no `settings` input". That
second half is false at `origin/main` — the workflow passes a `settings` input registering a
`PreToolUse`/`Bash` hook that execs this guard, plus a dedicated
`Harden PreToolUse guard closure (unconditional)` step. A reader checking the stated
evidence would find it false and could reasonably conclude the guard is live.

**What it says now.** The conclusion survives — the guard does not execute here — but rests
on the real mechanism: the registration is *shipped but unreachable*, because
`devflow-runner.yml` declares `workflow_call` as its only trigger and nothing in the tracked
tree calls it, its sole caller `devflow-review.yml` having been the auto-review tier
withheld under #936. Written explicitly as **not dead code**: the withheld tier and its
helpers are deliberately retained, and a consumer that installed the tier before it was
withheld still has the caller this tree lacks, so the guard can be live on that copy.

`TRUST BOUNDARY` also dropped its inherited "both channels must land together, or `settings`
alone runs pull-request-editable guard code in a secrets-bearing job". Issue #908 closed
that hole from the other side instead — the unconditional harden step is deliberately *not*
conditioned on the #458 relevance gate, and always materializes a trusted base copy of the
guard's three-file import closure or stubs it fail-closed.

### One correction to the brief I was given

I was asked to write that there is **no committed `.claude/settings.json`**. That is not
true at `origin/main`, so I did not write it:

```
$ git ls-tree origin/main .claude/settings.json
100644 blob 1576c610a1d38603a2d9c123cbe1b52502491181	.claude/settings.json
$ git show origin/main:.claude/settings.json | wc -c
1430
```

The file is tracked; its `hooks` object carries exactly one event, `Stop`, and no
`PreToolUse` key. So the docstring's original wording — "`.claude/settings.json` carries no
`PreToolUse` key" — was **accurate**, and it is the half of that sentence I kept. Only the
`devflow-runner.yml` half was stale. Describing the file as absent would have introduced a
new false claim of exactly the class this PR exists to remove.

### Verification (re-run after the second commit)

- `ruff check scripts/pretooluse-shape-guard.py` — clean.
- `lib/test/test_python_scripts.py` — **3684 passed, 0 failed**.
- Non-docstring AST still byte-identical to `origin/main` — comment-only, both commits.
- Diff against `origin/main` is still just this file plus the changeset; no workflows, docs,
  or `CLAUDE.md`.

The existing changeset was widened rather than duplicated: it now carries the registration
and trust-boundary correction as its own bullet, and states plainly that the change is
comment-only.

The01Geek:
## CI caught one thing; fixed in `bfe64e9`

The monolith shard failed on `#936 surviving devflow-review.yml references match the
checked-in allowlist exactly`. My rewrite named `devflow-review.yml` when explaining why
`devflow-runner.yml` has no caller, and `lib/test/run.sh` pins an exact inventory of every
tracked file still referencing that deleted workflow — this file was not on it.

**Reconciled by rewording, not by widening the allowlist.** This file ships to consumers
through the vendor slice, so adding a surviving reference to a deleted workflow here would
grow the sweep inventory for no reader benefit. The sentence now reads "its sole caller
having been the auto-review tier withheld under issue #936" — the issue number is the
pointer, and nothing is lost.

Worth noting the focused test could not have caught this: the assertion lives in the
monolith shard and reads the committed tree, which is exactly the gap the shard split
exists to cover.

Verified locally by reproducing the assertion's own derivation (parsing the allowlist
heredoc out of `run.sh` and diffing it against the tree's reference set) — exact match.

**CI is now fully green** on the merge commit, monolith shard included. `main` was merged
into the branch mid-review by a freshness update; I re-checked afterwards that the diff
against the updated `origin/main` is still only this file plus the changeset, that the
non-docstring AST is still byte-identical, and that the `#936` assertion still matches.

Issue #1264, opened 2026-08-04
The consumer prompt extension reached the agent in only 8 of 18 review runs and 1 of 4 implement runs, so repo policy is applied non-deterministically


The01Geek:
## Problem Statement

The consumer prompt extension is loaded by the reviewing **agent** running a command, and nothing verifies it happened. In 10 of the 18 most recent `/prflow:review` runs the extension never reached the prompt, so those reviews ran with **none** of this repository's own review policy applied — no `Writing-skills evidence:` gate, no wording-only-pin policy, no declaration-marker limb-one input, no `$PR_BASE_BRANCH` naming rule. The reviews still posted a normal APPROVE/REJECT verdict, and nothing in the verdict, the workflow, or CI distinguishes a policy-applied review from a policy-free one.

This was found from two PRs reviewed three minutes apart that got opposite treatment on the same gate.

**The same non-determinism was subsequently measured on the `/prflow:implement` entry, which is also reached through the `Skill` loader.** Across four sampled cloud implement runs, the extension reached the agent in **one**: run `30983427353` (issue #1206) emitted `.prflow/vendor/prflow/scripts/load-prompt-extension.sh implement` and its workpad carries a `Verification evidence:` record, while runs `30983414046` (issue #1310), `30983423580` (issue #534) and `30983419424` (issue #1261) emitted **no** loader invocation for `implement` at all — in those transcripts the only occurrence of the string `load-prompt-extension.sh implement` is the prose fence in `skills/implement/SKILL.md` itself. The consequence on this entry is wider than on the review entry, because `.prflow/prompt-extensions/implement.md` is the producer of *both* evidence-marker contracts and of the "Prompt-surface edit routing" rule that mandates the writing-skills subagent: PR #1327 edited two prompt-surface files, dispatched no writing-skills subagent, and recorded no marker. Note the failure is **non-emission, not refusal** — `skills/implement/SKILL.md`'s matcher-refusal arm never fires, so even the pending-note record it mandates is absent. The sample is small (4 runs) and the entry demonstrably does work sometimes: the workpads for issues #1333 and #1179 both carry `Writing-skills evidence: ... skill-loaded=yes`, a marker whose contract exists only in that extension.

## Current Behavior

`skills/review/SKILL.md` tells the agent to run `load-prompt-extension.sh review` before doing the skill's work. Whether that happens is entirely at the agent's discretion, and both failure modes are silent:

1. **The agent never attempts the load.** Nothing detects it.
[...]

PR #1342, opened 2026-08-05
Add the issue-1264 render-time placeholder precondition probe

The01Geek:
Discharges the first acceptance criterion of #1264, which is a precondition on that issue's own implementation:

> **Precondition, recorded in this issue thread before implementation begins.** A `matcher-probe.yml` arm is dispatched and its run ID recorded, establishing under `claude-code-action` that (a) a `` !`…` `` placeholder in a plugin-sourced `SKILL.md` is substituted when the prompt is a slash command, (b) the injected command sees `DEVFLOW_PROMPT_EXTENSION_ROOT` as exported through `$GITHUB_ENV`, and (c) rendering is not refused by the resolved `--allowed-tools`. A negative result on any limb routes the work to the workflow-side composition alternative.

Every figure in #1264's "Measured facts" table came from the bare CLI against a throwaway skill; that issue records the `claude-code-action` behavior as explicitly **unmeasured**. So this decides a design, not a detail — which is why the verdict lives in a unit-testable helper rather than inline in YAML.

## Why a throwaway plugin

The measurement needs a **plugin-sourced** `SKILL.md` — a plain local file is not what a consumer runs. This repo *is* the prflow marketplace, so a probe skill under `skills/` would ship to every consumer **and** trip the `lpe-coverage` pin requiring each `skills/*/SKILL.md` to invoke the prompt-extension loader for its own name. The probe therefore carries its own marketplace under `.github/probe-plugin/`, which `install.sh` never copies and the vendor slice never reaches.

## Three limbs, one line

The skill body holds a single placeholder:

```
PHPROBE_LINE_A: !`/bin/echo "PHPROBE_ENV ${DEVFLOW_PROMPT_EXTENSION_ROOT:-UNSET}"`
```

- **(a) substitution** — output present at all.
- **(b) environment** — the value carried is the job's step-level sentinel, the same `$GITHUB_ENV` channel #874 uses for the trusted base-ref closure.
- **(c) allowlist** — the head is `/bin/echo` while `--allowed-tools` grants only `Bash(printf:*),Skill`, so any output at all proves rendering ran an **ungranted** head.

Limb (c) rides on limb (a) by construction: the only head the placeholder could have run is the ungranted one, so there is no "substituted but gated" observation to make.

The **zero-turn abort hazard** (#1264's measured facts: a non-zero exit from an injected command aborts the invocation before the model sees the body) is avoided by construction — `/bin/echo` cannot exit non-zero here, so a missing `PHPROBE_SKILL_REACHED` control points at the harness rather than the command, and the verdict helper says so in that arm.

## Verification

`scripts/placeholder-probe-verdict.py` derives the verdict deterministically from the execution file; the model's prose is never the measurement. Every arm is driven from `lib/test/run.sh` (the `#1264` block) — the three real measurements, all degraded arms resolving `INCONCLUSIVE` rather than a confident negative, both routing lines, plus:

- two **discrimination guards** — the marker in unexpanded template form must not count as a report, and an incidental mention of the placeholder head elsewhere must not forge `NOT_SUBSTITUTED`;
- one **coupling assertion** binding the helper's constants to the workflow job and the probe skill, including that `--allowed-tools` does **not** grant the placeholder's own head. Without that, widening the allowlist would leave limb (c) measuring nothing while every other assertion stayed green.

Locally: all 12 verdict arms return the expected token at exit 0, `ruff` clean, `actionlint` clean, `shellcheck --extended-analysis=false lib/test/run.sh` clean, and the coverage-map guard passes.

This PR **adds** the job and does not run it — maintainer-dispatched, like its probe siblings. The run ID and per-limb result get recorded in the #1264 thread, which is what unblocks that issue.

Writing-skills evidence: .github/probe-plugin/skills/placeholder-probe/SKILL.md mode=inline-degraded skill-loaded=no (the skill authored here is a throwaway measurement fixture in a repo-internal probe plugin that no consumer receives and no runtime path loads, not a shipped prompt surface; the writing-skills discipline governs skills that shape agent behavior in production) guidance-applied=no (no writing-skills guidance was consulted; the body is a scripted probe instruction sequence modelled directly on the existing env-propagation and background-tasks probe prompts) pressure-scenario=no (the body elicits no discipline — it is a fixed three-action measurement script whose only reader is a haiku probe session) micro-tests=no (the fixture's behavior is asserted by the #1264 run.sh block against the verdict helper, which is the executable coverage that matters here)

Closes nothing on its own — it unblocks #1264.


The01Geek:
## Review reception — deferral record (no code change)

A reception pass triaged all 7 notes from the APPROVE-with-notes review at
`4e48a62b1c9ab238c2804d9c3afb0ee12c7f03fb`. **Outcome: no change; merge as-is.**

Every note was verified against the code. Six of seven are factually correct; none is
fixed, and the reason is recorded per note below so a later pass can downgrade rather
than re-litigate.

### The two facts that govern the whole triage

1. **The measurement is already taken.** The `placeholder-probe` job ran green on this
   exact head (run `31062226103`, job `92492378957`) and produced
   `VERDICT: SUBSTITUTED_ENV_VISIBLE`, `RECORD IN #1264 THREAD: yes`,
   `ROUTES TO: the placeholder mechanism (selected direction)`. The reason clause it
   emitted is the *correct* one — it states limb (c) came back NEGATIVE. So the probe's
   deliverable is in hand and correct.
2. **Any push re-fires all 16 paid probe jobs.** `matcher-probe.yml`'s `pull_request`
   `paths:` filter is evaluated against the whole PR diff, not the push delta — confirmed
   empirically: head `4e48a62b` is a `main` merge commit touching no probe file, and the
   probe workflow ran anyway. A push to correct behavior-inert prose therefore costs 16
   paid sessions plus a re-review, and changes no measurement.

### Per note

| # | Note | Verdict | Why deferred |
|---|---|---|---|
| Important 1 | No fixture exercises the single-JSON-document parse or nested `content` recursion | **Correct, no live defect** | `json.loads` on a real single-array file returns a list and `walk` is shape-generic, so the untested path works. Coverage gap only; fail-direction is a loud `INCONCLUSIVE`, never a silent wrong verdict. Review self-annotated it a suspected over-grade; agreed. |
| Sugg 1 | `SUBSTITUTED_ENV_UNSET` reason says "ungranted head — limbs (a) and (c) positive" | **Correct, unreachable arm** | Verified stale: the head *is* granted (`Bash(.github/probe-plugin/phprobe-read-env.sh:*)`) and limb (c) is settled NEGATIVE. But this string lives in an arm that did not fire and now fires only if limb (b) regresses upstream — limb (b) measured POSITIVE. Verdict token, `RECORD IN`, and `ROUTES TO` in that arm are all correct; only the explanatory clause is stale. |
| Sugg 2 | coverage-map note says "six" degraded arms, enumerates seven | **Correct, cosmetic** | Counted: 7 enumerated, 7 matching `INCONCLUSIVE` assertions in `run.sh`. Nothing reads the note prose (the coverage guard reads `owner`/`focused_test`), and `lib/test/` is pruned from the vendor slice. |
| Sugg 3 | "Always exits 0" docstring vs unguarded `main` | **Correct, loud + hard to reach** | Real contract imprecision. The reachable trigger is a `RecursionError` in `walk` on pathological nesting; the failure is a traceback and non-zero exit, i.e. fails closed and visibly, not a wrong verdict. |
| Sugg 4 | `unexecuted` arm precedes the reported/env arms, so contradictory evidence yields a confident `NOT_SUBSTITUTED` | **Correct, fail-open but gated by an explicit instruction** | The genuine risk in the set. Requires a SAW-scoped echo carrying *both* a valid sentinel and the raw placeholder text — the skill instructs a verbatim single-line echo, and the substituted output names neither the script path nor the backtick-bang form. Already-taken run confirms the intended shape. Revisit if a future probe run returns `NOT_SUBSTITUTED` while the transcript shows a sentinel present. |
| Sugg 5 | `$GITHUB_ENV` attribution in the `SUBSTITUTED_ENV_VISIBLE` reason and the limb-(b) YAML comment | **Correct as to naming; the inference it supports is sound** | The sentinel does come from the step-level `env:` mapping, not the `$GITHUB_ENV` file. But `devflow.yml` and `devflow-runner.yml` write `$GITHUB_ENV` precisely so the value "arrives through the environment" of the Claude step — both channels present to the injected render-time command as the same inherited process env var. So the conclusion "render-time injection can inherit the #874 closure" is supported by what was measured; only the channel label is loose, and the YAML's own sentinel comment already hedges it. |
| Sugg 6 | Minor test-completeness gaps (`_OBSERVED` filter, `collect_results`, input-less `tool_use`, `OSError` branch) | **Correct, low value** | All resolve into already-covered `INCONCLUSIVE` paths. |

### Revisit condition

Re-open notes 1, 3, 4 and 5 together — in one push — if and only if `matcher-probe.yml`
is being edited for another reason, or a future probe dispatch returns anything other
than `SUBSTITUTED_ENV_VISIBLE`. Fixing them on their own does not change a verdict, a
routing decision, or any recorded measurement, and would cost 16 paid probe sessions.

<sub>Recorded by a reception pass because this PR closes no issue, so no workpad exists.</sub>

PR #1353, opened 2026-08-06
fix: deliver the consumer prompt extension by render-time injection (#1264)

### LINKED ISSUE #1264: The consumer prompt extension reached the agent in only 8 of 18 review runs and 1 of 4 implement runs, so repo policy is applied non-deterministically

The01Geek:
## Summary
- The consumer prompt extension now reaches the three engine skills by **render-time injection** — a `` !`…` `` placeholder Claude Code executes before the model sees the skill — instead of a loader command the agent could simply not run. Measured before this change: the extension reached the agent in **8 of 18** sampled `/prflow:review` runs and **1 of 4** sampled `/prflow:implement` runs, and both failure modes were silent.
- The silence is the actual defect. A review that never loaded the policy still posted a normal APPROVE/REJECT verdict, and nothing in the verdict, the workflow, or CI distinguished it from one that had — two PRs reviewed three minutes apart got opposite treatment on the same gate for this reason.
- The new wrapper **always exits 0** and always prints one status line, because a non-zero exit from an injected command aborts the whole skill invocation at zero turns. Wiring the existing reader in naively would have traded a benign no-op for a silent no-verdict run at the merge gate.

## Changes

**New helper**: `scripts/render-prompt-extension.sh` wraps `load-prompt-extension.sh`, absorbing its exit 2 and emitting exactly one `PROMPT-EXTENSION-STATUS:` line — `content-present` (then the extension's bytes), `present-empty`, or `unestablished (<reason>)`. `unestablished` is never collapsed onto `present-empty`; it additionally covers an absent trusted closure, which the reader alone would report as an ordinary absent file. `DEVFLOW_PROMPT_EXTENSION_ROOT` is read inside the wrapper's body, never in the placeholder, because an expansion in placeholder text is refused with `Contains expansion`.

**Engine skills**: `skills/review/SKILL.md`, `skills/review-and-fix/SKILL.md` (both blocks — its own extension and `receiving-code-review`) and `skills/implement/SKILL.md` carry the placeholder and route on the status token. Their existing loader prose is demoted to an explicit fallback for runners without render-time preprocessing (Copilot CLI, Cursor, Codex CLI, Gemini CLI); the portable anchor line and its exit-code prose are preserved verbatim, since the suite pins both.

**Permissions**: `Bash(*/render-prompt-extension.sh:*)` plus the vendored literal are granted on the `review`, `implement` and `command` profiles. The wildcard is the load-bearing one — the placeholder's `${CLAUDE_SKILL_DIR}` resolves to an absolute path no vendored literal matches. This widens the read-only reviewer, so `lib/review-profile.tokens` moved in the same change and the workflow literals were regenerated, never hand-edited.

**Tests**: `lib/test/modules/prompt-extension-reader.sh` gains 24 assertions covering all four input shapes, both exit-2 families, the absent-closure arm, the single-line collapse of the reason, and — because an inline placeholder is invisible to the head/closure scanners that guard fenced call sites — the wildcard grant in each profile, the review-profile lock, and each skill rendering **its own** extension name. Floor moved 127 → 151 at both coupled sites.

**Docs**: the render-time mechanism, its scope limits, and two disclosed residuals are recorded in `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md`, `docs/internal/cloud-allowlist.md` and `docs/internal/cloud-setup.md`, with consumer-facing troubleshooting and a release note.

## Resolves
Resolves #1264

## Test Plan
- [ ] `lib/test/run-module.sh prompt-extension-reader` reports 151 passed, 0 failed.
- [ ] Breaking the wrapper's always-exit-0 contract (changing the `unestablished` arm to `exit 2`) turns exactly the four exit-status assertions RED, and restoring it turns them green.
- [ ] `python3 lib/generate-capability-profiles.py --check` exits 0, and the review-profile lock carries the new wildcard token.
- [ ] `lib/test/lint-anchor-fallback-arm.py` exits 0 with `skills/implement/SKILL.md` newly enrolled.
- [ ] `grep -Fxq` finds the exact portable anchor line, and `grep -qF` the phrase `a consumer extension exists but could not be loaded`, in all three edited skills.
- [ ] The full suite passes with `0 failed` and `0 skipped`.

## Post-Merge Verification
- [ ] A `/prflow:review` run on this repository reaches the reviewing agent with the resolved `.prflow/prompt-extensions/review.md` text present and the status line reading content-present, **without** the agent invoking `load-prompt-extension.sh`.
- [ ] A `/prflow:implement` run on this repository likewise reaches the agent with the resolved `.prflow/prompt-extensions/implement.md` text present and the status line reading content-present, **without** the agent invoking `load-prompt-extension.sh`. Verified against that run's execution-transcript artifact by the procedure in the issue's *Verification* section.

## Reviewer Notes

**Two residuals are disclosed rather than closed, and the second is the sharper one.** The dispatched precondition probe used a *bare literal* path, so no run has exercised an anchor-bearing placeholder. (1) That `${CLAUDE_SKILL_DIR}` is substituted inside placeholder text is inferred from its documented substitution in skill markdown, not measured. (2) The constraint the probe established is titled *no shell expansion in the command text* — and the shipped placeholder carries `${CLAUDE_SKILL_DIR}`, which is syntactically an expansion. The design assumes Claude Code substitutes the anchor *before* the command is analyzed, so the analyzer never sees `${…}`; that distinction is assumed, not measured. If it is wrong, every placeholder is refused and the run silently degrades to the fallback prose — i.e. today's behavior, not something worse, since a refusal is not the zero-turn abort a non-zero *exit* causes. Both residuals are recorded in `docs/internal/cloud-allowlist.md`, which also names the `matcher-probe.yml` arm that would retire the second one ahead of the post-merge checks above.

**Why a wrapper rather than a flag on the existing reader.** A `--render` mode inside `load-prompt-extension.sh` would be the deeper fix and would remove a file, a process, and several grant tokens. AC2 of the issue explicitly mandates that a tracked wrapper be the injected command, so that alternative was recorded rather than taken unilaterally.

**Phase 3.3 in-run review coverage is not verified.** A faithful run of the full review engine did not fit the implementing run's remaining context, so the first-party `prflow:code-reviewer` agent was dispatched against the real diff instead; it raised three Important findings, all three accepted and fixed in `fix: address review findings (iteration 1)`. No checklist generation, no verification pass, and no shadow pass ran. The cloud `/prflow:review` that fires on CI green is the merge gate.

## Visual Changes
N/A

## Breaking Changes
None. Consumers on runners without render-time preprocessing keep the existing loader path, unchanged, as an explicit fallback; a consumer with no extension file still sees a no-op.


The01Geek:
Confirming the Important (informational) finding, as requested.

**The AC19 deferral is the intended handling** — not a baseline bump in this PR. Registering `render-prompt-extension.sh` here would defeat the mechanism's own design: a newly-granted head is registered only once `LEGACY_PROFILE_BASELINE` advances past the release that first shipped it, and pairing-2 goes RED only for a *required* head, which this deliberately is not yet.

The residual you named is real and accepted: a one-release window in which the wrapper's stdout becomes the merge-gating reviewer's prompt while its bytes are not in the runtime hash manifest. It is systemic to the deferral mechanism rather than specific to this change.

**Tracked so the registration is not lost at the next baseline advance:** #1359.

Independent verification performed before merge, in addition to this review: `generate-capability-profiles.py --check` exits 0 against the head; `lib/review-profile.tokens` moved in the same change so the read-only reviewer lock did not widen beyond the manifest; and the new locate-failure assertion was mutation-checked (neutering the guard turns exactly that assertion RED, 156 passed → 155 passed / 1 failed), confirming it discriminates rather than passing vacuously.

Merging on that basis.

Issue #582, opened 2026-07-18
Make the cloud-tier workflow runner configurable via a `DEVFLOW_RUNNER` variable (enable self-hosted runners)


gaborp:
## Problem Statement
DevFlow's cloud tier hardcodes `runs-on: ubuntu-latest` in every workflow job. Adopters whose CI must run on **self-hosted runners** — concretely, a .NET Framework shop that can only build on a Windows self-hosted runner labelled e.g. `[self-hosted, windows, DevFlow]` — cannot use the cloud tier at all. Their only current workaround is to fork the vendored workflow files and rewrite each `runs-on:`, which collides with every `devflow_version` re-vendor and is unmaintainable. There is no supported knob for runner selection today.

## Current Behavior
Every job in the consumer-shipped workflows pins `runs-on: ubuntu-latest`, and every `run:` step relies on the implicit GitHub-hosted-Linux default shell (bash). No `.devflow/config.json` key or GitHub variable controls the runner, and no workflow sets an explicit shell. Verified `runs-on: ubuntu-latest` sites in the consumer-shipped workflows (12 total):

- `Verified:` `grep -n 'runs-on:' .github/workflows/{devflow,devflow-implement,devflow-review,devflow-runner,telemetry-push}.yml` → devflow.yml (4), devflow-implement.yml (3), devflow-review.yml (3), devflow-runner.yml (1), telemetry-push.yml (1).

## Desired Behavior
A single GitHub **repository/organization variable, `DEVFLOW_RUNNER`, selects the runner for every consumer-shipped workflow job.** When it is unset or empty, every job runs on `ubuntu-latest` exactly as today (existing Linux adopters set nothing and see no change). When it is set to a **bare single label** (`windows-latest`), jobs run on that runner; for multiple labels it accepts a **JSON array** (`["self-hosted","windows","DevFlow"]`). In addition, each consumer-shipped workflow declares `defaults: run: shell: bash`, so `run:` steps execute under bash on a non-Linux runner (the workflows' `run:` bodies are bash, and a self-hosted Windows runner otherwise defaults to PowerShell/cmd). Runner selection lives in GitHub Settings (Actions → Variables), not in the versioned `config.json`: it is infrastructure (which machine runs the job), kept separate from `config.json`, which governs how DevFlow behaves.

[...]

Issue #1309, opened 2026-08-05
lint-shipped-pruned-path flags the consumer's own configured doc roots, so shipped skill prose must carry suppression markers to name a documented `docs.*` default


The01Geek:
## Problem Statement

`lib/test/lint-shipped-pruned-path.py` exists (issue #1072) to stop a shipped skill body from
naming a path that will not exist in a consumer's checkout, because the vendor slice deletes it
from the vendored plugin. It derives its forbidden-path set from the slice's own `rm` line
rather than a hardcoded literal, and fails the suite on any `skills/**` / `agents/**` reference
to one of those paths that carries no `<!-- pruned-path-ok: <reason> -->` declaration marker.

Two of the paths that slice prunes — `docs/external` and `docs/internal` — are simultaneously
the **documented default values of consumer-facing config keys**
(`.docs.external`, `.docs.internal`). The pruned thing and the configured thing are different
directories that happen to share a spelling: `.prflow/vendor/prflow/docs/external/` is deleted
from the vendored plugin, while `<consumer repo root>/docs/external/` is exactly where the
consumer's public docs are supposed to live and is a path the docs skills are supposed to teach
consumers about.

The lint cannot tell them apart. Its match is bare substring containment, so every shipped
sentence that names a documented `docs.*` default — a configuration preamble, an illustrative
directory layout, a worked example — is a finding, and the only available discharge is to write
a PRFlow-internal lint declaration into consumer-facing prose.

This is already live on `main`, not a future risk: fifteen `pruned-path-ok` marker lines under
`skills/**` today suppress this false-positive class, and the sites are enumerated in Current
Behavior below.

## Current Behavior

`lib/test/lint-shipped-pruned-path.py --print-prune-set` on `main` (`5a589cc78`) emits:

```
.claude-plugin/marketplace.json
docs/external
docs/internal
docs/site
lib/test
```

`scan_text()` then tests each line with plain Python substring containment, through the shared
`_scan` primitive the citation check below also uses:

```python
def scan_text(text: str, targets: list[str]) -> list[tuple[int, str]]:
    """Return (1-based line number, matched target) for each unmarked pruned-path reference."""
    return _scan(text, lambda line: next((t for t in targets if t in line), None))
```

There is no word boundary, no path shaping, no anchoring, and no exemption mechanism besides
the per-line marker. So `docs/external` matches inside `docs/external/release-notes.md`, and
`docs/internal` matches inside `docs/internal/orders/`.

[...]

PR #1418, opened 2026-08-08
Remove the never-shipped matcher-probe pointers from the shipped skill surface, and lint the class

### LINKED ISSUE #1402: Shipped skill bodies point at the repo-internal matcher probe, and no lint can see it
## Problem Statement

`CLAUDE.md` requires that a skill or phase file carry "a pointer only to an artifact that ships with the plugin."

Eleven lines across six `skills/**` files point at `.github/workflows/matcher-probe.yml`, five of them as "the evidence of record". That workflow reaches no consumer checkout, so all eleven violate the rule.

The cause is narrow. A consumer *does* have `.github/workflows/` with DevFlow workflows in it — `install.sh` copies `devflow.yml` and `devflow-implement.yml` there. But `vendor-slice.sh`'s `devflow_copy_slice()` copies no `.github/` at all, and `install.sh` names `matcher-probe` nowhere. The path family resolves; this member does not. A blanket `.github/` ban would therefore be wrong.

`lib/test/lint-shipped-pruned-path.py` exists for this defect class and is **green on the tree carrying all eleven pointers**: it derives its forbidden set from paths the slice *prunes*, and a never-*copied* file is not a prune target.

## Current Behavior

Eleven lines, twelve occurrences, six files, at `origin/main` = `48915361b`:

| File | Lines |
|---|---|
| `skills/review/SKILL.md` | 5 |
| `skills/review/phases/phase-0-setup.md` | 2 |
| `skills/review/phases/phase-4-verdict.md` | 1 |
| `skills/review-and-fix/SKILL.md` | 1 |
| `skills/implement/SKILL.md` | 1 |
| `skills/implement/references/deferred-ac-followups.md` | 1 |

**Three spellings, splitting 8 / 3 / 1.** Eight occurrences are the full `.github/workflows/matcher-probe.yml`; three are the bare basename `matcher-probe.yml`; one is the bare stem with no extension — `skills/review/SKILL.md:111` carries the phrase `matcher-probe row 9`. A path-keyed guard catches 8 of 12; a basename-keyed guard catches 11 of 12. The stem is not matchable without banning `ci` too.

Two of the eleven are trailing `#` comments inside fenced bash blocks the engine emits verbatim into a consumer's shell.

Nothing instructs a read of the probe file — every pointer is parenthetical provenance and the operative shape lists are inline — so no consumer run breaks. The harm is a false affordance ("re-runnable after any action/CLI upgrade", told to a reader with no such workflow), caveats whose evidence can never be resolved, and context spent on row ordinals naming absent machinery.

## Desired Behavior

No line under `skills/**` or `agents/**` points at a `.github/workflows/` file that reaches no consumer, enforced by a new forbidden class in `lib/test/lint-shipped-pruned-path.py`.

The class's set is the tracked `.github/workflows/*.yml` basenames absent from the union of two parsed `install.sh` declarations: the copy loop's operand list (`install.sh:1507`) and `DEVFLOW_WITHHELD_TIER` (`install.sh:799`). On today's tree that selects `ci`, `matcher-probe`, `version-consolidate`, `agents-seam-probe`, and excludes `devflow-runner` and `telemetry-push` by construction.

The probe's evidence-of-record status stays where it already lives: `CLAUDE.md` states it, and `CLAUDE.md` never ships.

## User Impact

Consumer maintainers are told to consult and re-run a workflow their repository does not contain, and consumer agents spend context on ordinals resolving to nothing. This class has produced seven issues in five days (#1072, #1190, #1204, #1241, #1309, #1310, #1329); the guards built so far all leave never-shipped paths uncovered.

## Technical Context

*Measured at `origin/main` = `48915361b`; re-measure rather than trusting these.*

- `devflow_copy_slice()` copies `.claude-plugin agents docs lib scripts skills LICENSES` plus three `.prflow/` templates. No `.github` argument.
- `install.sh:1507` is `  for w in devflow devflow-implement; do`; `install.sh:799` is `DEVFLOW_WITHHELD_TIER="devflow-review devflow-runner telemetry-push"`.
- **That copy-loop line already has two parsers**, and this change adds a third. `lib/test/run.sh` derives `_582_SHIPPED` from it with bash builtins; `lib/test/modules/review-stall-backstop.sh`'s `loop_ships_probe801_term()` scans it to assert `matcher-probe` stays *out* of the loop, with a `sed`-mutated planted-defect control over the same line. That module carries an `exact` assertion floor of 459.
- All six files are SHA-256 pinned in `scripts/devflow-cloud-writer-contract.json`. Any byte change needs `python3 lib/test/cloud_writer_contract.py generate` in the same commit, or `lib/test/run-parallel.sh` refuses to launch.
- The lint already carries `--slice-source`, `--schema-source`, `--files-from` and two `--print-*-set` flags in a mutually exclusive group — the pattern a third derived set follows.
- Both green-path near-misses are live prose, not fixtures: `skills/create-issue/references/step-4-present-create.md` points at `devflow-implement.yml`, and `skills/implement/phases/phase-1-setup.md` at `devflow-runner.yml`.
- `lib/test/run.sh` already classifies the probe internal via `_582_INTERNAL="ci matcher-probe version-consolidate agents-seam-probe"`, pinned complete by a partition assertion — a transcription, which is why the lint derives from `install.sh` instead.
- No open issue owns this. #1072 built the lint and enumerated its out-of-scope set; never-shipped `.github/` paths appear in neither its scope nor its carve-outs.

## Acceptance Criteria

- [ ] `grep -rn 'matcher-probe' skills/ agents/` exits 1 over the post-change tree, and no `pruned-path-ok` marker naming the probe was added in any spelling — the eleven pointers are resolved by removal, not suppression. The exit status is the instrument; empty output alone also means a denied `grep`.
- [ ] Each rewritten paragraph keeps the instruction its pointer paragraph carried and at most one sentence naming what breaks without it, per the instruction-plus-consequence rule. Discharged by reviewer judgement; no automated instrument.
- [ ] `lib/test/lint-shipped-pruned-path.py` exits non-zero, naming file, line and matched basename, on an unmarked reference to a never-shipped basename in each of these two spellings independently: `.github/workflows/matcher-probe.yml` and bare `matcher-probe.yml`.
- [ ] The lint exits 0 on exactly these four near-misses, complete by construction: a line citing `.github/workflows/devflow-implement.yml`; one citing `.github/workflows/devflow-runner.yml`; one citing `matcher-probe.yml` with an HTML `pruned-path-ok` marker; and one inside a fence citing `matcher-probe.yml` with a shell `# pruned-path-ok:` marker.
- [ ] The forbidden set is computed at run time by **word-list membership** over the two parsed `install.sh` declarations — exactly these two, complete by construction — with no basename literal transcribed into the lint, and is printable via a new flag joining the existing mutually exclusive print group.
- [ ] When a parsed `install.sh` declaration cannot be established, the lint exits non-zero naming `install.sh` and audits no file, so an unestablished set is never silently an empty one.
- [ ] `python3 lib/test/cloud_writer_contract.py verify` exits 0, and `CLAUDE.md`'s description of the lint names the new forbidden class alongside the two it already documents.
- [ ] `lib/test/run-parallel.sh` completes with a clean aggregate and a zero exit status.

## Implementation Notes

**Approach.** One PR, one commit: the prose sweep, the lint class, and the regenerated writer contract. The lint is RED against the tree as it stands, so any commit landing it ahead of the sweep leaves a red required check; and the writer contract is stale from the first prose byte changed.

Each pointer is deleted because the rule says a pointer must name a shipped artifact. Where the sentence still needs its point, restate the operative clause inline — which is also where most of this probe-epistemics prose exceeds the one-consequence budget, so both halves of the rule are discharged by the same rewrite.

**Relevant files** — at minimum:

- The six files above — the eleven pointer lines.
- `lib/test/lint-shipped-pruned-path.py` — the new derived set, print flag, fail-closed refusal, docstring residuals.
- `lib/test/run.sh` — fixture-driven red, green and fail-closed paths beside the existing `#1072` and `#1241` blocks.
- `scripts/devflow-cloud-writer-contract.json` — regenerated, not hand-edited.
- `CLAUDE.md` — the lint's description.

**Gotchas.**

- Membership must be over parsed word lists, never a substring search: `ci` appears as a substring on hundreds of unrelated `install.sh` lines, and `devflow-review` appears in text that *removes* it.
- Three `lib/test/run.sh` pins quote section titles that sit on pointer lines — `Cloud command-shape discipline (implement tier)`, a `==4` count of `Cloud-emission discipline (label helpers)`, and a `==1` count of a review-bundle `cat`-heredoc clause. The edits are intra-line; those titles must survive byte-identically.
- Two disclosed residuals for the lint's docstring: the extensionless `matcher-probe` stem stays invisible to a basename key, and a pointer to a workflow deleted from the tree (`devflow-review.yml`) is invisible to a tracked-listing intersection.
- `skills/implement/phases/phase-1-setup.md`'s `devflow-runner.yml` block is out of scope and must not be touched — it already guards the absent case with `[ -f "$VENDORED" ]`. The derived set excludes it by construction, as it does the `telemetry-push.yml` mention in `phase-3-review.md`.

## Sequencing — do not run concurrently with #1401

Both this ticket and #1401 modify `lib/test/lint-shipped-pruned-path.py`. Two open PRs in one file is
the conflict this repo's convention forbids, so these must land one at a time. Whichever is picked up
second rebases onto the first and re-derives its own acceptance criteria against the updated lint: the
two forbidden-set derivations are independent and compose, but the surrounding scan structure and
`CLAUDE.md`'s description of the lint will both have moved.

## Notes from the phase-1-setup pass (#1400 / PR #1403)

- **Build a pin tripwire before the first edit.** A ~40-line scratch script that extracts every string literal ≥10 chars from `lib/test/run.sh`, `lib/test/modules/*.sh` and `lib/test/*.py`, counts each one's occurrences in the target file, and snapshots the result; re-count and diff after every edit round. It turns "did I break a pin?" from a full suite run into a two-second check.
- **All defects came from the reviewers, none from the machinery.** That pass found 23; pins, lints and the full suite caught 0 of them. A green suite is a regression guard, not evidence the compression is correct — budget review rounds accordingly.
- **Fixes introduce defects at about the same rate as the original pass.** Two of five second-iteration changes were fixing first-iteration fixes. Read every `+` line of a fix adversarially against the acceptance criteria, not just "did the fix work".
- **If this file contains a coupled mirror, check the other half before editing.** A prior pass compressed one half of a `§1.6 ↔ §2.1` mirror and left the other while both still claimed to be verbatim twins. If the other half has already drifted, reconciling it is in scope — adopt the already-compressed half verbatim and edit only your own file, which keeps the "no other file modified" criterion true.
- **Calibration:** the rule yielded **−14.5%** on a dense phase file with every instruction, prohibition, degraded arm, failure token, command form and closed-set enumeration preserved. A much larger number is a signal to re-check for over-deletion, not a win.
- **`lib/test/regenerate-artifacts.py` with no arguments is slow (~8–20 min), not hung — do not kill it.** Row 5 (`exact-module-floors`) runs the real focused module runner serially once per exact-policy module; output is buffered to a single flush at the end by design. Use `--preflight` for the five fast rows and run row 5 separately. Open issue #1126 owns bounding it.

The01Geek:
## Summary
- Eleven lines across six shipped skill files pointed at `.github/workflows/matcher-probe.yml`. Every `skills/**` body is copied verbatim into a consumer repo and the vendor slice copies no `.github/` at all, so those lines told a consumer to consult and re-run a workflow their repository does not contain. Each paragraph was rewritten to keep its instruction and drop the provenance ledger.
- A fourth forbidden class in `lib/test/lint-shipped-pruned-path.py` stops the family coming back. It derives the never-shipped set at run time from `install.sh` rather than transcribing basenames, so a workflow the installer starts shipping leaves the set with no edit to the lint.
- The class fails closed throughout: an unestablished — or ambiguous — declaration exits non-zero naming `install.sh` and audits nothing, because an empty forbidden set would audit the shipped surface against nothing and report clean.

## Changes

**Shipped skill prose**: Removed every pointer to the matcher probe from `skills/review/SKILL.md`, `skills/review/phases/phase-0-setup.md`, `skills/review/phases/phase-4-verdict.md`, `skills/review-and-fix/SKILL.md`, `skills/implement/SKILL.md` and `skills/implement/references/deferred-ac-followups.md`. Every permitted shape, denied shape, degraded arm and two-denials rule survives; what went is the row-and-verdict provenance, replaced where needed by a statement of the shape's epistemic status.

**The lint**: `parse_shipped_workflow_names` reads the workflow copy loop's literal operand list and `DEVFLOW_WITHHELD_TIER` by parsed word-list membership, never a substring search — `ci` occurs on dozens of unrelated `install.sh` lines. Neither declaration is read by first match: a second literal-bearing workflow loop, or a second withheld-tier assignment, refuses naming both line numbers, because the selection conjuncts say nothing about a loop's direction and a removal loop over literal workflow names would invert the set silently. `derive_never_shipped_basenames` subtracts that union from an index-read `git ls-files` listing, so an untracked scratch workflow cannot move the set locally on a tree CI reports green. The scan is filename-boundary-aware, so a consumer's own `docs-ci.yml` is not reported as PRFlow's `ci.yml`.

**Lint CLI**: `--install-source` and `--workflows-source` override the two new derivation sources; `--print-never-shipped-set` joins the existing mutually exclusive print group and exits before the slice and schema reads, so a query about the workflow set is never diagnosed against a source it does not consult.

**Test suite**: A driver block in `lib/test/run.sh` covering the derivation over synthetic installer and workflow-source pairs, every refusal arm attributed to its own diagnostic, both forbidden spellings independently, the four near-misses, the disclosed residuals, and a marker-blind live gate over the tracked shipped surface — so restoring a pointer behind a declaration marker turns the suite red while leaving the lint green. Fixtures under `lib/test/fixtures/shipped-pruned-path/{installs,workflows,skills}/`. A new assertion reconciles the derived set against the `#582` plugin-internal partition, which nothing previously cross-checked.

**`CLAUDE.md`**: The lint's description names the new class and points at the docstring for its residuals.

**`lib/test/mutation-pin-census.py`**: `_SOURCE_PARSE_CACHE_SIZE` raised 76 → 84. The new installer fixtures took the census's swept shell population from 71 to 79, past the old bound, so every memo entry was evicted before reuse and its own reuse contract went red — the remedy that assertion states, and the third such raise this constant records.

## Resolves
Resolves #1402

## Test Plan
- [ ] `grep -rn 'matcher-probe' skills/ agents/` exits 1, and the branch diff adds no `pruned-path-ok` marker naming the probe.
- [ ] `python3 lib/test/lint-shipped-pruned-path.py` exits 0 over the tree, and `--print-never-shipped-set` prints `agents-seam-probe ci matcher-probe version-consolidate`.
- [ ] The lint exits non-zero naming file, line and matched filename for both `.github/workflows/matcher-probe.yml` and bare `matcher-probe.yml`, and exits 0 on the four near-misses.
- [ ] Each unestablished or ambiguous `install.sh` declaration exits non-zero naming `install.sh` with no `audited N of M` line.
- [ ] `python3 lib/test/cloud_writer_contract.py verify` exits 0.
- [ ] `lib/test/run-parallel.sh` completes with a clean aggregate and a zero exit status.

## Visual Changes
N/A

## Breaking Changes
None. The lint is desk-time only and the rewritten prose preserves every operative instruction, so no consumer-visible behavior changes.


The01Geek:
## Reception pass on the APPROVE at `2fc17c44e` — deferral record

Ran the code-review reception discipline against the approving review at this head. **No code changed.** Each of the four advisory notes was verified against the code and deferred, with the reason recorded here so the next pass can downgrade rather than re-litigate.

Configured `.receiving_review.fix_severity_threshold` is `suggestion`, so all four notes are *eligible* to re-open the diff. They are deferred on cost, not on grade: any push discards a green APPROVE plus fully-green CI at this head and buys a fresh full CI run and a fresh paid cloud review, on a PR six other tickets are queued behind. All three files involved are outside `scripts/devflow-cloud-writer-contract.json`, so none of this needs artifact regeneration whenever it is picked up.

**Note 1 — `_WITHHELD_TIER` misses `export DEVFLOW_WITHHELD_TIER=`.** Verified: `^\s*DEVFLOW_WITHHELD_TIER=(.*)$` does not match an `export`-prefixed form, and the outcome is `NeverShippedParseError("no DEVFLOW_WITHHELD_TIER assignment found")` → both call sites exit 1 → suite RED naming `install.sh`. Maximally loud, no wrong output, no skipped guard — a Suggestion, below the reviewer's Medium. Deferred, with a scope correction: `_COPY_LOOP` recognizes its own narrow source shape and refuses on anything else in exactly the same way, so "tolerate a leading `export`" is one instance of a class, not a singleton. Loosening one of the two shape recognizers and not the other would make the parser asymmetric for no gain. **Revisit when:** `install.sh`'s declaration shape is actually refactored, or the next change to this parser touches both recognizers — fix them as a pair or not at all.

**Note 2 — the `and "/" not in name` nested-workflow guard has no covering test.** Verified: the fixture tree under `lib/test/fixtures/shipped-pruned-path/workflows/` is flat, so the guard is uncovered. Deferring, and grading it below the reviewer's Suggestion: GitHub Actions reads only the top level of `.github/workflows/`, so a file in a subdirectory there is not a workflow at all. Excluding it is *correct behavior*, not a defensive approximation of one, and the class has no live instance and cannot acquire one while that remains true. A fixture would pin a true statement about a definitionally-empty population. **Revisit when:** GitHub begins executing nested workflow files, or the derivation's population source widens beyond that directory's own children.

**Note 3 — census comment terminology.** The reviewer's remedy is right and its rationale is right; the framing understates where the confusion actually bites. Measured: tracked `lib/test/**/*.sh` is 72 at the base and 80 at this head. The non-UTF-8 fixture `lib/test/fixtures/ghapi-repo-path/adversarial-nonutf8.sh` takes a `continue` before ever reaching `_definition_scan`, so the memoized population — the thing `_SOURCE_PARSE_CACHE_SIZE` actually bounds — is 71 → 79. **The new paragraph's numbers are correct; it is the more precise of the two.** The real trap is that the paragraph immediately above it, the #1309 raise, ends at "72" using the *tracked* convention (tracked was indeed 72 at `5843fe490`, where this same fixture was already present). A reader going down the comment block hits 68 → 72, then 71 → 79, with nothing explaining the step. Reviewer's proposed half-clause — "minus the one non-UTF-8 fixture the sweep skips" — fixes it, and belongs on the **new** paragraph. **Revisit when:** the next change touches this comment block; state the convention rather than adjusting either number.

**Note 4 — disclosed scope boundaries.** Agreed, not defects. See below.

### On the two follow-ups this PR's own run filed

**#1423 — the withheld-tier treatment. Verified true, and correctly deferred; it does not block this merge.** `DEVFLOW_WITHHELD_TIER` members are folded into the shipped set, and the five references it names are real. Two things make "a fail-open in its own new lint" the wrong frame:

- A fail-open reports clean on the input class the guard declares it catches. `lib/test/run.sh`'s pre-existing `#582` partition is three-way — `_582_SHIPPED` (copy loop), `_582_RETAINED` (`devflow-runner telemetry-push`, present in consumers that adopted the tier, absent from a fresh install), `_582_INTERNAL`. This lint's forbidden set is asserted *equal to* `_582_INTERNAL`, and it reports every member of it. `RETAINED` is a third category the lint does not claim, not a hole in the one it does — and the new coupling assertion makes that collapse visible against the repo's own taxonomy rather than hiding it.
- All five references predate this PR. None of the three files is in its diff, and #1402's Implementation Notes ruled the `phase-1-setup.md` block explicitly out of scope. Folding #1423 in would flip the derived set, break the `_582_INTERNAL` coupling assertion, and rewrite three further shipped bodies — a larger change than this PR's own shipped-prose delta.

One residual worth carrying into #1423: `CLAUDE.md`'s new sentence introduces the residual list with a definite article ("The residuals — … — are enumerated in the lint's own docstring"), and the `RETAINED` collapse is not among the four. The sentence is literally true and the surface is behaviour-inert, so it is a Suggestion rather than a documented falsehood — but whichever way #1423 decides, that sentence should be re-read in the same change.

**#1422 — correctly deferred.** It is genuinely *further* than the disclosed extensionless-stem residual, not a restatement of it. That residual is about what a filename-keyed scan cannot see; #1422 is about citations that carry no filename at all — row ordinals, shape numbers, a run id. Same root cause, different and strictly larger surface, and no mechanical instrument reaches it in either case. Its own ticket is the right home.

### Bottom line

Nothing found here blocks the merge. AC1 re-verified at this head: `grep -rn 'matcher-probe' skills/ agents/` exits 1, with no suppression marker added.

Issue #1230, opened 2026-08-03
Close the elective path: shadow coverage: "not_verified" is never a budget choice (decision: NO)


The01Geek:
## The question

The fix loop ends with a **shadow pass**: the whole multi-agent review engine runs a second time, blinded to the loop's own findings, to check that the loop really converged instead of merely agreeing with itself. When the shadow cannot complete — a reviewer returns nothing, the roster falls short, the `Agent` tool is unavailable — the loop records `coverage: "not_verified"` and reports the run as not independently verified rather than as clean.

This issue asks for a decision on one question:

> May a run **choose** `coverage: "not_verified"` as a budget decision, when the shadow pass *could* have run?

---

## Decision (2026-08-09): **NO — `coverage: "not_verified"` is never elective.**

A run may not choose it. It remains what it is today: a **consequence** of a shortfall, produced only by the closed set of conditions `shadow-review.md`'s *Honest-degradation fail-safe* enumerates. A run that could dispatch the shadow dispatches it. A run that cannot afford it does not get to convert that into a recorded "choice" — it stops, escalates, or ships without a clean approve on a genuinely degraded record.

### Answering the Cost note's "net cost is zero in their workflow"

The Cost note justifies the shadow by saying it replaces the manual second review an experienced user already runs, making the marginal cost zero rather than doubled. That justification **survives** this decision and in fact decides it: an elected skip only saves anything if the user then does **not** run that second review — which is exactly the case where the audit was needed. The saving is real only in the scenario where the loss is also real. So the note's own reasoning is an argument against election, not a cost to be traded away.

### The evidence that decided it

The "for electing it" argument held that on a converged run the shadow "re-reviews an unchanged diff… a second opinion on a surface nothing has touched." That claim was tested directly and did not hold.

On the implement run for issue #1466 (PR #1468), the orchestrator elected `not_verified` under budget pressure and recorded it honestly — behaving exactly as the permissive option imagines, hiding nothing. The shadow was run afterwards, on that same converged, unchanged diff. It returned **12 findings with zero overlap** with the loop's own pass: four genuine fail-opens in the checker the PR had already shipped (an import guard that never fired for a missing file, a silently-dropped fence invocation, a `$`-parameterized operand waved through as a placeholder, and an exemption set with no dead-entry check), plus two false claims in an already-published release note. A follow-up delta review found two more. Those fixes are PR #1472.

Two conclusions follow:

[...]

PR #1486, opened 2026-08-09
Gate Status: Complete on a machine-readable review-coverage record

### LINKED ISSUE #1453: [devflow-retrospective] meta: deferred-verification-unverified-review-coverage-ships-complete — Gate Status: Complete on a machine-readable review-coverage record (shadow, roster, checklist)
## Problem Statement
A maintainer merging a DevFlow-produced PR expects the "Complete" status on its workpad to mean the review engine's own verification steps actually ran. In at least eight merged PRs, an implement or review-and-fix run truncated its own shadow pass, its full reviewer-agent roster, or its checklist generate/verify steps under context-budget pressure, then wrote Status: Complete and marked the PR ready anyway — with the shortfall recorded only as a self-reported workpad reflection nobody is required to read before merging.

## Current Behavior
`skills/implement/phases/phase-3-review.md` and the review-and-fix references (`shadow-review.md`, `loop-control.md`) instruct a run to complete the full Step 2.6 shadow pass, the full Phase 3 reviewer-agent roster, and the checklist generate/dedupe/verify steps before declaring the PR review-ready. When a run cannot afford one of these — most often because the orchestrator's context budget runs out mid-Phase-3 — it degrades on its own judgment and still writes Status: Complete, runs `gh pr ready`, and lets the workpad's Devflow Reflection bullets carry the only record of what was skipped.

**The degradation is not merely tolerated, it is routed.** `skills/implement/phases/phase-3-review.md` §3.3's loop-verdict-marker table maps `CLEAN-NOT-VERIFIED` — a clean approve-family verdict whose shadow was *not verified* — onto the clean-completion path, and states the intent outright: it "surfaces the gap without blocking". The "Safe direction — non-negotiable" paragraph immediately above is narrower than it reads: it forbids treating a **missing or malformed** marker as clean-and-full, but a **well-formed** `coverage=not-verified` marker is explicitly admitted. So the loop reports the shortfall correctly and the caller is simply not required to act on it.

**A terminal gate now exists, and review coverage is not one of its members.** Since this issue was filed, `scripts/workpad.py` gained `_terminal_complete_gate`, which structurally aborts a `--status Complete` write (no PATCH) on three members: every non-post-merge Acceptance Criteria row ticked; exactly one `completion-verification:` verification-flight marker whose record passes the implement-completion validator (#1087); and a row for every member of `_REQUIRED_ARTIFACTS` (#1348, initially the base-update checkpoint-4 record). Review coverage is absent from that set — `grep -ci shadow scripts/workpad.py` returns `0`.

## Desired Behavior
A run cannot write Status: Complete or mark a PR ready while its own recorded review-coverage state shows an incomplete shadow pass, an incomplete reviewer-agent roster, or a skipped checklist step, unless the same write also carries a structured disposition record naming the gap and the reason it is being carried forward. The enforcement is mechanical, not a compliance expectation left to the run's own prose.

**The coverage state must first be produced in a machine-readable form.** Today the coverage fact does reach the workpad — `phase-3-review.md` §3.3 records `shadow agreed, full coverage` vs `shadow agreement not verified` through a `workpad.py … --note` before ticking the `review-and-fix` gate — but only as **free-text prose inside a `## Progress` note row**, with no keyed marker, no dispatch-attempted operand, and no roster/checklist axis. The machine-readable operands (the shadow's `coverage`, `reviewers_dispatched` versus the expected roster, and `checklist_skipped`) live in the fix loop's `iter-<N>.json` under the gitignored `.prflow/tmp/review/<slug>/<run-id>/`, which the workpad never sees. A gate written against an absent or free-text comparand is the fail-open this issue exists to close, so the producer is in scope: Phase 3.3 already resolves the coverage fact from the loop-verdict marker, and must stamp it onto the workpad in a form the gate can read — a keyed `--checkpoint` marker being the established idiom.

## User Impact
A maintainer who sees Status: Complete on a workpad can trust that the review engine's own verification steps actually ran, or that a documented, structured exception explains why they did not — instead of having to open the workpad's reflections to discover the coverage gap that a later post-PR review, or a human, ultimately had to close.

- **Relevant Classes/Files** — `scripts/workpad.py` (the workpad status writer and the existing terminal gate), `skills/implement/phases/phase-3-review.md` (§3.3, which resolves the coverage fact and owns the Phase 3 exit), `skills/implement/phases/phase-4-documentation.md` (§4.3, which owns the `gh pr ready` publish and the finalize write), `skills/review-and-fix/references/shadow-review.md`, `references/loop-control.md` and `references/loop-exit.md` (the shadow-coverage, roster and verdict-render sources of truth), `scripts/loop-verdict-marker.py` (the machine-readable coverage token the loop already emits), `lib/cheap-gate.jq` (the retrospective routing gate).

- **Architecture Alignment** — the precedent is `scripts/workpad.py`'s **existing terminal gate**, not the `--rewrite-ac` abort. `_terminal_complete_gate` already reads recorded workpad state and raises a structural `_UpdateError` before any PATCH; `_completion_evidence_verdict` and `_required_artifact_verdict` are its two standalone verdict helpers, and `_REQUIRED_ARTIFACTS` is the declared member set a new required record joins. This issue adds a fourth member on that established shape rather than inventing an abort mechanism.

- **#1230 is DECIDED and has SHIPPED (PR #1474): `coverage: "not_verified"` is never elective.** The shipped paragraph in `skills/review-and-fix/references/shadow-review.md`, beside the *Honest-degradation fail-safe*, reads: a run under cost pressure that could dispatch the shadow dispatches it; a fan-out that **was dispatched** and fell short records `coverage: "not_verified"` and its **true** cause, **cost included** — the coverage record reports coverage, not blame; what is refused is a budget or elective cause offered **in place of a dispatch that was never attempted**. The matching `shadow-review.md` row in `skills/review-and-fix/SKILL.md`'s failure map carries the same prohibition. **This is narrower than "refuse any reason containing the word cost", and this issue's escape hatch must match it exactly (#1230 AC8) — see *The disposition predicate* below.**

- **Verified against `origin/main` (2026-08-09) — every path in this issue exists and every quoted behaviour still reads as described:** `_terminal_complete_gate`, `_completion_evidence_verdict`, `_required_artifact_verdict` and `_REQUIRED_ARTIFACTS` are all present in `scripts/workpad.py`; `lib/cheap-gate.jq` still keys friction on the reflection KIND ("FRICTION (every reflection kind EXCEPT the informational `note`)"); `phase-4-documentation.md` §4.3 still reads "**Publish decision — `implement_pr_state`.** Whether the run publishes the PR or leaves it the draft created in Phase 3.1 is a per-consumer config choice." with no review-completeness precondition; `scripts/loop-verdict-marker.py` still normalizes every phrase other than the exact `shadow agreed, full coverage` to `coverage=not-verified`.

### Non-viable and viable operands (empirical findings from the #1230 / PR #1474 run)

These were established by a blinded multi-agent shadow review during the #1230 implementation and re-verified against the tree. They are stated here so the implementer does not rediscover them at Phase 2.

- **`pending_dispatch` CANNOT serve as the "was a dispatch attempted?" operand.** Both `skills/review-and-fix/SKILL.md` (Step 3 item 1) and `references/loop-control.md` (*Schema field semantics*) mandate that it is **cleared** once the returned attempt is joined or dispositioned — explicitly including failure, timeout, exhausted-retry and **not-verified** outcomes — and retained only while unresolved. On any run that reached a disposition it is therefore absent whether or not a dispatch ever happened. A gate keyed on it either fails open (absent read as "no dispatch needed") or blocks every clean run. It is a **named non-viable operand**.
- **The dispatch-attempted fact must be produced explicitly.** The only durable record of it today is `reviewers_dispatched` on the `shadow` block inside the gitignored `iter-<N>.json`, which `workpad.py` cannot read. So a positive, run-stamped "shadow dispatch attempted" operand is part of the producer's job — it is not derivable from anything the gate can already see.
- **The absent-`shadow`-block state already has a *proceed* routing that this gate must change.** `skills/review-and-fix/references/loop-exit.md`, under **Coverage → Shadow agreement**, carries the arm: *"If the verdict is non-REJECT but no iteration has a `shadow` block: 'Shadow agreement NOT verified — the shadow result was not recorded (workpad write may have failed); the loop's tentative verdict stands but was not independently audited.'"* That still renders an **APPROVE-family** verdict line, which `scripts/loop-verdict-marker.py` normalizes to `coverage=not-verified`, which `skills/implement/phases/phase-3-review.md` §3.3's `CLEAN-NOT-VERIFIED` row routes to the **clean-completion path**, which ticks the `review-and-fix` gate and reaches §4.3's `Complete` write. **These are the surfaces the gate must actually modify.** A gate stated only in `shadow-review.md` would be inert — the #1230 review caught exactly that mistake twice, because `shadow-review.md` is a *gated* Step 2.6 reference a run that never dispatched may never read.

### Is this the completion-evidence mechanism extended, or a separate one?

**Assessed against the implementation: it is the same *gate*, a fourth *member*, and a NEW standalone verdict helper — it is NOT an extension of the completion-evidence flight mechanism.** Evidence:

- `scripts/workpad.py` already carries `--record-completion-evidence FLIGHT_KEY`, which validates a record under `.prflow/tmp/verification-flights/<KEY>.json` via `scripts/check-completion-evidence.py` and refuses a `Status: Complete` write without one ("refusing to finalize Status: Complete — no completion verification-flight marker present [missing-evidence]").
- That mechanism's subject is the run's **in-env test/verification flight identity** (`scripts/verification-flight.py`, `check-completion-evidence.py`'s eight-token ladder: `missing-evidence → stale-candidate → verification-not-pass → …`). Its operands are a flight record on disk plus a claim identity re-derived from the tree. **Review coverage is none of those things** — it is a fact resolved from the loop-verdict marker at Phase 3.3, with no flight record and no tree-identity currency semantics. Overloading the flight validator with it would couple a review-engine fact to a test-suite staleness check that would then also invalidate the coverage record on any post-evidence edit.
- The correct structural precedent is **`_required_artifact_verdict` / `_REQUIRED_ARTIFACTS`**: a declared member constant, a **pure read** over the post-mutation `## Progress` section resolving a keyed `<!-- prflow:checkpoint KEY -->` marker through `_marker_variants(_checkpoint_marker(key))` (so both the `prflow:` and superseded `devflow:` spellings count), a member satisfied by **any** of its `accept_keys`, and a structural `_UpdateError` naming the exact producing command as the remedy. `_REQUIRED_ARTIFACT_CHECKPOINT_KEYS` is **derived** from that set, so a new member also extends the Phase 1.3 resume strip by construction rather than by a hand-transcribed second list — check whether the new coverage member belongs in that derivation (it describes *this attempt*, so it very likely does).

- **Cross-layer Impact** — four layers: the producer in Phase 3.3, the gate in `workpad.py`, the publish precondition in Phase 4.3, and the retrospective routing in `lib/cheap-gate.jq`.

- **Ordering hazard — the two writers are a pair, and Phase 4.3 acts first.** `workpad.py` cannot refuse `gh pr ready`; that command runs in Phase 4.3, which **publishes before** the finalize `--status Complete` write. Gating only the workpad therefore produces a published PR beside a Blocked workpad — strictly worse than today's behavior. The publish precondition and the workpad gate ship together or not at all. **There is an in-file precedent to mirror:** §4.3's base-update checkpoint-4 arm already states "**refuse to run `gh pr ready` and refuse to flip `Status` to `Complete`**" on a non-clean token, and routes to `--status Blocked --reflection-kind blocked`. The coverage precondition takes that same shape and sits beside it.

### The disposition predicate (what is actually machine-checkable)

The escape hatch cannot be implemented as a reason-string blocklist, and an enum of "the closed degradation set" is not implementable either — `shadow-review.md`'s set deliberately ends in the load-bearing residual "`reviewers_dispatched` falls short of the expected roster **for any other reason**" (#1230 AC2 protects that residual), so it is not a closed vocabulary of strings.

The predicate that **is** checkable, and that matches the shipped #1230 text exactly, is the **dispatch-attempted** fact:

- shadow **dispatched** and fell short → `not_verified` is a genuine degradation; the disposition is accepted, and its reason may name **any** true cause, **cost included**;
- shadow **never dispatched** → per #1230 the run has no legal way to record `not_verified` and proceed: it does **not** write `Complete`; it stops at a non-terminal or `Blocked` status naming budget exhaustion. The disposition is **refused**.

A blocklist on the words cost/budget/context would **contradict** shipped `shadow-review.md` prose by refusing a legitimate dispatched-but-truncated record. Implement the positive dispatch-attempted operand instead.

## Acceptance Criteria

Each criterion is tagged **[executable]** (carries automated regression coverage) or **[prose]** (agent-executed prompt prose whose only reader is the runtime agent — per `CLAUDE.md`'s recorded #843/#876 decision it carries **no** automated coverage and **no** wording pin; the review pass is the compensating control). Do not add a wording pin to satisfy a `[prose]` criterion.

- [ ] **AC1 [prose + executable producer contract]** Phase 3.3 stamps the run's resolved review-coverage state onto the workpad in a machine-readable keyed form — derived from the loop-verdict marker it already reads (`CLEAN-FULL` / `CLEAN-NOT-VERIFIED` / the `NO-MARKER`/`MALFORMED` fallback arms) rather than from prose, and **replacing or supplementing today's free-text `--note`** — so a later reader can establish coverage without parsing free text. The record carries, at minimum: the coverage verdict, the **dispatch-attempted** fact, and the roster/checklist comparison **results**. An unresolvable coverage fact is recorded as **unestablished**, never as full. *Verified by:* the gate's own tests reading a planted marker (the prose half — the instruction to stamp it — is `[prose]`; the marker's shape and the gate's reading of it are `[executable]`).
- [ ] **AC2 [executable]** `scripts/workpad.py` refuses (structural `_UpdateError`, no PATCH performed) a `--status Complete` write when that recorded coverage state is anything other than complete, unless the same write supplies a structured disposition naming the specific gap and a stated reason. The check is added as a member of the existing terminal-gate mechanism, alongside the acceptance-criteria, `completion-verification:` and `_REQUIRED_ARTIFACTS` members, as a **standalone verdict helper** in the shape of `_required_artifact_verdict` (pure read over `## Progress`, no mutation on any path). An **absent** coverage record is treated as unestablished — never as complete. *Verified by:* `lib/test/test_python_scripts.py`, asserting the refusal and asserting no PATCH was attempted.
- [ ] **AC3 [executable]** The disposition is refused when the run **never dispatched the shadow**, and accepted when a dispatched fan-out fell short — the predicate is the recorded dispatch-attempted operand, **not** a reason-string blocklist. A disposition whose reason names cost on a **dispatched-and-fell-short** record is **accepted** (shipped `shadow-review.md`: "records `coverage: "not_verified"` and its **true** cause, cost included — the coverage record reports coverage, not blame"). A disposition offered over a **never-dispatched** record is refused regardless of its reason. This satisfies #1230 AC8; the two issues must not disagree, and a blocklist implementation would make them disagree. *Verified by:* paired negative/positive controls on both operand values.
- [ ] **AC4 [prose]** Phase 4.3 makes the `gh pr ready` publish conditional on the same recorded coverage state, in the shape of the existing checkpoint-4 refusal arm ("refuse to run `gh pr ready` and refuse to flip `Status` to `Complete`"), so an incomplete pass leaves the PR a draft (or publishes only under an accepted disposition) rather than publishing a PR whose workpad the gate will then refuse to finalize. This criterion and AC2 ship in the same change; neither alone is a coherent end state.
- [ ] **AC5 [prose]** The `CLEAN-NOT-VERIFIED` clean-completion routing in `skills/implement/phases/phase-3-review.md` §3.3 and the missing-`shadow`-block APPROVE-family arm in `skills/review-and-fix/references/loop-exit.md` (*Coverage → Shadow agreement*) are updated so the not-verified state no longer reaches a `Complete` write unconditionally. Naming the rule only in `shadow-review.md` does **not** satisfy this criterion — that reference is gated behind a Step 2.6 entry a never-dispatching run may never reach.
- [ ] **AC6 [executable]** The refusal is exercised by automated tests in `lib/test/test_python_scripts.py` that plant each incomplete-coverage shape (shadow coverage not verified, a reviewer roster short of the expected set, checklist generation skipped) plus the **absent-record** shape and the **never-dispatched** shape, asserting the write is refused when no acceptable disposition is present and succeeds when one is.
- [ ] **AC7 [executable]** A run whose Phase 3 review pass completed in full — full shadow pass, full reviewer roster, full checklist verify — writes Status: Complete through the same call path unaffected by the new check, proven by a positive-control test alongside the negative controls above.
- [ ] **AC8 [executable]** A disposition record is written with a reflection kind `lib/cheap-gate.jq` classifies as friction (any kind other than `note` — the six kinds are `blocked`, `deferred`, `dropped-failed`, `improvement`, `issue-accuracy`, `note`; `dropped-failed` is the established kind for a "needs human action" outcome, per §4.3's `publish_failed` arm), so an incomplete-coverage disposition always routes the PR to the retrospective. The implementing PR names the kind chosen, and a compliant full-coverage run is asserted **not** to trip the gate.
- [ ] **AC9 [executable]** The disposition cannot be satisfied by a generic placeholder: it names the specific gap, and a negative-control test asserts a boilerplate value is refused.
- [ ] **AC10 [executable]** The gate is scoped to `Complete` and does not make a legitimately `Blocked` or `Failed` terminal status unreachable, proven by a test writing each of those statuses over an incomplete-coverage record.

### Proposed additions (added during audit — new, not in the original filing)

- **AC10** above (the Blocked/Failed reachability control) was previously only a *Potential Gotcha*; it is promoted to a criterion because it is the one way this gate can make the #1230 exit — "stop at a non-terminal or `Blocked` status naming budget exhaustion" — unreachable, and it is cheaply testable.
- **AC3**'s reframing from a closed-set reason vocabulary to the dispatch-attempted predicate is a **correction**, not an addition; the original AC3's intent (the disposition must not become an election channel) is preserved.

## Implementation Notes
- **Approach** — four coupled parts. (1) Producer: have Phase 3.3 stamp, as a keyed `--checkpoint` marker, the coverage fact it already resolves from the loop-verdict marker plus the dispatch-attempted operand. (2) Gate: add a coverage member to `scripts/workpad.py`'s `_terminal_complete_gate`, following `_required_artifact_verdict` in shape, with an absent record treated as unestablished; consider whether it joins the `_REQUIRED_ARTIFACT_CHECKPOINT_KEYS` derivation for the Phase 1.3 resume strip. (3) Publish: add the matching precondition to Phase 4.3, ahead of `gh pr ready`, mirroring the checkpoint-4 refusal arm. (4) Routing: choose the disposition's reflection kind so `lib/cheap-gate.jq` classifies it as friction. Then re-point the prose in `phase-3-review.md`, `loop-exit.md`, `shadow-review.md` and `loop-control.md` that restates "don't ship incomplete coverage" at the mechanical gate, per the helper-cutover convention.
- **Relevant files** — `scripts/workpad.py`; `skills/implement/phases/phase-3-review.md`; `skills/implement/phases/phase-4-documentation.md`; `skills/review-and-fix/references/shadow-review.md`; `skills/review-and-fix/references/loop-control.md`; `skills/review-and-fix/references/loop-exit.md`; `scripts/loop-verdict-marker.py` (read); `scripts/check-completion-evidence.py` (read, for the verdict-helper idiom); `lib/cheap-gate.jq`; `lib/test/test_python_scripts.py`.
- **Code Patterns** — mirror `_required_artifact_verdict`'s shape: a declared member constant, a pure read over the post-mutation `## Progress` section, marker resolution through `_marker_variants(_checkpoint_marker(key))` so both rename spellings count, a member satisfiable by any of its `accept_keys`, and a structural abort naming the exact producing command as the remedy. Do not invent a second abort mechanism, do not extend the `--rewrite-ac` retag abort (a different call path), and do not overload `check-completion-evidence.py`'s flight validator (a different subject — see *Is this the completion-evidence mechanism extended* above).
- **Scope judgement — one PR, but ship it in a strict order.** The four parts are a single coherent change and are individually incoherent (a gate with no producer fails closed on every run; a producer with no gate changes nothing; a workpad gate with no publish precondition is strictly worse than today). One PR is right. Within it, build in the order producer → gate → publish precondition → routing, and land the tests with the gate rather than after it. If the PR must be split for size, the only defensible seam is **AC5's prose re-pointing** (`loop-exit.md` / §3.3 routing wording) as a follow-up — and even that is risky, since AC5 is what makes the gate non-inert.
- **Potential Gotchas**
  - A run whose diff genuinely needs no full reviewer roster (config-only or docs-only) must not be forced through it. The expected roster is computed per diff profile in `shadow-review.md`, but it is recorded in `iter-<N>.json`, not the workpad — so the producer must carry the *comparison result* (roster complete versus short) rather than expecting the gate to re-derive a roster it cannot see. Note that `shadow-review.md` also makes an *intentional* checklist skip (`small_diff` + `config_only`) explicitly **not** a coverage shortfall — the producer must not report it as one.
  - `scripts/loop-verdict-marker.py`'s route-token vocabulary is a machine-consumed contract read across a plugin-version boundary; changing the meaning of `CLEAN-NOT-VERIFIED` breaks a caller loaded from a different version. Add a producer field rather than re-purposing an existing token. (PR #1474 also added a drift check that goes RED if the helper's result vocabulary grows — expect it to fire if you extend the vocabulary.)
  - The disposition's reflection kind interacts with `CLAUDE.md`'s *Recording a whole-suite launch* rule, which mandates `note` for the per-launch evidence marker precisely because every other kind trips the retrospective gate. Choose a kind that fires on a real coverage gap without firing on compliant runs.
  - **Do not add a tier-refused escape for the coverage member.** `_REQUIRED_ARTIFACTS` has a `…-tier-refused` accept key (#1347) for checkpoint 4, and the shape is tempting here. #1230 decided there is no tier carve-out for shadow coverage: a run that cannot dispatch stops at `Blocked`, it does not complete under an alternate accept key. Adding one would reopen the election path through a different door.

### Out of scope — filed separately, do not fold in

The **completion-evidence** member of the same terminal gate has no arm for a tier whose operator has delegated the whole-suite obligation to CI. On such a tier a compliant run can never produce a flight key in-run (the engine's #405 rule forbids citing CI as an in-run verification channel), so it is forced to choose between a false `Complete` and a `Blocked` that misreads as broken work — this happened on the #1230 / PR #1474 run itself. **Assessment: this belongs in its own issue, not here.** It is a different gate member, a different producer, a different validator (`check-completion-evidence.py`'s eight-token ladder, which has no such arm today), and a different decision — whether operator delegation of verification is legitimate at all — which #1453 has no mandate to settle. It is recorded here only so the implementer recognises it if the same shape surfaces during the work.

---

> **Audit note (2026-08-09, third revision).** Revised three times on 2026-08-09: first after an audit against the tree; second to carry the #1230 decision (`not_verified` is never elective); third after PR #1474 (the #1230 implementation) landed its shipped text, and after a full re-verification of every factual claim against `origin/main`. Corrections in this revision: (a) the claim that "no workpad field records [coverage] today" was **imprecise** — §3.3 does record it, as a free-text `--note`; the gap is machine-readability, not absence; (b) AC3's "closed set of reason strings" predicate was **wrong** and would have contradicted shipped `shadow-review.md` text, which permits cost as a *true* cause on a dispatched-and-fell-short record — replaced with the dispatch-attempted predicate; (c) `loop-exit.md` was added as an in-scope surface, because its missing-`shadow`-block arm is one of the two routings that carry a not-verified state to `Complete`; (d) `pending_dispatch` is recorded as a named non-viable operand; (e) the completion-evidence mechanism is assessed as a sibling gate member rather than a mechanism to extend; (f) ACs are now tagged executable/prose per #843/#876. Corrections from the first revision stand: a mechanical Complete gate does exist (`_terminal_complete_gate`), and `workpad.py` cannot gate the PR-ready transition (Phase 4.3 owns it, and publishes first). The retrospective provenance below is unchanged and is a past-time record.

## 🔁 Retrospective provenance
- **Pattern:** `deferred-verification` · first seen 2026-06-24T20:43:47Z · last seen 2026-08-02T09:18:20Z · 19 occurrences · status: regressed
- **Evidence base:** this root cause was re-derived from 8 of 19 occurrence bundles (Stage B's bundle set is a most-recent-first subset bounded by `audit_bundle_cap`; the `occurrences[]` list on the pattern record is the authoritative full history). Full primary sources (PR/issue bodies, reviews, comments, workpad body, commits) were read for PR #1070, #854, and #697; PR #131, #515, #531, #580, and #607 are drawn from the pattern record's own per-occurrence summary and descriptor fields rather than a freshly re-read bundle, since they fell outside this run's 8-bundle allotment.
- **Motivating PRs:** #131, #515, #531, #580, #607, #697, #854, #1070
- **Root cause (re-derived from primary sources):** Across these eight PRs, an implement or review-and-fix run exhausted its orchestrator context budget mid-Phase-3 and, rather than stopping or escalating, degraded its own verification silently: #697's workpad states plainly that "the orchestrator context budget could not carry the iteration-2 engine re-pass," leaving nine applied fixes with no independent review pass, and that the Step 2.6 shadow was "never dispatched"; #854's PR body discloses that only one of six Phase 3 reviewer agents ran and only 11 of 27 checklist items were verified; #1070's workpad self-labels the outcome "APPROVE WITH CAVEAT, never a clean approve" after running a bounded two-agent audit instead of the full five-agent shadow fan-out; #531, #580, and #607 show the same shape at larger scale, with #607 explicitly shipping a PR whose Phase 3 gates were "budget-truncated" and whose prompt-surface edits were made in "DEGRADED INLINE mode" rather than through the mandated subagent dispatch. In every case Status: Complete was written and the PR marked ready regardless. This confirms the corpus's own repeated diagnosis rather than contradicting it — none of the eight retrospective summaries appear to hallucinate a coverage gap the workpads do not actually record — but it sharpens the finding: the gap is not that the rule "don't skip verification" is unwritten (`CLAUDE.md`, `phase-3-review.md`, and `shadow-review.md` all state versions of it already), it is that nothing mechanically stops a run from writing Complete anyway, so the rule is only as strong as an individual run's willingness to self-report a shortfall it is under budget pressure to paper over.
- **Counterfactual:** A gate that is too rigid could force every run through a full reviewer roster and full shadow pass regardless of diff size, turning a config-only one-line change into the same cost as a 60-file rewrite, or it could deadlock a genuinely budget-constrained run into never being able to ship anything. The proposal avoids both: it reads the coverage the diff profile actually requires (not a fixed count) so a small diff needing fewer reviewer agents is unaffected, and it does not forbid shipping under a truncated pass — it only forbids shipping *silently*, by requiring the same write to carry a structured, specific disposition naming the gap. A run that is genuinely budget-constrained can still ship; it just cannot ship a "Complete" status without also recording, in a machine-readable field, exactly what it did not verify.
- **This sub-pattern:** This finding is the dominant sub-pattern in the corpus (8 of 19 occurrences) and is distinct from the two sibling findings in this return: it is specifically about the review engine's own internal verification steps (shadow pass, reviewer-agent roster, checklist generate/verify) being truncated and unrecorded, as opposed to the test-suite-execution shortfall (the second finding) or the acceptance-criteria exemption bookkeeping gap (the third finding) — three different mechanisms that all discharge the "deferred verification" pattern's diagnosis in different parts of the pipeline.

The01Geek:
## Summary

Gates `Status: Complete` on a **machine-readable review-coverage record**, so a run can no longer write `Complete` and publish a PR over a review pass that silently fell short.

Before this change the coverage fact did reach the workpad — as free text in a `## Progress` note — with no keyed marker, no dispatch-attempted operand, and no roster/checklist axis. The machine-readable operands lived in the fix loop's gitignored `iter-<N>.json`, which `workpad.py` never sees. `phase-3-review.md` §3.3's `CLEAN-NOT-VERIFIED` row routed a not-verified shadow onto the clean-completion path and said so outright — "surfaces the gap without blocking".

Four coupled parts, in the order the issue prescribes:

1. **Producer** — `skills/implement/phases/phase-3-review.md` §3.3 stamps the coverage fact it already resolves from the loop-verdict marker as a keyed `## Progress` row: `workpad.py update <issue> --record-review-coverage <coverage> <dispatch> <roster> <checklist>`, on every Phase 3 exit.
2. **Gate** — `scripts/workpad.py` gained `_review_coverage_verdict`, a further member of the existing `_terminal_complete_gate` beside the acceptance-criteria, completion-evidence and required-artifact checks, in `_required_artifact_verdict`'s shape: a pure read over the post-mutation `## Progress`, raising a structural `_UpdateError` before any PATCH.
3. **Publish precondition** — `skills/implement/phases/phase-4-documentation.md` §4.3 refuses `gh pr ready` on the same condition, mirroring the existing checkpoint-4 refusal arm. `gh pr ready` runs *before* the finalize write, so gating only the workpad would publish a PR beside a `Blocked` workpad — strictly worse than the old behaviour.
4. **Routing** — each accepted disposition auto-files a **`dropped-failed`** reflection (the kind chosen for AC8), which `lib/cheap-gate.jq` counts as friction, so a run that carried a gap forward to `Complete` always reaches the weekly retrospective. A compliant full-coverage run files none.

## The record

Four axes, declared once in `_REVIEW_COVERAGE_AXIS_SPECS` with every view derived from it and the table's invariants **enforced at import** — the edit that would fail the gate open is adding `unestablished` to an axis's clean set, so the validator refuses it.

`unestablished` is a first-class member of every axis: an unresolvable fact is recorded as unknown, never collapsed onto a clean value. An absent, duplicated or malformed record is unestablished too — never complete.

`dispatch` distinguishes three states that a single "did the shadow run" boolean would conflate: `attempted` (dispatched), `never` (a shadow the run **owed** and did not dispatch — the #1230 abuse), and `not-applicable` (a pass that owed no shadow, e.g. a REJECT the loop routes straight to Loop Exit). Without `not-applicable` the severity-aware soft-proceed exit — whose contract is "do **NOT** block; the PR is review-ready" — could never reach `Complete`.

## The escape hatch

`--review-coverage-disposition <gap> "<reason>"`, one per recorded gap. Per **#1230 AC8** its **cause** predicate is the recorded dispatch-attempted fact, **not** a cost/budget word blocklist: a fan-out that was dispatched and fell short may state any true cause, cost included, while a disposition over a `never`/`unestablished` dispatch is refused regardless of its reason. The reason string is separately screened for being a generic placeholder, so a disposition names the specific gap.

Four bracketed refusal tokens make every arm attributable: `[review-coverage-unestablished]`, `[review-coverage-gap]`, `[review-coverage-undispatched]`, `[review-coverage-boilerplate]`.

## Forgery surface

The gate's readers locate their marker inside a `## Progress` bullet, so any free-text channel could otherwise write a record that passed no validation and filed no evidence. Three controls close it: `--checkpoint` refuses a key in a reserved namespace; the readers accept only the **trailing** marker of a canonical timestamped bullet; and the free-text screen sits at the `_append_progress_note` **chokepoint**, so a channel it was never hand-listed for — the `--record-classification` rationale, or one added later — is screened by construction.

## Verification

`lib/test/test_python_scripts.py` — **4256 passed, 0 failed**; `ruff` clean. The new `#1453` block restores the real verdict while the sibling gate members stay bypassed, so every refusal is attributable to this member and not a sibling. It covers each incomplete axis alone, the absent/duplicated/malformed/out-of-vocabulary record shapes, the paired dispatch controls (including the load-bearing "a cost-naming reason over a **dispatched** record is **accepted**"), the read-time reason re-check with a planted-row positive control, both strip helpers with two-gap cardinality, the producer's five refusal branches, the marker-forgery screens, the import-time table validation with a positive control, and `Blocked`/`Failed` reachability over an incomplete record.

## Review

`/simplify` (4 agents) → the four parallel axis tables collapsed into one derived member table. Phase 3 roster (4 agents) → 2 Critical, 5 Important, 8 Suggestion, all fixed. A blinded **shadow** re-dispatch of the same roster on the fixed tree → **2 new Critical** (a free-text marker forgery, and the REJECT/soft-proceed arms dead-ending at `Blocked`), both fixed. A type-design pass → a third Critical (the classification rationale bypassing the hand-listed guard), fixed at the chokepoint above.

**Shadow coverage: `not-verified`, roster short, dispatch attempted.** The fan-out was dispatched and returned 4 of the 6-agent expected roster — `prflow:requesting-code-review` is a skill, not a registered agent type on this runner, so the final-pass reviewer could not be dispatched as a subagent. Recorded honestly rather than reported as full; this PR is review-ready, not auto-merged.

Resolves #1453


The01Geek:
## Reception pass — finding dispositions (`/prflow:receiving-code-review`, head `bd6d16d57`)

Base merged (`origin/main` → branch; the only conflict was the generated
`scripts/devflow-cloud-writer-contract.json`, resolved by regeneration, not by hand).
Findings unioned across both review runs (`31327692995` @ `2373d872d`, `31330298466` @ `1ecc5be70`).
Configured `.receiving_review.fix_severity_threshold` is `suggestion`, so each Suggestion was
triaged rather than auto-recorded.

### Fixed

- **R2-1 — `--review-coverage-disposition` arity.** Accurate. The sibling
  `--record-review-coverage` already guards its own arity, with a comment naming the exact
  reason (the suite builds `args` directly, so a programmatic caller can pass a short list);
  that rationale applies verbatim to the disposition flag, which had no such guard. Added the
  guard plus a table row in the producer-validation block. Mutation-checked: with the guard
  disabled the run aborts at `for gap, reason in review_dispositions` with
  `ValueError: not enough values to unpack (expected 2, got 1)` — the exact defect the finding
  named. Restored; tree verified clean.
- **R2-3 — `[review-coverage-boilerplate]` enumeration.** Accurate.
  `docs/internal/implement-skill.md` named two of the four branches
  `_review_coverage_reason_rejection` raises the tag from. Now enumerates all four
  (placeholder, under-length, HTML-comment delimiter, duplicate-disposition) and carries the
  duplicate row's remedy.

### Already resolved before this pass (verified against HEAD, not assumed)

- **R1-Important-1 — `not-applicable` missing from the `roster`/`checklist` axes.** Resolved in
  `1ecc5be70`. All four axis specs now carry `not-applicable` in both `values` and `clean`,
  `phase-3-review.md` §3.3 instructs `not-applicable not-applicable not-applicable not-applicable`
  on both the REJECT branch and the severity-aware soft-proceed, and
  `_review_coverage_incoherence` refuses any *mixture* — so the unreachable-to-Complete path the
  finding described is closed, and a dispatched pass still cannot borrow the value to hide a
  short roster.
- **R1-1 (garbled duplicate-disposition message)** and **R1-2 (`implement-skill.md` overstating
  the reserved-marker screen)** — both already corrected in `1ecc5be70`.

### Declined, with reasoning

- **R2-2 — read-time reason regex truncates at an embedded `<!--`.** Declined. Write time
  refuses both `<!--` and `-->` outright, so the producer cannot emit such a row. The residual is
  a hand-planted or older-workpad row — and a hand-planter gains nothing by embedding a
  delimiter, since they could simply write a well-formed reason instead. The gap token still
  resolves from its own marker (a separate pattern), so the truncation costs a partially-rendered
  reason in a human-readable row, never a wrong gate verdict. No fail-open.
  *Revisit if* a producer path is ever added that can write an unscreened reason.
- **R2-4 — changeset "gained a fourth member beside the … members".** Declined as not a defect.
  "beside" states membership, which is what the sentence claims and what the reviewer confirms is
  accurate; it asserts no sequence. The run position is stated where it belongs, in
  `DEVFLOW_SYSTEM_OVERVIEW.md` ("After the completion-evidence and required-artifact members, and
  before the AC reconciliation"). Rewording a correct release note to pre-empt a misreading is
  churn.

### Deferred (test attribution / defense-in-depth only — no fail-open in any)

- **R2-5 / R1-3 — attribution gaps:** gate-member ordering only half-asserted (the
  completion-evidence and required-artifact members are globally no-op'd in-process, so their
  ordering against the coverage member is not RED-testable); the multi-`## Progress`-section
  `_rc_survivors` breadcrumb untested; a dispositions-only call against the no-op replay
  short-circuit untested; the HTML-comment-delimiter rejection not separately attributable from
  the length/placeholder branch. Each is message-attribution or defense-in-depth; the fail-closed
  controls they sit behind (import-time `unestablished`-in-`clean` refusal, per-axis
  unknown-is-not-zero, dispatch-operand-vs-wording pairing) are present and directional.
  *Revisit if* a future change makes any of those four branches independently reachable in a way
  a wrong branch could satisfy.
- **R2-6 — negative real-CLI subprocess path.** Already carried as a documented DEFERRED with a
  REVISIT trigger, and the reviewer judged that justification adequate. Re-affirmed unchanged:
  the residual is CLI plumbing, every refusal arm is covered in-process against the real
  function, and the un-bypassed gate is reached end-to-end by the `run.sh` positive fixture.

Focused-test selection: `scripts/workpad.py` → `lib/test/test_python_scripts.py`
(coverage-map `owner: unmodularized`), run green at 4342 passed / 0 failed after
`lib/test/regenerate-artifacts.py` (`batched-regeneration: run` — the cloud-writer manifest
regenerated and committed with the edits). Whole-suite aggregate is delegated to CI.

The01Geek:
### Class sweep on the arity fix — scope note

Grepping the same unguarded-unpack shape across `scripts/workpad.py` finds three further
instances, all **pre-existing and untouched by this PR**: `--rewrite-ac` (`for old, new in
args.rewrite_ac`), `--scope-decision-deferred` (`for pr, text in …`) and
`--scope-decision-rewritten` (`for pr, old, new in …`). Each is `nargs`-constrained at the CLI
exactly as the disposition flag is, so each carries the same programmatic-caller residual.

Not swept here deliberately: the finding was internal inconsistency *within this change* (a new
flag lacking the guard its new sibling declares and justifies), and widening the diff into three
unrelated older flags would take this PR outside issue #1453's scope. Recorded so the next pass
sees a decision rather than an oversight. *Revisit as* a standalone follow-up if the shape is
worth normalising across the file.

The01Geek:
## Reception pass — round 2 dispositions (review at `bd6d16d57`)

The review at the current head returned **APPROVE with notes, no Critical or Important findings
surviving across four reviewers**. The five new items are all Suggestions, and all are the
fresh-advisory-batch an exhaustive engine produces on an already-clean verdict. Per the reception
skill's anti-treadmill rule I bounded the advisory re-open to the round-1 set (already applied in
`bd6d16d57`) and am recording this batch rather than opening a further cycle — each additional
push costs a full CI run and a paid review, and the reviewer's own recommendation is to merge with
these addressed opportunistically.

Deferred, with revisit conditions:

- **Duplicate-gap comment misstates the reader.** Real inaccuracy: the guard's comment says "the
  reader keeps only the last" row, but `_review_coverage_dispositions` resolves a duplicate gap to
  the `_REVIEW_COVERAGE_DUPLICATE_DISPOSITION` sentinel, refused as `[review-coverage-boilerplate]`
  — strictly safer than the comment describes. Behavior-inert rationale comment (no tool parses
  it), capped at Suggestion by the engine's own Phase 4.1.5 rule. *Revisit* on the next edit that
  touches this guard — reword to the sentinel behavior then, so the correction rides a change that
  is already paying for a review.
- **`_REVIEW_COVERAGE_REASON_MIN_LEN = 20` transcribed into `phase-4-documentation.md`.** Correct
  today; the drift risk is real but unguarded. *Revisit* by moving to floor-agnostic phrasing (as
  phase-3 already uses) rather than adding a pin — a wording pin would fall foul of the repo's
  own prohibition on prose-presence assertions.
- **All-`not-applicable` files no retrospective friction.** Accurate and bounded: every clean
  record is agent-attested, and the coherence check already refuses *mixing* `not-applicable` with
  measured values, which is the abuse path that matters. Filing friction on a pass that genuinely
  owed no shadow would put noise into the retrospective on the one arm that skipped nothing.
  *Revisit* if retrospective sampling ever shows the soft-proceed arm being used to skip an owed
  shadow.
- **Reason screening is shallow (20-char floor + 7-member boilerplate set).** By design: the hard
  gate is the recorded `dispatch=attempted` fact, not the wording — issue #1230's honest-degradation
  decision, which deliberately rejected a word blocklist. A 20+ char platitude still costs the
  author a `dropped-failed` reflection into the retrospective. *Revisit* only if the retrospective
  shows platitude reasons in practice.
- **`action='append'` with a mutable `default=[]`.** Inert on the shipped one-invocation-per-process
  CLI. Shares the shape with the file's other append flags, so it is a file-wide normalisation,
  not a change for this PR — same scope reasoning as the arity-sweep note above.

Gates re-verified at `bd6d16d57`: zero Critical/Important; `lib + python tests` = success; the
newest `prflow:review-verdict` marker reads `head=bd6d16d57 verdict=APPROVE`; branch 0 behind
`main`; `mergeStateStatus` CLEAN.

PR #1487, opened 2026-08-09
Bound the fix loop's per-entry instruction cost: scope the always-resident re-read rule

### LINKED ISSUE #1470: Bound the fix loop's per-entry instruction cost: scope the always-resident re-read rule
## Problem Statement

Every entry into the shared review engine costs the `/prflow:review-and-fix` orchestrator **259,585 B (~65k tokens) of instruction text** before a single line of the diff under review is read. That is the engine root plus the six phase files a default-config entry reads: `skills/review/SKILL.md`, `phase-0-setup.md`, `phase-0-6-stale-prose-lint.md`, `phase-1-checklist.md`, `phase-2-verification.md`, `phase-3-agents.md` and `phase-4-verdict.md` — the stale-prose lint counted because `.prflow/config.schema.json` defaults `prflow_review.stale_prose.enabled` to `true`. A modal converging run makes three such entries (one fix iteration, one clean iteration, one shadow pass), and inside `/prflow:implement` that lands on top of 432,011 B of implement prose already consumed before Phase 3.3.

What the loop *retains* from an engine entry is small. Across the 85 `iter-*.json` records under `.prflow/tmp/review/` in this checkout, the whole state record has a median of 6,024 B and a maximum of 32,098 B — so the engine's instruction text is between 8x and 43x larger than everything the loop keeps from running it. (Every byte figure in this section was re-derived against `071af828` on 2026-08-08 by summing `os.path.getsize` over the named files; the retention figures come from the 85 `iter-*.json` records present in this checkout at that time, a transient scratch population that a fresh checkout will not reproduce.) This is not a diff-size problem; the engine already hands diffs and slices by path. It is an instruction-load problem, and it is why runs reach the fix loop with the window already spent and silently substitute a narrower Phase 3 than the engine specifies.

One term is worse than the rest and nobody has priced it. The fix loop's **Always-resident re-read rule** fires after *every* `Agent`/`Task`/`Skill`-tool return while executing a reference. Step 1 *is* the engine entry, and it dispatches Phase 1 generators, the 1.5 deduper, Phase 2 verifier batches, and the 5–7 Phase 3 reviewers — on the order of 10 to 40 dispatch returns inside one engine entry. On the reading that treats `loop-control.md` as the active reference throughout Step 1, the rule re-reads its 44,371 B after each return: **111k–444k tokens per engine entry**, against an entry whose own instruction cost is ~65k tokens. On the reading that treats the engine's own phase file as the active reference, the rule costs nothing extra. The rule's text does not settle which reading is right.

## Current Behavior

Verified: `skills/review-and-fix/SKILL.md` states the rule as "After **every** `Agent`/`Task`/`Skill`-tool return while executing a reference, and **before the next cross-reference routing action**, re-`Read` the active reference — identified by `current_step`/`current_substep` in the active `iter-<N>.json`, **not** by conversational memory — and resume the interrupted substep; never re-dispatch the agent/skill that just returned."

Verified: `skills/review-and-fix/references/loop-control.md` states "`current_step` is `\"loop-control\"` during config resolution and Steps 0.5–2, then the routed step id". The same paragraph closes "The root's re-read rule consumes these operands after every dispatch return."

Those two sentences are the whole ambiguity. Step 1 — the engine entry — sits inside the span where the durable operand reads `loop-control`, while the file the orchestrator is executing at that moment is an engine phase file. Which one "the active reference" denotes is left open, and the two answers differ by up to 444k tokens per entry.

## Desired Behavior

The always-resident re-read rule states its own firing scope: it fires on a return that resolves a dispatch the active reference itself issued and that precedes a cross-reference routing action, and dispatches issued by the engine's own phases during Step 1 and Step 2.6 are governed by the engine's phase-entry gate instead. Per-entry orchestrator cost stops depending on how one sentence is read.

The durable-operand contract is untouched: the re-read still fires wherever a routing decision is about to be made, still keys on the stamped operands rather than recall, and still fails closed when an operand is absent.

## User Impact

A `/prflow:implement` run that today arrives at Phase 3.3 with its window largely spent on instructions arrives with measurably more of it left for the diff under review, which is the input that decides what the review actually catches. Consumer repos get the same effect, since the fix-loop surface ships verbatim into their checkouts.

## Technical Context

- **Relevant Classes/Files** — `skills/review-and-fix/SKILL.md` carries the rule. `skills/review-and-fix/references/loop-control.md` carries the durable-operand paragraph the rule reads and the Step 1 engine-entry procedure the ambiguity lands on.
- **Documentation Drift** — the Step 1 assessment ran on its **degraded** bounded-inline arm, so the drift picture is unestablished rather than clean. `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` describes the fix loop's step structure and is the site to re-check during implementation.
- **Architecture Alignment** — the engine already established the by-path discipline that keeps bulk out of the orchestrator's context. `skills/review/phases/phase-1-checklist.md` states "The slice content must never transit the orchestrator's context", Phase 3 agents read `{DIFF_PATH}` themselves, and Phase 2 verdicts land in nonce-bound files. Closed issue #218 split a monolithic `SKILL.md` into a thin orchestrator plus phase references; closed issue #693 established caching a value once and handing it to consumers by path.
- **Dependencies** — none beyond what `lib/preflight.sh` already guarantees. This change adds no helper and no new invocation.
- **Data/Schema Considerations** — none. No `iter-<N>.json` field is added, removed, or re-typed.
- **Cross-layer Impact** — the shipped prompt surface only. `scripts/devflow-cloud-writer-contract.json` SHA-256-pins the shipped prompt assets, so the tracked manifest is regenerated in the same commit as the edit.

**The saving does not close the gap.** This change removes between zero and 111k–444k tokens per engine entry, depending on which reading a given run took; its value is eliminating that variance rather than a guaranteed number. It does not make budget-driven truncation impossible: a run that reaches Phase 3.3 at 300k of a 1M window still enters the fix loop facing on the order of 200k tokens of instructions, and open issue #1453 remains the backstop that makes truncated coverage unshippable.

**Related open work.** #1453 gates budget-truncated review coverage at the `workpad.py` Complete write; it is this issue's consequence side, and with #1453 landed while this stays unfixed the runs that degrade silently today would instead block, turning a silent-quality problem into a throughput wall. #1230 closed the elective path — a converging loop may **not** elect `coverage: "not_verified"` to skip a second fan-out (decision recorded 2026-08-09: NO) — and the ~9 minutes measured there is wall-clock while the cost here is window; the two have been conflated. #1299 concerns observability of the engine's own prompt-extension load inside the fix loop, which lands on Step 1's engine-load path; this issue deliberately leaves Step 1's engine resolution and the `<engine-dir>` binding unchanged.

**Out of scope, recorded so it is not re-derived.** The instruction-plus-consequence compression family closed six tickets over this surface (#1419, #1420, #1421, #1424, #1425, #1428) and left exactly one member unfiled: `skills/review-and-fix/references/shadow-review.md`, at 84,791 B the largest file in the fix-loop and engine surface. #1421's body records it as "separately owned"; a full-corpus search finds no such ticket. That is a separate `chore(skills)` issue matching the family's shape, not work for this one.

## Acceptance Criteria

Grounding block: the single changed behavior below is a prose scope amendment to one shipped file, with no reviewer-prompt, roster, coverage, verdict or severity content in scope. The value-comparison rule is discharged for the whole set by the prompt-surface measurement instrument named in the last criterion. Every enumeration below declares its own closure inside the criterion that carries it.

- [ ] The always-resident re-read rule in `skills/review-and-fix/SKILL.md` states its firing scope, and the numbered verification checklist in the Testing Strategy below reaches a single verdict on whether a return from a Phase 3 reviewer dispatched during Step 1 triggers a re-read.
- [ ] The rule names the engine's phase-entry gate as what governs dispatches issued by the engine's own phases during Step 1 and Step 2.6, so a reader is not left to infer that those returns are simply unhandled.
- [ ] The durable-operand contract is unchanged after the edit: `current_step`, `current_substep` and `pending_dispatch` are still the stated step predicate, agent recall is still excluded as a sole predicate, and the absent-operand fail-closed arm is still present — exactly these four, complete by construction.
- [ ] The re-read still fires before a cross-reference routing action, so a routing decision made after a dispatch return is still made against a freshly-read reference.
- [ ] `skills/review-and-fix/references/loop-control.md`'s durable-operand paragraph and the amended rule agree on which reference the operands identify during Step 1, with no sentence in either file contradicting the other.
- [ ] Step 1's engine resolution, the `<engine-dir>` binding, every reviewer prompt, and every roster and coverage rule are unchanged by this change — exactly these four surfaces, complete by construction. This is the stated non-goal that keeps this work from racing #1299.
- [ ] The tracked cloud-writer manifest `scripts/devflow-cloud-writer-contract.json` carries the edited file's regenerated digest, and the manifest check exits zero.
- [ ] The pull request body records the branch's prompt-surface delta as rendered by `scripts/prompt-surface-growth.py`, invoked as a direct leading token.

## Implementation Notes

- **Approach** — a prose scope amendment in one file. There is no coupled artifact beyond the regenerated digest manifest, no generated allowlist literal, no capability grant, and no test pin on the sentence. The change removes tool calls and adds no wait, so it is headless-safe, and it is finding-neutral by construction: no reviewer prompt, gate, verdict rule or roster is touched.

- **Relevant files** — this touches `skills/review-and-fix/SKILL.md` and `scripts/devflow-cloud-writer-contract.json`, plausibly `skills/review-and-fix/references/loop-control.md` if the durable-operand paragraph needs a matching clarification, and plausibly `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md`.

- **Code Patterns** — closed issue #1351 is the working precedent for editing a shipped prompt file under a prose rule with no byte target. `scripts/create-issue-context-eval.py` with `docs/internal/create-issue-context.md` (from closed issue #767) and `scripts/implement-context-eval.py` with `docs/internal/implement-context.md` (from closed issue #1209) are this repository's two existing instruments for this measurement class; the fix loop has no equivalent, and the byte measurement the criteria require is the minimum form of one.

- **Designs rejected during drafting, recorded so they are not re-proposed.** Gating the repeat engine read on hash equality is the largest single saving available (~65k tokens per repeat entry) and is rejected: discharging a read on "the bytes are still in my context" substitutes agent recall for a tool read, which `loop-control.md`'s own durable-operand rule forbids, and under compaction an agent that no longer holds the text but believes it does would improvise a phase from memory. Closed issue #774 is the direct precedent — closed not-planned precisely because safe removal needs an in-run compaction-detection signal nobody has built. Dispatching the engine, or the shadow pass alone, into a subagent is refuted in-tree: `skills/review-and-fix/references/shadow-review.md` records that "a subagent cannot dispatch its own subagents (nested `Agent`/`Task` dispatch is unsupported in the harness — this is structural, not a permissions gap, so granting `Agent` to a subagent does not fix it)", and such a subagent "silently collapses to a degraded single-agent self-check that returns a plausible clean `APPROVE`". That determination carries no recorded date, so the only admissible evidence for revisiting it is a `matcher-probe.yml`-style probe. Relocating the Phase 4.1 markdown report template out of `phase-4-verdict.md` was proposed on the premise that the shadow pass discards its report; that premise is false — `shadow-review.md` instructs the parent to "Collect the engine's Phase 4.1 markdown report, its verdict, the aggregated Phase 3 findings", so every engine entry authors a report and the relocation would move 12,969 B from entry-time to Phase 4.1 with no net saving.

- **Testing Strategy** — governing conventions consulted: `CLAUDE.md`, `CONTRIBUTING.md`, and `docs/internal/`.

  This change has no executable boundary. Its only reader is the runtime agent, so per this repository's recorded decision — agent-executed prompt prose whose only reader is the runtime agent carries no automated regression coverage, and the compensating control is the review pass itself — it gains no wording pin and no source-presence assertion. Adding one would be a new prose pin the repository's own rule prohibits.

  The stand-in is a numbered reproducible verification, each item tied to a criterion above and concrete enough that a second reviewer reaches the same verdict. (1) Read the amended sentence in isolation, with no other file resident, and record a yes-or-no answer to "does a return from a Phase 3 reviewer dispatched during Step 1 trigger a re-read?"; a second reviewer repeating the step must record the same answer. (2) Confirm the amended text names the engine's phase-entry gate as the governing mechanism for those returns. (3) Confirm `current_step`, `current_substep`, `pending_dispatch` and the absent-operand fail-closed arm are each still present in the rule. (4) Confirm the cross-reference firing condition still reads as mandatory rather than advisory. (5) Read `loop-control.md`'s durable-operand paragraph immediately after the amended rule and confirm no sentence in one contradicts the other. The one automated assertion in scope is the cloud-writer manifest check, which must fail first against an edited prompt file whose digest was not regenerated.

- **Documentation Needed** — `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md`

- **Potential Gotchas** — the rule is a shipped prompt surface, so its bytes reach every consumer repo verbatim; a clause narrowed too far would stop the re-read firing on a genuine cross-reference routing action, which is the failure the rule exists to prevent. The manifest digest must be regenerated in the same commit or the required check goes red. Editing this file makes the pull request `engine_self_modifying`, which arms the early shadow pass after iteration 1.

## ✅ Decisions — both blockers resolved (2026-08-08)

Both questions that blocked implementation have been decided. The reasoning is recorded in full so nobody re-derives it.

### 1. The `dirty-tree-compare` fence's portability question

**DECIDED: rewrite the NUL membership test portably, in place — as its own separate change, not here.**

- The one GNU-dependent construct is a single line in `skills/review/phases/phase-3-agents.md`: `grep -qzxF -- "${rec:3}" ".prflow/tmp/review-dirty-tree-before-paths"` — a NUL-delimited exact-match membership test.
- **The portable idiom is already used inside the same fence.** `while IFS= read -r -d '' rec; do` appears at two places in that block, and a third `read -r -d ''` loop drives the restore. So this replaces one line with a construct already proven a few lines above it — not new design work carrying its own test matrix.
- **Independently of any relocation, the construct is arguably already a latent convention violation.** `grep` is not in this repository's preflight-guaranteed set (`git`, `gh`, `jq`, `python3` + PyYAML), and this value *selects which paths get restored*. That is the non-preflight-PATH-tool rule: a missing tool does not fail — the pipeline runs, the value comes out empty, and the wrong thing is selected. Worth fixing wherever the fence lives.
- **Declaring the helper GNU-only was rejected.** It would ship a knowingly degraded backstop on macOS — the platform the no-GNU-flags convention exists to protect — to avoid work that turns out to be a few lines.
- **This rewrite is not dropped along with the relocation.** It ships as its own small change against `phase-3-agents.md`, in place, independent of this issue.

### 2. One issue or two

**DECIDED: the relocation is descoped from this issue entirely.**

This issue ships its **primary fix only** — scoping the always-resident re-read rule. That is prose-only, one file, no coupled artifact beyond the regenerated digest manifest, and it eliminates the 0-vs-111k–444k-token variance that dominates every other term.

Relocating the marked payload region (`devflow:dirty-tree-compare`, and the `devflow:dirty-tree-restore` region nested inside it) out of the per-entry read path is **descoped, not refuted.** The saving — on the order of 6k tokens per engine entry, over a region measuring 12,050 B at `071af828` — is real, and someone may revisit it. It simply does not justify its coupling now: a new cloud-reachable helper needs the conditional vendored-literal invocation form, `lib/test/lint-anchor-fallback-arm.py` enrolment, capability grants across three profiles, regeneration of the generated allowlist literals, and a **`lib/review-profile.tokens` security-boundary lock update** — plus re-anchoring three separately-extracted executable blocks and four equality-pinned assertions in `lib/test/run.sh`.

**The two audit rounds' relocation findings, preserved for whoever picks it up.** Two independent rounds found no defect in the re-read amendment and eight in the relocation: a nested second marker region inside the first; three separately-extracted executable blocks in `lib/test/run.sh` rather than two; four equality-pinned bundle-scoped assertions that go red on the move; an input contract that is one supplied value plus two defaulted paths plus a fixed sentinel plus a cwd assumption rather than three arguments; a `cmp`- and `tr`-derived selection that the non-preflight-tool rule governs; a `REQUIRED_HELPER_HEADS` registration in `lib/test/cloud_writer_contract.py`; a capability grant across three profiles; and a `lib/review-profile.tokens` security-boundary lock. That is a different-sized change from the amendment this issue now carries alone.

The01Geek:
Resolves #1470

## What changed

The fix loop's **always-resident re-read rule** fired after *every* `Agent`/`Task`/`Skill`-tool return while executing a reference. During Step 1 the loop's `current_step` is `"loop-control"` while the orchestrator is actually executing a review-engine phase file, and an engine entry produces dozens of dispatch returns — so it was open whether each of them re-read `loop-control.md`. The two readings differed by the whole cost of the entry.

The rule now names **two** returns of control that fire it, keyed on an operand the producer actually emits:

- an `Agent`/`Task`/`Skill`-tool return that resolves — in whole or in part — a `pending_dispatch` **the active reference stamped**; and
- **the handback that ends an engine entry this reference executed inline** — control coming back to Step 1's verdict routing, or to the Step 2.6 shadow's parse-and-compare. Nothing covered that before, so the loop resumed verdict routing from recall after dozens of intervening returns.

A dispatch made while executing the engine's phases inline, which the reference never stamped — including every Phase 3 reviewer the engine dispatches **during Step 1** — does not fire it; the engine's own phase procedure governs its phases. The **Step 2.6 shadow fan-out keeps firing**, because it is the loop's own stamped dispatch, and "in whole or in part" is what makes each fan-out member's return qualify against a single-slot record.

Both stamping imperatives — `SKILL.md`'s entry-gate item 1 and `loop-control.md`'s — now carry the same scope inline, so producer and consumer agree on the dispatch set. The durable-operand contract is otherwise untouched: `current_step`, `current_substep` and `pending_dispatch` remain the step predicate, agent recall is still excluded as a sole predicate, and the absent-operand fail-closed arm still applies, now with a reachable remedy per operand class. Step 1's engine resolution, the `<engine-dir>` binding, every reviewer prompt, and every roster and coverage rule are unchanged, as is `lib/test/run.sh`.

## Review record

Three review rounds plus a blinded shadow pass. The shadow earned its keep:

- **The change was inert.** An earlier commit reverted the producer scoping to protect a pinned literal, which left `SKILL.md` mandating a stamp before *every* dispatch while the new predicate keyed on "a `pending_dispatch` the active reference stamped". The record carries no issuer field, so an engine-inline stamp is indistinguishable from a loop one and the exclusion selected nothing. **Four of five shadow reviewers found this independently**; the two earlier rounds did not. Fixed by placing the scoping sentence *between* the two pinned literals rather than editing either, so both still resolve exactly once and no pin census re-keying is needed.
- **Two CI-breakers, caught before merge.** A `/simplify` edit broke an `assert_pin_unique` literal; the attempted fix for it de-keyed that pin's frozen-census row and turned the `#810` mutation-routing gate RED. Both reverted.
- **Also fixed from the shadow:** the absent-operand arm's single remedy could not re-stamp a missing `pending_dispatch`; the rule's opening sentence read as the full predicate while a second, structurally different trigger followed it; and a single-slot `pending_dispatch` meant fan-out members 1..N-1 did not "resolve" it.

**Shadow pass: dispatched, full roster.** Six blinded reviewers — the four always-on plus `type-design-analyzer` and `pr-test-analyzer`, both dispatched by the absent-comparand tripwire (this loop persisted no `iter-N.json`, so the tripwire had no comparand and failed closed to the full gated roster). All six returned and joined 1:1.

**Deferred to #1488:** the fail-closed arm's in-flight discriminator resolves from the orchestrator's live context rather than from `iter-<N>.json`. Graded Minor and non-blocking; closing it needs an issuer field in the record, which this issue's Data/Schema Considerations rule out.

**Verification:** `cloud_writer_contract.py check` — closure OK. Suite results at this head are in the PR checks.

### Prompt-surface size

Derived at `761fdd7e9ea94f72b810a3637bbd5d366b4fd52c` against merge-base `4e4a0c0e3ac34b67a134ac2f1dc99f2f3d64ab68` (`origin/main`). Covered: tracked `*.md` under `skills/`, `agents/`, `.prflow/prompt-extensions/`.

| Path | Δ bytes | Bytes at HEAD |
| --- | ---: | ---: |
| `skills/review-and-fix/SKILL.md` | +1,123 | 35,613 |
| `skills/review-and-fix/references/loop-control.md` | +331 | 44,702 |
| **Whole covered surface** | **+1,454** | **2,107,897** |

Generated via `/prflow:implement 1470`


The01Geek:
# PRFlow Review — PR #1487



**Status:** 🚀 Reviewing
**Diff profile:** _(pending Phase 0.5)_
**Run:** _(local run)_
**Last updated:** pending

## Blueprint
- [ ] Classify diff (Phase 0.5)
- [ ] Review agents (Phase 3)
- [ ] Run complete — everything this run owed

## Findings (live)
_(Phase-3 findings appear here as each agent returns.)_

## Verdict
_(pending)_

PR #1599, opened 2026-08-11
Cap boundary-gated references and skill roots at 95% of the reader's Read cap

### LINKED ISSUE #1595: Cap the size of boundary-gated references and skill roots at 95% of the reader's Read cap
## Problem Statement

PRFlow loads its phase and reference files progressively behind a **boundary gate**: the file's first
line must be its `start` marker and its last line the matching `end` marker, and the loading agent
verifies both before acting on the body. Seeing the `start` marker and no `end` marker is the
`truncated` shape, which `/prflow:implement`, `/prflow:review`, `/prflow:review-and-fix`, and
`/prflow:docs-verify` all treat as fail-closed.

The Claude Code Read tool has a hard **25,000-token cap**. A file above that cap returns only its
first page, so a single whole-file read yields `start` and not `end` — on a file that is intact on
disk. Nothing in the repository measures the size of these files, so an author gets no signal at edit
time and no signal in CI, and nobody trimming one has a number to aim at. Ten files are over the
threshold today.

## Current Behavior

**Measured against `origin/main` at `049e73a59`, 2026-08-11.** The population is 57 files — 39
boundary-gated references plus 18 skill roots. Ten exceed 61,750 bytes:

| bytes | file |
|---|---|
| 134,965 | `skills/implement/phases/phase-2-implement.md` |
| 110,140 | `skills/implement/phases/phase-3-review.md` |
| 83,427 | `skills/retrospective-weekly/SKILL.md` |
| 81,869 | `skills/create-issue/references/step-3-6-audit.md` |
| 75,961 | `skills/review-and-fix/references/shadow-review.md` — open PR #1593 |
| 75,922 | `skills/implement/phases/phase-4-documentation.md` — open PR #1592 |
| 73,799 | `skills/review-and-fix/references/fixing.md` — open PR #1591 |
| 68,901 | `skills/implement/phases/phase-1-setup.md` — open PR #1590 |
| 65,822 | `skills/review/SKILL.md` |
| 62,267 | `skills/init/SKILL.md` |

All four `/prflow:implement` phase files are in that set.

**Four of the ten have trims already open, none merged.** All four are open, CI-green, and not
drafts; `main` is unchanged, so all ten are still over the threshold as this issue is filed. Whoever
implements the guard should re-derive the list at that time rather than re-trimming a file whose fix
is already in flight:

| PR | file | on `main` | lands at | margin under 61,750 |
|---|---|---|---|---|
| #1590 | `phase-1-setup.md` | 68,901 | 61,731 | 19 bytes |
| #1591 | `fixing.md` | 73,799 | 61,691 | 59 bytes |
| #1592 | `phase-4-documentation.md` | 75,922 | 61,742 | **8 bytes** |
| #1593 | `shadow-review.md` | 75,961 | 61,599 | 151 bytes |

Those margins are themselves evidence for the guard and a caution about it. Each trim aimed at the
number and landed within 151 bytes of it, one within 8, so every one of those files will breach again
on the next ordinary sentence added to it. A file that clears the threshold by single-digit bytes is
compliant and not comfortable, and the guard firing on it later is the design working rather than a
false positive.

**The threshold is not insensitive.** `skills/implement/SKILL.md` at 61,039 bytes is the largest file
under the line, with **711 bytes** of headroom, and the smallest file over is 62,267. Moving the
threshold anywhere inside that 1,228-byte window changes which files are covered. The next files down
are `skills/receiving-code-review/SKILL.md` 57,887, `skills/review-and-fix/references/loop-exit.md`
56,324, and `skills/review/phases/phase-3-agents.md` 54,804.

**Verified: the cap is 25,000 tokens.** Reading a file whole with the Read tool returns a notice
carrying its token count and the cap. Observed 2026-08-11: `step-3-6-audit.md` 28,356 tokens,
`phase-1-setup.md` 26,496 tokens, `fixing.md` 26,053 tokens, each against `cap 25000`. Those three
are the measured members of the ten; the rest are over the byte threshold with token counts not
measured here.

**Verified: two marker families exist, and both gate fail-closed.**
`skills/implement/SKILL.md` states: "**On any boundary row: stop that phase**, report the stop label
with the phase id and the reference path, and do **not** act on the body, improvise the phase from
its stub/orientation text, or repair the file." `skills/review-and-fix/SKILL.md` uses a different
marker family and states that a truncated result and a result missing either marker "all mean
**reference unreadable** and take the failure-map row below."

**Verified: the absence was noticed before and never closed.**
`docs/internal/cutovers/issue-815-deferred-ac-followups-relocate.md` states: "**No byte ceiling on
this file is enforced anywhere in the tree, and none ever was.**"
`docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` states that keeping references lean "is a matter of
authoring discipline, not a mechanically enforced ceiling."

**Verified: marker integrity is asserted; size is not.** `lib/test/run.sh` and
`lib/test/modules/create-issue-contract.sh` assert marker position, count, and self-naming for each
family. A sweep of `lib/test/`, `scripts/`, `lib/*.sh`, and `.github/workflows/` for byte, line, or
token measurement over `skills/**` and `agents/**` returns nothing.

## Desired Behavior

The suite fails when a file in the covered population exceeds 61,750 bytes. The covered population is
every boundary-gated reference plus every skill root.

The ten files already over are covered by **expiring exemptions**, not by a permanent allowance. Every
other file — every new file, and every file that becomes boundary-gated later — is enforced at the
real threshold from the day this lands. An exemption is attributable to the one file it covers,
carries a recorded condition under which it ceases to apply, and fails the suite once that condition
is met, so the exemption list shrinks by the suite noticing rather than by anyone remembering. Adding
a file to that list is not a way to pass the check.

## User Impact

`/prflow:implement` reads a phase file at every phase entry and again at each mid-phase re-anchor, and
all four of its phase files are over. `/prflow:review-and-fix` reads two over-threshold references on
every shadow pass and fix iteration.

**Skill roots are covered because the orchestrator is the worst thing to lose.**
`skills/implement/SKILL.md` is the always-resident orchestrator, and it holds the re-anchor triggers
and boundary contract that tell a run how to recover from exactly this failure. A truncated
orchestrator loses the rules for recovering from truncation. It is 711 bytes under the line today,
which makes it the single most consequential file in the repository to cross it, and it is why the
guard covers skill roots rather than boundary-gated references alone.

**The bound stated honestly:** what is established is the contract text, the file sizes, and the
Read-tool cap. What is **not** established is an observed halted run — the truncation notice names an
offset for the next page, so an agent that pages through eventually holds both markers, and no run
halting on this cause has been recorded. Whether a paged read satisfies the boundary contract is
tracked separately and is out of scope here; that question bounds this issue's severity claim, and
this issue does not resolve it. The exposure this closes is unsignalled growth, not a proven outage.

## Technical Context

> **Scope note:** The files and details below are the known starting points, not the full list. Before
> implementing, trace the change through the codebase to find every affected call site, consumer, and
> layer — this issue maps the work, it does not bound it.

- **Relevant Classes/Files** — `lib/test/lint_population.py` owns the shared population contract for
  this repo's `git ls-files` lints and states that its index-reading form "reads the index only (no
  `--others`) and is worktree-immune on a bare clone". The existing `lib/test/` static lints are the
  shape precedent. `docs/internal/implement-context.md` carries the record that must be reconciled
  (below).
- **Documentation Drift** — no internal document states a size limit, byte budget, or
  read-truncation hazard for these files, and none enumerates the covered population as one set.
  `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` §7 names the symptom, "a read that returns a partial
  body", and names no cause.
- **Architecture Alignment** — this is one more member of the repo's desk-time static-lint family,
  driven from the suite the same way the others are.
- **Dependencies** — `git` and `python3` only, both preflight-guaranteed.
- **Cross-layer Impact** — the suite, the coverage map, `CLAUDE.md`,
  `docs/internal/implement-context.md`, and a changeset. No `skills/**` file is edited.

**The two marker families.** One is an HTML comment pair `<!-- prflow:<command>-ref … start -->` /
`… end -->`, used by `/prflow:implement`, `/prflow:review`, `/prflow:create-issue`, and
`/prflow:docs-verify`, whose contracts specify the **literal** first and last lines. The other is a
`# Reference: <title>` heading paired with an `<!-- END <name>.md -->` marker, used by
`/prflow:review-and-fix`, whose contract specifies the first and last **non-blank** lines. The second
family contributes two of the ten, so a one-family implementation reports a clean audit that is wrong.
Two files under `skills/create-issue/references/` carry no markers and are correctly outside the
gated half; issue #1467 would add markers to one of them, which grows the gated population by one.

**Threshold derivation, and why the floor.** The cap is in tokens and a check can only measure bytes,
so the conversion is the calibration. Measured densities: `step-3-6-audit.md` 2.887 bytes per token,
`fixing.md` 2.833, `phase-1-setup.md` 2.600. The threshold is **0.95 × 25,000 × 2.60 = 61,750
bytes**, derived from the **floor** of the observed densities and never their mean — a denser file
truncates at a smaller byte size, so the mean would set a threshold that passes files the reader
cannot return. The mechanism behind the spread: the cap counts the tokens of the *rendered* read
output, which carries a line-number prefix per line, so bytes-per-token falls as lines get shorter.
The floor absorbs that overhead rather than modelling it. The residual is stated rather than closed —
a file that tokenizes below 2.60 bytes per token truncates below 61,750 bytes and passes.

**The skill-root half rests on an unmeasured premise.** A skill root reaches an agent through the
Skill tool rather than the Read tool, and whether that path shares the Read tool's cap has not been
established here. The guard covers skill roots because the downside of leaving the orchestrator
unguarded is severe, not because the cap is known to apply. The separate investigation into the
Skill-tool loading limit is what will later confirm this half or relax it; this issue does not
pre-empt that finding.

**Reconciling the #765 record.** Issue #765 retired the static word-budget apparatus, and
`docs/internal/implement-context.md` currently states that it adds "**no** new static word-count or
prompt-length gate." This change adds a static byte gate, so that record stops being true and is
reconciled here. The distinction is real and is what the reconciliation must state: #765 retired an
**authoring** budget — a judgment about how long prose ought to be — while this is a
**reader-capability** ceiling, a property of what the tool can return. A future reader needs that
distinction where they will actually find it, not only in this issue.

**No new cloud grant is needed.** Nothing here invokes a helper that
`prflow_implement.allowed_tools` does not already grant as a direct leading token.

## Acceptance Criteria

The measurement instrument for every byte figure below is the length of Python `Path.read_bytes()`
over the file named. A file is **boundary-gated** when its first non-blank line is a `start` marker
and its last non-blank line is the matching `end` marker of the same family. A **skill root** is a
`SKILL.md` at the top level of a skill directory. The **covered population** is both together.

- [ ] The suite fails when a file in the covered population exceeds 61,750 bytes and holds no live
      exemption.
- [ ] The covered population is derived by inspecting boundary markers and by the skill-root shape,
      covers both marker families, and is not a transcribed list of paths, so a new skill and a
      newly-gated reference are both covered without a second edit.
- [ ] The covered population is sourced from an index-reading `git ls-files` enumeration with no
      `--others`, and no check added by this change performs a repository-root-anchored recursive
      filesystem walk.
- [ ] An audit of the repository tree that cannot establish its population, or that selects nothing
      for a marker family or for the skill-root shape, fails rather than reporting a clean pass. This
      is scoped to the whole-tree audit; a run given an explicit narrower population is not required
      to see every family.
- [ ] An audit that cannot read a file in its enumerated population fails, naming that file, rather
      than dropping it and reporting a clean pass. Membership is decided by reading each file, so an
      unreadable file is one whose membership was never established.
- [ ] Each exemption is attributable to exactly one file, and no exemption applies to a file it does
      not name.
- [ ] Each exemption records the condition under which it stops applying, and the suite fails once
      that condition is met, so an exemption that has outlived its reason is surfaced by the suite
      rather than by someone remembering to look.
- [ ] The live exemptions are enumerable, so their count and their membership can be read off the
      repository at any commit without re-deriving them from history.
- [ ] A file that holds no exemption when this change lands cannot acquire one later: the check
      refuses an exemption naming a file outside the recorded set, so exempting a file is not
      available as a way to make the check pass.
- [ ] Every failure message names the offending file, its measured size, the limit it breached, and
      the remedy. For a file over the threshold the stated remedy is to trim the file; it is never to
      exempt it.
- [ ] The 61,750-byte threshold is derived in the source from separately recorded constants for the
      reader's cap in tokens, the safety factor, and the bytes-per-token density floor, each carrying
      its derivation, and the source states that the constant is the floor of the measured densities
      rather than their mean.
- [ ] Against the tree as it stands after this change, the new checks pass, and the count of files
      they audited is reconciled against a population count derived independently of the check's own
      selection logic.
- [ ] The recorded exemptions cover the files named in Current Behavior that are still over the
      threshold when this change lands, and no others.
- [ ] An exemption record that is not well-formed fails closed with a breadcrumb naming the shape
      rejected.
- [ ] `CLAUDE.md` names the threshold and the check that enforces it.
- [ ] `docs/internal/implement-context.md` no longer states that no new static prompt-length gate
      exists, and the surviving record states the distinction between the authoring budget issue #765
      retired and the reader-capability ceiling this change adds.
- [ ] Every coupled site this change touches is reconciled in the same change, including any suite
      registration or generated artifact the new checks require.
- [ ] `lib/test/run-parallel.sh` reports zero failures, and any skip it reports is surfaced with its
      kind and reason rather than read as a clean pass.
- [ ] The change adds a `.changeset/*.md` file with `bump: patch` frontmatter.

## Implementation Notes

**Approach.** Add a size check over the covered population, driven from the suite alongside the
existing static lints, with the current offenders carried as expiring exemptions rather than as a
permanent allowance. Trimming the exempted files is separate work, partly tracked already in #1581,
#1557, and #1590, and this check is what gives those trims a number to aim at.

Relevant files: this likely touches `lib/test/` (the check, its exemption record, and the suite
driver), `lib/test/lint_population.py` as an import, the coverage map, `CLAUDE.md`,
`docs/internal/implement-context.md`, and `.changeset/`.

**Testing Strategy.** The check has an executable boundary — an exit code and emitted verdict lines —
so it is covered by ordinary automated tests at that boundary, written before the code and proved RED
first by producing the wrong verdict rather than an absent-file error.

Cover: a file at the threshold and a file one byte above; an exempted file within its exemption and
one whose exemption has expired; an exemption naming a file outside the recorded set; an exemption
naming a file that is not in the covered population; an empty selection for each marker family and
for the skill-root shape; an unreadable file in the population; an unestablishable population; and a
file carrying a `start` marker with no matching `end`, which is outside the gated half and must
produce no violation.

The exemption record is hand-editable, so it is a best-effort parser: walk the repo's adversarial
shape matrix over it. `governing conventions consulted: CLAUDE.md` (the six-shape matrix for
config-JSON consumers), `CONTRIBUTING.md`, and testing guidance under `docs/internal/`. The matrix
has accepting rows here, so it is not a list of rejections.

Test the expiry itself, not only the size check. An exemption whose condition is met must fail the
suite; without that assertion the expiring property is a claim rather than a mechanism, and the
exemption list becomes the permanent allowance this design exists to avoid.

**Documentation Needed** — `CLAUDE.md`, `docs/internal/implement-context.md`.

**Potential Gotchas.**

- An exemption list is where work goes to die. The expiring property is the entire difference between
  this design and a permanent allowance, so an implementation that records the exemptions but never
  fails when one outlives its reason has delivered the allowance and not the guard.
- The threshold sits in a 1,228-byte window between the largest covered file under it and the
  smallest over it. It is not a value that can be nudged for convenience; moving it changes coverage.
- `skills/implement/SKILL.md` has 711 bytes of headroom. Any change that lands prose in the
  always-resident orchestrator should expect to hit this guard, and trimming it is the intended
  response.
- "Two families" understates the collapse exposure. The first family spans four command namespaces
  with very uneven memberships — one has a single member — so a regex slip that drops a whole
  namespace can leave a family-level count positive and the audit green. Reconcile the audited count
  against an independently derived population at a granularity fine enough that losing one namespace
  is visible.
- A population sourced from a recursive filesystem walk descends into sibling checkouts under
  `.claude/worktrees/` and produces a count that varies between local runs on the same commit while
  CI stays green.
- An empty audited population must fail. "Audited nothing" reading as "audited everything, found
  nothing" is the failure mode this whole check exists to prevent, one level up.
- The two families define the boundary line differently — one by the literal first and last lines,
  the other by the first and last non-blank lines. Picking one rule for both is a decision to make
  deliberately and record, not to inherit by accident.
- The fixtures must be real files with real bytes. A harness that supplies only a path list, copied
  from a check that judges path strings, leaves every size expectation passing for the wrong reason.
- The threshold is a byte proxy for a token cap, so its correctness rests on the recorded density
  floor rather than on the byte number.

The01Geek:
## Summary
- A markdown file that an agent must read whole can grow past the reading tool's 25,000-token limit, at which point the read comes back truncated and the instructions at the end of the file silently go missing. Nothing in this repository noticed when that happened.
- This adds a check that measures every file an agent is required to read in full and fails the test suite when one grows past a byte ceiling derived from that token limit.
- Six files are already over the ceiling. They are recorded as expiring exemptions rather than being trimmed here, so the check can land without blocking on unrelated edits, and each exemption removes itself once its file is trimmed.

## Changes

**New check — `lib/test/lint-reference-size.py`**: Caps each covered file at 61,750 bytes. That number is 95% of the reader's 25,000-token cap converted at 2.60 bytes per token, which is the *floor* of the three measured byte-and-token pairs recorded in the file, never their mean — a mean would admit files the reader cannot return whole. The 5% margin leaves room for a sentence or two before the next read truncates.

**Covered population**: Derived from the tree rather than from a hand-maintained list. Each tracked `.md` file's first and last non-blank lines are read against both boundary-marker families, and every `skills/<name>/SKILL.md` at any depth is included regardless of depth. That selects 58 files on the current tree. A file that opens with a start marker but has no matching end marker is deliberately outside the population.

**Exemption record — `lib/test/reference-size-exemptions.json`**: Six already-oversized files carry expiring exemptions. Each exemption names exactly one file, and one naming a file outside the record's frozen `recorded_set` is refused. Both a live exemption and an orphaned roster row turn the suite red once the file it names drops to or under the ceiling, so the record cannot outlive the problem it documents.

**Self-check mode**: `--self-check` proves the 2.60 density floor does not exceed the minimum of the recorded `(path, bytes, tokens)` measurements. The recorded byte figures are the sizes of the copies measured at the time, not current repository sizes — two of the three files have since been trimmed — and they are stored beside their token counts precisely so no one divides a stale token count by a current file size.

**Suite wiring — `lib/test/run.sh`**: A new `#1595` block drives the check, covering verdicts either side of the threshold, the start-marker-without-end exclusion, the live/expired/outside-the-recorded-set/outside-the-population exemption semantics, the hand-editable record's malformed-shape matrix, the whole-tree empty-family and empty-skill-root refusals, the unreadable-file and unestablishable-population fail-closed arms, the threshold derivation, and a real-tree run reconciled against an independently derived per-namespace population.

**Coverage map and documentation**: `lib/test/modules/coverage-map.json` records the new surface's owner. The system overview, the create-issue and implement context pages, and two deferred-findings records are reconciled with the ceiling this introduces, and a changeset records the bump.

## Resolves
Resolves #1595

## Test Plan
- [ ] `lib/test/lint-reference-size.py` exits 0 against the current tree and reports the audited-file count.
- [ ] `lib/test/lint-reference-size.py --self-check` exits 0 and prints each recorded density along with the derived ceiling.
- [ ] `lib/test/run-parallel.sh` completes with an `aggregate CLEAN` line, zero failures and zero skips. **Not yet run on this tree** — two of the issue's acceptance criteria name execution outcomes that only this whole-suite pass establishes, and they remain unverified until it has run.
- [ ] `git ls-files '*.py' | xargs -r ruff check` is clean, and the shell lint over the tracked `.sh` files is clean.
- [ ] Confirm each of the six exemption entries still names a file that is genuinely over the ceiling, and that removing one turns the suite red.

## Visual Changes
N/A

## Breaking Changes
None. The check is additive: it fails the test suite on an oversized covered file and changes no runtime behavior of the plugin itself.

### Prompt-surface size

prompt-surface growth: no tracked `*.md` under `skills/`, `agents/`, or `.prflow/prompt-extensions/` changed between `9f5fce34cba29e1d77888894b5e1ddadacadb07b` and `07f816a63a5c91d90984cae6cf733d5d0077c270` — no table rendered.

Generated via `/prflow:implement 1595`


The01Geek:
## Review reception — five sub-threshold findings

Triaged under `.receiving_review.fix_severity_threshold` = `suggestion` (this repo's configured value, not the `critical` default), so all five re-open the diff and each needed a stated disposition.

### Addressed (commit `6a79071`)

**1. `--self-check`'s failure arm and the non-positive-token guard were never driven.** Confirmed: only the `rc=0` arm was asserted. Both refusing arms are now driven against **mutated copies** under a temp root — the working-tree lint is never modified.

The finding's own reasoning is why this mattered most: an inverted comparison keeps the suite green while the ceiling it certifies rests on nothing. Each refusal is attributed to its **own message** rather than a bare exit code, and that turned out to be load-bearing — dropping the non-positive-token guard *still* exits 1, via a `ZeroDivisionError`, so an exit-status-only assertion would have read as a pass against exactly the mutation it was written to catch.

**2. The whole-tree empty-family-A completeness refusal was untested.** Confirmed: `RSZ_NOB` and `RSZ_NOSR` covered family B and skill roots; `gated-a` had no fixture. Added `RSZ_NOA` in parallel, carrying a positive control so the refusal is attributable to the empty family rather than to a fixture that lost its tree.

**3. `load_record` fail-closed branches lacking coverage.** Confirmed and swept wider than cited — five shapes, not three: `_entry`'s missing-key arm on **both** sides of the record, an entirely absent `schema_version`, the exemptions-side empty path, and a non-string `expires_when`. Every needle names the rejecting guard's own breadcrumb, so a rejection from a neighbouring guard cannot masquerade as the one under test.

Each added assertion was **mutation-checked** against a copy — inverting the self-check comparison, dropping the non-positive-token guard, removing `gated-a` from `COVERED_KINDS`, and removing `_entry`'s missing-key check each turn their assertion red for the reason it pins.

### Declined, with reasoning

**4. `typing.Literal["gated-a","gated-b","skill-root"]` for the `kind` enum.** Declined — the proposed remedy cannot deliver what the finding asks of it.

The finding's stated benefit is making the three-way agreement *checker-visible*. This repository runs **no Python type checker**: CI's lint job is `lint (shellcheck + actionlint + ruff)`, and ruff performs no type inference that would flag a `Literal` mismatch. (The `mypy` strings in the tree are cloud-allowlist grant tokens, not a CI gate.) The annotation would therefore be inert decoration.

The hypothesised failure is also **fail-loud, not silent**. A typo in either `classify` or `COVERED_KINDS` makes that kind's `any(...)` permanently false, so the whole-tree audit appends its refusal on every run — and `#1595 the real-tree audit is clean as the tree stands` turns the suite RED immediately.

The residual the finding gestures at is real but different: adding a family to `classify` **without** adding its `COVERED_KINDS` row ships an unchecked family. A `Literal` alias would not catch that either — it would need a runtime reconciliation of the two sets. That coupling is already called out in the inline comment above `COVERED_KINDS`. Revisit if a type checker is ever added to CI, or if a fourth family lands.

**5. `NamedTuple(kind, detail, size)` for the `covered` triple.** Declined as polish.

The mis-index risk is confined to two reads (`entry[0]`, `entry[2]`) inside one function, and it is fail-loud: `entry[1] <= MAX_BYTES` compares `str` to `int` and raises `TypeError` rather than silently mis-measuring. Set against the repo's standing preference for minimal diffs on an approved PR, the change is churn without a defect behind it. Revisit if the tuple gains a fourth member or escapes this function.

---

Both declines are recorded here rather than only in the commit message, so a later pass can down-rank them instead of re-litigating from scratch. No ceiling constant (`READER_CAP_TOKENS`, `SAFETY_FACTOR`, `BYTES_PER_TOKEN_FLOOR`, `MAX_BYTES`), no exemption row, and no skill file was touched by this pass.

Issue #1550, opened 2026-08-10
Test that workpad.py --help covers the flags and subcommands the implement phase files invoke


The01Geek:
## Problem Statement

PR #1549 (issue #1531) deleted `skills/implement/SKILL.md`'s duplicated `workpad.py` subcommand and flag tables and replaced them with a pointer to `workpad.py --help` / `workpad.py update --help`.

That makes the helper's argparse help the **sole** documentation surface the implement orchestrator reads for the workpad CLI — and nothing asserts it covers what the run actually invokes. A `help=` string dropped in a future refactor would silently leave every implement run without its reference, with no test going RED.

Raised by the `prflow:pr-test-analyzer` review agent on PR #1549 and deliberately deferred there: the finding is sound, but it is outside issue #1531's acceptance criteria, and a derived-set assertion added at the end of that run carried its own drift risk.

## Current Behavior

`lib/test/test_python_scripts.py` already carries one instance of this shape — `#1462 update --help advertises --reconcile-extension-rows` — but it pins a single hand-picked flag rather than the set the phase files actually drive.

## Desired Behavior

A behavioral test that:

- derives the set of `--flag` tokens and subcommand names actually invoked from `skills/implement/**/*.md`, and
- asserts each appears in the output of a real `workpad.py --help` / `workpad.py update --help` subprocess run — not a source grep.

The neighbouring `_live_ticks` block in the same file is the precedent for the derived-set style.

This is **not** replacement coverage for the two prose pins #1531 retired — there was no behavior there to replace. It covers a new dependency that change introduced.

## Acceptance Criteria

1. A test in `lib/test/test_python_scripts.py` derives the invoked flag/subcommand set from the shipped implement surface rather than hardcoding it.
2. It asserts each derived token appears in real `--help` output captured from a subprocess run.
3. It fails RED when a `help=` string for an invoked flag is removed.
4. The suite and both shellcheck lints pass.

## Implementation Notes

- `lib/test/test_python_scripts.py` — the `#1462` block and the `_live_ticks` derivation are the models.
- `scripts/workpad.py` — the argparse definitions under test.
- Disclosed residual worth stating in the test: a flag whose operand is a shell variable is outside the derivation, the same residual `_TICK_RE_1462` already discloses.

PR #1660, opened 2026-08-13
Neutralise terminal colour in the focused Python test so a colour-forcing host stops producing 39 false failures

### LINKED ISSUE #1653: Neutralise terminal colour in the focused Python test so a colour-forcing host stops producing 39 false failures
## Problem Statement

The repository's focused Python test file spawns the workpad command-line tool as a child process and reads the help text it prints. It does not tell that child to turn colour off. On a machine whose environment forces colour on, and whose Python is new enough to colourise help output, the help text comes back full of terminal escape codes, and 39 assertions that check "this flag has a description" stop matching. No production code is broken; one test file is reading the wrong bytes. The change makes that test file turn colour off for the child processes it starts, so the same tree gives the same answer on a machine that forces colour and on one that does not.

## Current Behavior

Running the focused Python test directly on a machine that forces colour produces 39 failures that have no cause in the code being tested.

- Measured on this checkout, on macOS with Python 3.14.6 and `FORCE_COLOR=3` exported: `FORCE_COLOR=3 python3 lib/test/test_python_scripts.py` reports **4883 passed, 39 failed**, while `env -u FORCE_COLOR python3 lib/test/test_python_scripts.py` on the identical tree reports **0 failed**.
- All 39 failures sit in one block and are all of one shape — 10 named `#1550 workpad.py --help documents (help= present) the subcommand …` and 29 named `#1550 update --help documents (help= present) the invoked flag …`. The sibling assertions in the same block that only check the flag token is listed all pass.
- The cause is that Python's argument parser colourises the usage line, the program name and the flag lines when colour is forced. Under `FORCE_COLOR=3` the help output begins with the bytes `ESC[1;34musage: ESC[0m`, and a flag line reads `  ESC[1;36m--reflection-kind ESC[0m …`. Both detectors identify a flag line by the characters it starts with, and the leading escape sequence breaks that identification.
- The same parser does **not** colourise the help *description* text. That is why the other rendered-help checks in this repository still pass: they match description phrases, not flag lines. Those checks live in `lib/test/run.sh`, `lib/test/modules/create-issue-contract.sh`, `lib/test/modules/parallel-suite-runner.sh`, `lib/test/modules/regenerate-artifacts.sh` and `lib/test/modules/issue-audit-state.sh`.
- The block's own negative control is defeated by the same bug. It proves the detectors are not vacuous by stripping a description from a copy of the tool and confirming the detector then reports "no description". Under forced colour the detector reports "no description" for everything, so the control passes while proving nothing.
- The whole-suite paths are unaffected, because the shared test harness already turns colour off before running this file. Only the direct invocation is exposed — and the direct invocation is exactly the form the project's own instructions mandate for a focused Python test.
- Continuous integration is green for two independent reasons: it sets no colour-forcing variable, and it pins Python 3.11, which predates help colourisation.
- The shell test modules that read rendered help were run on the same host under `FORCE_COLOR=3` during drafting and are unaffected: `create-issue-contract` reported 300 passed and 0 failed, and `regenerate-artifacts` reported 372 passed and 0 failed.

## Desired Behavior

The focused Python test file turns terminal colour off for itself and for the child processes it starts, at the top of the file, above its first child-process invocation. A developer and an automated run each get the same result from that file on a host that forces colour and on one that does not, and both when the file is run directly and when it is run through the shared harness. A short page under the repository's internal documentation states this as a convention for tests that read a command's rendered output, and names the one place in the repository that deliberately does the opposite.

## User Impact

The people who benefit are developers and autonomous implementation runs working in this repository on a machine with a colour-forcing environment and a recent Python.

- Today such a run sees 39 failures appear immediately after its own edit. The obvious conclusion is that the edit broke the workpad command-line tool, which is false.
- The project's own guidance forbids waving such a failure off — `CLAUDE.md` states, verbatim, "**There is no known-flake set — every FAIL is a real failure to diagnose.**" Following that guidance correctly costs an entire extra suite cycle to reach a conclusion the environment, not the code, determined.
- The reporter of this defect did exactly that, and paid that cost.

## Technical Context

**Relevant Classes/Files**

- `lib/test/test_python_scripts.py` — the focused Python test file. Its `#1550` block spawns the workpad tool twice to read help text, and two helper functions decide whether a description was rendered.
- `lib/test/module-harness.sh` — the shared harness that every suite runner uses to execute the file above.
- `lib/test/modules/issue-audit-state.sh` — the only help probe in the repository that already neutralises colour.
- `lib/test/modules/regenerate-artifacts.sh` — the one place that deliberately tolerates the escape byte instead of removing it.
- `docs/internal/` — where the new convention page belongs, as `docs/internal/test-suite-probe-conventions.md`.

**Verified:** the two probes pass no environment to the child, so it inherits the host's colour-forcing variable. `lib/test/test_python_scripts.py` contains verbatim "_top_help_1550 = _sp1550.run(", then verbatim "[sys.executable, str(SCRIPTS / 'workpad.py'), '--help'],", then verbatim "capture_output=True, text=True)" — no env= argument appears in that call.

**Verified:** the shared harness already neutralises colour, which is why only the direct invocation fails. `lib/test/module-harness.sh` contains verbatim the comment `# PYTHON_COLORS=0 keeps the captured diagnostics deterministic: a host that` followed by `# forces color (FORCE_COLOR) would otherwise interleave ANSI codes into the`, and the invocation `  if PYTHON_COLORS=0 python3 "$script_path" > "$output_path" 2>&1; then`.

**Verified:** the repository already has a working precedent for the fix, with the hazard named in its own comment. `lib/test/modules/issue-audit-state.sh` contains verbatim `# NO_COLOR/PYTHON_COLORS: argparse colorizes its help on python >= 3.13 when the` followed by `# rendering path allows it, and the ANSI escapes would land INSIDE a pinned phrase and` and `# fail the match on exactly the newer interpreters this repo supports.`, immediately above `IAS_HELP_546="$(NO_COLOR=1 PYTHON_COLORS=0 python3 "$IAS" --help 2>&1 | tr -s '[:space:]' ' ')"`.

**Verified:** one site deliberately works around colour rather than removing it, and is to be named as a sanctioned exception rather than changed. `lib/test/modules/regenerate-artifacts.sh` contains verbatim "# argparse's help is ANSI-colored here, so the boundary character is commonly an escape" and verbatim "# byte rather than a space; `[!-]` accepts either and only excludes the hyphen that a".

**Verified:** interpreters new enough to colourise help are supported by this project, so the failure is not an unsupported-configuration artefact. `lib/preflight.sh` states verbatim "# python3 (>=3.11) with PyYAML. Date math and text extraction were written to", and a search of that file for an upper version bound returned nothing.

**Verified:** continuous integration cannot observe this defect, because it runs an interpreter that predates help colourisation. `.github/workflows/ci.yml` contains verbatim "uses: actions/setup-python@v6" three times, and the line following each of those three occurrences pins the version 3.11.

**Verified:** nothing in this change can reach a repository that installs this project as a plugin. `.github/actions/vendor-plugin/vendor-slice.sh` contains verbatim `  rm -rf "$stage/docs/site" "$stage/docs/external" "$stage/docs/internal" "$stage/lib/test"`, so the entire test directory is removed before the plugin is packaged.

**Verified:** setting environment variables at the top level of this test file is already the established idiom there. `lib/test/test_python_scripts.py` contains verbatim "os.environ['DEVFLOW_IAS_ACQUIRE_WINDOW_S'] = '0.30'", written at column zero and so at module scope.

**Documentation Drift**

The Step 1 documentation pass returned **drift found**, in the form of a missing convention plus three inline comments that contradict one another.

- No page under `docs/internal/` states any convention for tests that read a command's rendered output. `docs/internal/implement-skill.md` says to `pin the rendered `--help`/stderr surface instead` but says nothing about the environment that rendering happens in.
- `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` documents colour neutralisation for one harness helper only, describing `devflow_run_focused_python_test` as running `one standalone `python3` suite with `PYTHON_COLORS=0``. A reader can reasonably take that as the project's colour policy, while the probe family where the hazard actually bites is undocumented.
- The three inline comments quoted above give three different remedies for the same hazard: neutralise the environment, pin the terminal width, and tolerate the escape byte. The new page reconciles them.

**Architecture Alignment**

The fix follows the existing precedent in `lib/test/modules/issue-audit-state.sh` and in `lib/test/module-harness.sh`, both of which neutralise colour by environment variable rather than by post-processing text. Placing the assignment at module scope in the test file matches how that file already handles environment setup.

**Dependencies**

The Python standard library `os` module, already imported by the test file. No new dependency.

**Data/Schema Considerations**

None. The probes capture output in memory and write nothing to disk.

**Cross-layer Impact**

Confined to the test layer and the internal documentation. No shipped surface, no workflow, no configuration key, and no consumer-facing artefact is touched.

## Acceptance Criteria

The term "a colour-forcing host" below means a host on which the environment causes Python's argument parser to emit terminal escape sequences in help output, as `FORCE_COLOR=3` does on Python 3.14. The criteria are grounded as a set by the measurements recorded in Current Behavior above, taken on this checkout with Python 3.14.6 and `FORCE_COLOR=3` exported. Every criterion below is dischargeable on an ordinary host with no colour forced and on any supported Python version, because neither execution tier can make its own host colour-forcing: the cloud implement allowlist grants the test file as a bare leading token and a `FORCE_COLOR=3` prefix is a shape the matcher denies, and the local tier's completion gate is a continuous-integration reading on a runner that forces no colour.

- [ ] `os.environ` inside `lib/test/test_python_scripts.py` maps `PYTHON_COLORS` to the string value `"0"` and `NO_COLOR` to the string value `"1"` — string values, since every `os.environ` value is a string — whatever values the host exported for them, and an assertion in that file observes both values on its success path.
- [ ] The statement that establishes those two values stands earlier in `lib/test/test_python_scripts.py` than every one of that file's child-process invocations, the earliest of which is today the `git init` call in the repository-root resolution block, and an assertion in that file observes that source order. The placement is stated as its own criterion because the value criterion above is true of any placement above the assertion that reads it, including one that leaves earlier children inheriting the host's colour setting.
- [ ] A child process that `lib/test/test_python_scripts.py` spawns with an environment built from the file's own `os.environ` plus `FORCE_COLOR` set to the string value `"3"` returns help text containing no ANSI escape sequence, and an assertion in that file observes that. The environment is built from the file's own rather than replacing it, because the neutralisation lives in that mapping and a replacement environment would drop it.
- [ ] That forced-colour assertion reports its own non-vacuity: it records the interpreter version it ran under, and on an interpreter below 3.13 — where the argument parser colourises nothing, so the assertion would hold whatever the environment said — its recorded outcome reads as not exercised rather than as a pass. Without that report a green continuous-integration run, which pins Python 3.11, would be read as evidence of a behaviour it never exercised.
- [ ] The help text the two `#1550` probes capture contains no ANSI escape sequence, and an assertion in that file observes that.
- [ ] Running `lib/test/test_python_scripts.py` as a direct leading token reports zero failed assertions, and the assertions the change adds are among the passed. The criterion states no passed-count delta: on a colour-forcing host the 39 assertions that fail before the change move into the passed column as well, so a fixed delta would be false on exactly the host the issue exists for.
- [ ] Running `lib/test/run-shard.sh python-pool`, the shard whose membership is the pooled Python suites this file belongs to, reports zero failed assertions in its tally line, as it did before the change.
- [ ] `docs/internal/test-suite-probe-conventions.md` exists and does three things — exactly these three, complete by construction: it states that a new rendered-output probe, and an existing one a change touches, neutralises terminal colour for the child process, it names the environment variables that do so, and it names `lib/test/modules/regenerate-artifacts.sh` as a deliberate exception that tolerates the escape byte instead. The page states that it governs new and changed probes and that the rendered-output probes already in the tree are not audited by this change.
- [ ] `CONTRIBUTING.md` links to `docs/internal/test-suite-probe-conventions.md`.
- [ ] `docs/internal/test-suite-probe-conventions.md` is the repository's single statement of the rendered-output-probe colour convention, and the sentence in `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` that documents `PYTHON_COLORS=0` for the focused-test harness helper points at it.
- [ ] No file outside `lib/test/`, `docs/internal/`, and `CONTRIBUTING.md` differs after the change; the change is internal-only under the repository's versioning policy and therefore adds no `.changeset/` file.
- [ ] The required continuous-integration check named `lib + python tests` passes for the commit the change lands, which is the repository's whole-suite gate and the reading that covers the rendered-output checks in the files this change does not touch.

## Implementation Notes

**Approach**

Assign `PYTHON_COLORS` to `"0"` and `NO_COLOR` to `"1"` in `os.environ` once near the top of `lib/test/test_python_scripts.py`, above the file's first child-process invocation. Assigning into `os.environ` rather than passing `env=` at individual call sites means the neutralisation applies to the file's existing child processes and to any added later, without a second edit, and it holds however the file was invoked. Both variables are set because the repository's existing precedent sets both, and each was measured to override a forced-colour setting on its own. The change also adds four assertions, because the 39 assertions that fail today are green on continuous integration and on the pooled shard both before and after the fix, so on their own they would leave the fix unguarded against a later removal. One reads the environment the file establishes, one reads the source order of the statement that establishes it, one spawns a child with colour forced in the child's own environment, and one reads the existing probes' captured text. All four hold on an ordinary host, so an implementing run can discharge them without needing a colour-forcing machine, and the two that can only be exercised on a newer interpreter record the version they ran under so a vacuous pass is visible.

Separately, add a short page at `docs/internal/test-suite-probe-conventions.md` stating the convention, link it from `CONTRIBUTING.md`, and point the existing `PYTHON_COLORS=0` sentence in `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` at it so the tree carries one colour-policy statement rather than two.

**Relevant files**

- `lib/test/test_python_scripts.py` — the module-scope environment assignment goes here, near the existing imports and above the first `subprocess` call, and the four new assertions plausibly sit beside the `#1550` block. The block's two helper functions `_flag_help_present_1550` and `_sub_help_present_1550` are the detectors whose 39 assertions change from failing to passing; they likely need no edit themselves.
- `docs/internal/test-suite-probe-conventions.md` — the new convention page.
- `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` — the existing harness-helper sentence that mentions `PYTHON_COLORS=0`, which gains a pointer to the new page.
- `CONTRIBUTING.md` — a pointer to `docs/internal/test-suite-probe-conventions.md`, plausibly in its existing prerequisites or testing section.
- `lib/test/module-harness.sh` and `lib/test/modules/issue-audit-state.sh` — read for the existing pattern; they plausibly need no change, since both already neutralise colour.

**Code Patterns**

- Neutralise by environment variable, as `lib/test/module-harness.sh` and `lib/test/modules/issue-audit-state.sh` already do, rather than by stripping escape sequences from captured text.
- Set the variables in Python, never through a shell wrapper such as `env -u`. The project's portability convention rules out relying on tools beyond the small set the preflight guarantees, and no `env` head is granted in `prflow_implement.allowed_tools`, which grants the test file itself as a bare leading token.
- Place the assignment at module scope, matching the file's existing top-level environment handling.

**Testing Strategy**

*Test boundary.* An automated boundary exists, and the change adds four assertions at it. One reads the process environment the file establishes for its children; one reads the source order of the statement that establishes it; one spawns a child with colour forced in the child's own environment and reads the help text back; one reads the two existing probes' captured text. That is a command-line contract exercised end to end, at the level of the focused Python test file. No second test level applies: the change alters no function's return value and adds no new unit.

*Coverage walked.* The happy path is the neutralised environment, covered by the four new assertions and by the 39 existing `#1550` assertions. The degenerate input — a host that forces no colour — is covered by the same assertions, which hold there too. The adversarial input is colour forced in the child's own environment, which assertion three supplies directly rather than relying on the host. Error and failure paths do not apply: the probes already assert the child exits zero, and colourisation does not change the exit status. Concurrency and idempotency do not apply: setting an environment variable twice has the same effect as setting it once. Scale and security do not apply: neither is implied by any acceptance criterion, and the change crosses no access boundary.

*Named assertions.* Four are added and are named here; the 39 existing `#1550` assertions are named as supporting coverage.

- **Assertion one — the environment is established.** Assert that `os.environ['PYTHON_COLORS']` equals the string `"0"` and `os.environ['NO_COLOR']` equals the string `"1"`. It covers the environment-value criterion. It is RED before the change on every host and every supported Python version, because neither key is set, and it is the guard that makes a future removal of the fix visible.
- **Assertion two — the establishing statement comes first.** Read the file's own source and assert that the statement establishing those two values appears at a lower line number than every child-process invocation in the file. It covers the source-order criterion, and it is the assertion that catches a correct-looking assignment placed too late. It is RED before the change on every host and every supported Python version, because no such statement exists yet.
- **Assertion three — the neutralisation defeats forced colour.** Spawn `workpad.py --help` with an environment built from the file's own `os.environ` plus `FORCE_COLOR` set to `"3"`, and assert the captured text contains no escape byte. It covers the forced-colour-child criterion and the non-vacuity criterion. It is RED before the change on Python 3.13 and above; on earlier interpreters the argument parser colourises nothing, so the assertion records the interpreter version it ran under and reports itself not exercised rather than passed.
- **Assertion four — the probes' own text is clean.** Assert the two `#1550` probes' captured stdout contains no escape byte. On a host that forces no colour it holds vacuously, and it likewise records the interpreter version it ran under so that vacuity is visible rather than read as coverage.
- **The 39 existing assertions** named `#1550 workpad.py --help documents (help= present) the subcommand …` and `#1550 update --help documents (help= present) the invoked flag …` cover the direct-invocation and shard criteria, and they change from failing to passing on a colour-forcing host.

*Failing first, for the right reason.* This is a bug fix. Write assertions one and two first and confirm each fails against today's code — the first by reporting the two environment keys absent, the second by finding no establishing statement at all — the exact condition that lets the reported defect occur — then make it pass. Do not rely on the 39 existing assertions as the RED reading: they are green on continuous integration and on the pooled shard both before and after the change, because the runner pins Python 3.11 and the harness already exports `PYTHON_COLORS=0`.

*Manual confirmation, not an acceptance criterion.* On a host that forces colour and runs Python 3.13 and above, run `lib/test/test_python_scripts.py` before and after the change and record the two tallies. Neither execution tier can force colour on itself, so this reading is a human step recorded in the pull request, never a criterion an automated run must tick.

*Fixtures and doubles.* None. The probes and the new assertion three run the real workpad tool as a child process. The tool under probe must not be mocked, and neither must the child-process boundary; mocking one of them would remove the very rendering the assertions prove.

*The documentation half has no automated boundary.* The deliverable is prose. Verify it as a reproducible checklist, with each item tied to the two documentation criteria and the overview cross-link criterion.

1. Open `docs/internal/test-suite-probe-conventions.md` and confirm it states that a new rendered-output probe, and an existing one a change touches, neutralises terminal colour for the child process. It passes when a reader who has not seen this issue can apply the rule to a probe they are writing.
2. Confirm the page names the specific environment variables that neutralise colour. It passes when the variable names are present and spelled exactly as the code sets them.
3. Confirm the page names `lib/test/modules/regenerate-artifacts.sh` as a deliberate exception that tolerates the escape byte, and states that the probes already in the tree are not audited by this change. It passes when both sentences are present.
4. Confirm `CONTRIBUTING.md` links to `docs/internal/test-suite-probe-conventions.md` and that the link target resolves to the file that was added.
5. Confirm the `PYTHON_COLORS=0` sentence in `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` points at the new page. It passes when a reader arriving at that sentence is sent to the single convention statement rather than treating the sentence as the policy.

**Documentation Needed**

- `docs/internal/test-suite-probe-conventions.md`
- `CONTRIBUTING.md`

**Potential Gotchas**

- The environment assignment must sit above the file's first child-process invocation, not merely somewhere near the top. A child started before the assignment still inherits the host's colour setting.
- Do not "fix" `lib/test/modules/regenerate-artifacts.sh`. Its escape-tolerant pattern is deliberate and its module passes today; the decision on this issue is to document it as an exception, not to convert it.
- Do not change the two detector functions to strip escape sequences. The decided mechanism is environment neutralisation, and adding a second mechanism would leave two remedies in the file for one hazard — the same confusion this issue's documentation half exists to remove.
- Continuous integration cannot observe the original defect, because it runs an interpreter that predates help colourisation and forces no colour. Neither execution tier can force colour on itself: the cloud implement allowlist grants the test file as a bare leading token and a `FORCE_COLOR=3` prefix is a shape the matcher denies, and the local tier's completion gate is a continuous-integration reading. That is why the added assertions are written to hold on an ordinary host, and why the colour-forcing reading is a recorded human step rather than a criterion.
- This change adds no changeset file. The repository's versioning policy states verbatim that "Internal-only changes (tests, CI, dev-only docs) add none", and this change touches only the test directory, an internal documentation page, and the contributor guide.
- The file is long and its assertions are order-dependent at module scope. Inserting the assignment near the imports keeps it clear of the blocks that save and restore other environment variables.

The01Geek:
## Summary
- One test file in this repository starts the workpad command-line tool as a child process and reads the help text it prints, without telling that child to turn colour off. On a machine whose environment forces colour on, and whose Python is new enough to colour its help output, that help text comes back full of terminal escape codes and 39 assertions stop matching, even though nothing in the code under test is broken. This change makes the test file turn colour off for every child process it starts, so the same tree gives the same answer on a machine that forces colour and on one that does not.
- The fix is two assignments at the top of `lib/test/test_python_scripts.py`, placed above the file's first child process, plus new assertions in the same file that keep the fix from being removed unnoticed.
- A short page under `docs/internal/` now states the convention for tests that match against the text a command renders, so the repository carries one statement of it instead of three inline comments that gave three different remedies.

## Changes

**Focused Python test (`lib/test/test_python_scripts.py`)**: Sets `PYTHON_COLORS=0` and `NO_COLOR=1` in `os.environ` at module scope, above the file's first child-process invocation. Assigning into the mapping the children inherit — rather than passing `env=` at each call — covers the probes already in the file and any added later, and it holds however the file was invoked.

**New guarding assertions (same file)**: Thirteen assertions named `#1653` are added. They check the two values, check by parsing the file's own source that both assignments come before every child-process call in it, check that a child this file spawns really inherits both values, check that the help text the two existing `#1550` probes captured contains no escape byte, and — on an interpreter that colours help — check that a child spawned with `FORCE_COLOR=3` in its own environment still gets escape-free help.

**Convention page (`docs/internal/test-suite-probe-conventions.md`)**: New. States that a test matching against rendered command output neutralises terminal colour for the child it spawns, names the two variables that do it, gives the shell and Python forms, and names `lib/test/modules/regenerate-artifacts.sh` as a deliberate exception that tolerates the escape byte instead. It also states its own scope: the probes already in the tree were not audited here.

**Pointers to it**: `CONTRIBUTING.md` links to the new page, and the `PYTHON_COLORS=0` sentence in `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md` now points at it rather than reading as the policy itself.

**Comment correction (`lib/test/modules/issue-audit-state.sh`)**: Its comment said Python's argument parser colours help from 3.13 onward. Measurement on this host found coloured help present on 3.14 and absent on 3.12, so the comment now says 3.14. No code changed there.

## Resolves
Resolves #1653

## Test Plan
- [x] `python3 lib/test/test_python_scripts.py` run directly on this branch, on macOS with Python 3.14.6 and no colour forced: **5019 passed, 0 failed**, and 12 of the 13 `#1653` assertions ran (the thirteenth is the below-3.14 arm, which does not run here).
- [x] The same file run directly with `FORCE_COLOR=3` exported on the same host: **5019 passed, 0 failed** — the same tally as with colour not forced. This is the invocation that produces the 39 false failures on `origin/main`; the pre-fix tally was not re-measured on this branch.
- [x] The required check `lib + python tests` passed on continuous integration for the current head `00bdf3773` (run `31657812341`), with all five shards green.
- [x] `lib/test/run-shard.sh python-pool` on this local checkout reports 5020 passed and 1 failed. That one failure is the `issue-audit-state` module reporting 241 assertions against a minimum of 242 — a pre-existing floor gap on this host that reproduces identically at the merge base `278f1a755`. It is not caused by this change, and the same shard passed on continuous integration for this head.
- [ ] Reviewer check: confirm the two assignments in `lib/test/test_python_scripts.py` still sit above the file's first child process — the ordering assertion parses the file's source rather than a hardcoded line number, so it should catch a later move on its own.

### What the new assertions do and do not cover
- The one assertion that proves the mechanism on **every** supported interpreter is the inheritance probe: it starts a child of this file and reads both values back out of that child's environment. It is the only executable proof that runs on the Python 3.11 continuous integration pins, and deleting either assignment turns it red there.
- The four assertions that spawn a child with `FORCE_COLOR=3` are gated at `sys.version_info >= (3, 14)`. Below that version an else-arm asserts the interpreter really is below 3.14 and labels its own outcome as **not exercised**, so a green run on Python 3.11 is never read as proof of a behaviour it never ran. The gate sits at 3.14 because coloured help was measured present on 3.14 and absent on 3.12 on this host; 3.13 could not be measured here.
- The 39 pre-existing `#1550` assertions are the failures this change removes, but they are green on continuous integration and on the pooled shard both before and after the fix, so they guard nothing. That is why the `#1653` assertions were added.
- Nothing here audits the rendered-output probes that already exist elsewhere in the tree, and no mechanical check enforces the new convention. Deferred to a follow-up, with no issue filed for either yet: neutralising the two known un-neutralised sibling probes, `lib/test/test_module_runner.py` and `lib/test/test_render_audit_prompt.py`, and adding a mechanical backstop for the convention. The convention page states this scope limit itself.

## Visual Changes
N/A

## Breaking Changes
None. The change is confined to the test directory, two internal documentation pages, and the contributor guide. No shipped surface, workflow, or configuration key is touched, and no `.changeset/` file is added because the change is internal-only under the repository's versioning policy.

### Prompt-surface size

prompt-surface growth: no tracked `*.md` under `skills/`, `agents/`, or `.prflow/prompt-extensions/` changed between `278f1a755b3a2c941a56b5677429d589f372cdde` and `00bdf37732f6ee58a7efcb9e31b2c1add8546180` — no table rendered.


The01Geek:
# PRFlow Review — PR #1660



**Status:** 🚀 Reviewing
**Diff profile:** engine_self_modifying + detect_all_audit → full checklist + forced completeness-critic pass
**Run:** _(local run)_
**Last updated:** (seeding)

## Blueprint
- [ ] Review agents (Phase 3)
- [ ] Run complete — everything this run owed

## Findings (live)
_(Phase-3 findings appear here as each agent returns.)_

## Verdict
_(pending)_

PR #1681, opened 2026-08-14
feat: add quality-first create-issue evaluation

The01Geek:
## Summary

- extract the legacy create-issue context analyzer into an importable module while preserving the existing CLI and raw JSON shapes
- add schema-versioned, run-addressable manifests with occurrence boundaries, per-run state/checkpoint artifacts, provenance controls, and fail-closed paired comparisons
- add deterministic draft/checkpoint metrics, complete audit-state validation, semantic audit outcomes, rubric grading, and a quality-first comparison gate
- add a provider-neutral controlled A/B runner, statistical reports, and deterministic anonymized review export
- reject malformed, inexact, incomplete, path-escaping, or mixed-provenance evidence instead of converting unknowns into numeric savings or passes
- document the observational and controlled evaluation workflows, provenance rules, privacy boundary, variance, and quality-first interpretation

## Verification

Before the request to move remaining verification to CI:

- `python3 lib/test/test_create_issue_context_eval.py` — 129 passed
- `python3 lib/test/test_python_scripts.py` — 5106 passed
- directly affected recorder and coverage-map suites passed during the manifest foundation work
- Ruff and Python compilation passed during the manifest foundation work

Later hardening, runner, and documentation commits were intentionally not rerun locally. CI is the authoritative final gate.

## Compatibility and privacy

- legacy positional and `--before`/`--after` modes retain their serialized shapes
- raw mode does not emit transcript or issue-body text
- issue draft bodies are read only from explicit manifest checkpoint artifacts
- artifact paths are realpath-contained beneath the declared manifest root


The01Geek:
**Finding rejected:** the Important/Major finding on head `823973ea` — "The per-configuration `mixed_provenance` guard in `_manifest_comparison` (`scripts/create_issue_eval.py:1985-1989`) … has **no falsifying test**" — is stale. The falsifying test and its positive control were added in an earlier fix cycle on this branch and are present at the reviewed head:

- `lib/test/test_create_issue_context_eval.py:1852` — `test_mixed_provenance_within_one_configuration_is_unestablished`, which builds a 4-run manifest (2 repetitions x 2 configurations), drifts `skill_fingerprint` on one repetition-2 run and `repo_sha` on both, and asserts `comparison["diagnostic"] == "mixed_provenance"` (line 1870) plus the fully-unestablished delta and `len(pairs) == 2` (so the rejection is attributed to the per-configuration guard, not the earlier pairwise one).
- `lib/test/test_create_issue_context_eval.py:1845` — `test_repetitions_of_one_configuration_share_provenance`, the same-provenance positive control asserting the same 4-run manifest is otherwise `established`.

Verified by reading the file at the reviewed commit `823973ea41f275a3096db791f4c1337ff258d957`. The finding describes the branch as it stood at `1475dc4b`, before that cycle landed. No further fix pass is warranted; a later re-review re-raising this same finding on the same evidence does not reopen it.

PR #1728, opened 2026-08-17
fix: rewrite cloud-denied fence shapes in shipped skills

### LINKED ISSUE #1721: Shipped fences prescribe a redirect shape the cloud matcher denies, and one makes its own diagnostic channel unreachable
## Problem Statement

Shipped fences instruct a run to redirect into `.prflow/tmp/**`. On the cloud tier the harness refuses that shape, so the fence never executes: it returns no output, burns one request, and the agent recovers by improvising a different form on the next request.

Measured on a cloud `/prflow:implement` run (adopter repo `Radman-LLC/ADR`, GH run `31989737682`, 2026-08-17). Three prescribed redirects fired; **all three were denied**. The run recovered each time, so nothing failed loudly — which is why this has not surfaced before.

The cost is small and the correctness harm is not:

- The improvised fallback for the Phase 0 local-diff path has **none of the fail-closed staging** the prescribed 15 lines specify.
- `skills/review/phases/phase-0-setup.md:271` instructs the run, on failure, to *"read `.prflow/tmp/review/<slug>/<run-id>/acs.err` to quote the cause"* — **a file the blocked redirect can never create**. The diagnostic channel is dead by construction on this tier.

## Current Behavior

The three denials, from the run's transcript:

| fence | prescribed shape | harness response |
|---|---|---|
| `skills/review/phases/phase-0-setup.md:151-153` | `git diff … > …/diff.raw-candidate` then `awk … >` then `sed -n 'p' … >` (the fail-closed staging path) | `"The following parts require approval: git diff …, awk …, sed -n 'p' …"` |
| `skills/review/phases/phase-0-setup.md:262` | `workpad.py acs-resolve … 2>…/acs.err ; echo "acs-rc=$?"` | `"Output redirection to '…/acs.err' was blocked. For security, Claude Code may only write to files in the allowed working directories: '/home/runner/work/ADR/ADR'"` |
| `skills/implement/phases/phase-1-setup.md:251` | `printf '%s\n' "$CLAUDE_CODE_SESSION_ID" > …/implement-active-<n>` | `"Contains simple_expansion"` |

Note the second message: **the refused path is inside the allowed root it names.** The guard is on the redirect construct, not the destination — so "write it under `.prflow/tmp/`" is not a remedy on its own.

Recovery in each case was immediate and correct, which is the evidence that the *intent* is sound and only the *form* is wrong:

- `acs-resolve` re-issued without the redirect returned the criteria normally.
- The diff path re-issued as `git diff … | awk … | tee …/diff.patch | wc -l`, with the agent noting *"Redirects denied on this tier. Using the recorded `tee` form instead."* — a form that already appears as a comment at `phase-0-setup.md:146` beside the `gh pr diff` path.
- The session marker was re-issued via the Write tool.

CLAUDE.md already documents the class (*"the matcher also denies composite shapes … a `/tmp` redirect, a heredoc write"*) and already names the permitted alternatives (a Write into `.prflow/tmp/**`, `… | tee`). The fences were simply not written to it.

## Population

`grep -rcE '>\s*\.prflow/tmp' skills/` → **27 occurrences across 8 files**:

- `2>` form (6 files): `review/SKILL.md` (4), `review/phases/phase-0-setup.md` (2), `review/phases/phase-4-verdict.md` (1), `review-and-fix/references/loop-control.md` (1), `review-and-fix/references/loop-exit.md` (2), `implement/references/deferred-review-findings.md` (2)
- plain `>` form (15 occurrences), the majority in `retrospective-weekly/SKILL.md`

Not all are cloud-reachable — `retrospective-weekly` runs on the local/interactive tier, where the shape is not refused. **The audit must be scoped to fences a cloud tier can execute**, not to the raw grep population.

## Desired Behavior

Every fence a cloud tier can reach prescribes a shape that tier accepts, and no fence's error-handling arm depends on an artifact that shape cannot produce.

## User Impact

A consumer's cloud run stops silently discarding a request per denied fence, and — the load-bearing half — stops falling back from a fail-closed staging path to one with no staging, and stops being told to diagnose failures by reading a file that cannot exist.

## Scope boundary

- This is about **prescribed command shapes in shipped fences**, not about widening the allowlist. Do not add grants.
- It does not revisit which paths are writable; the Write tool and `| tee` already work.
- It does not cover the local/interactive tier, where these redirects execute normally — a fix must not degrade that tier to work around a cloud constraint.

## Technical Context

- The permitted alternatives are already recorded in CLAUDE.md's cloud-allowlist gotcha and demonstrated in-run: `| tee`, and the Write tool for whole-file artifacts.
- `phase-0-setup.md:146` already carries the `tee` form as a comment on the adjacent path — the knowledge is present, just not on the prescribed arm.
- Two shipped fences deliberately prescribe the **Write tool** precisely because the redirect is denied (`phase-3-fix-loop.md:33` *"Do not emit a shell redirect for this snapshot"*, `phase-0-setup.md:19`). Those are the pattern to follow.
- Whether a given fence is cloud-reachable is decided by which tier executes its phase; `lib/capability-profiles.json` and the per-command mode table in `docs/internal/cloud-allowlist.md` are the inputs.

## Acceptance Criteria

- [ ] AC1 — Every redirect-into-`.prflow/tmp` occurrence under `skills/` is adjudicated as cloud-reachable or local-only, and the adjudication is recorded. Verify: `grep -rnE '>\s*\.prflow/tmp' skills/` and a per-occurrence disposition.
- [ ] AC2 — Every cloud-reachable occurrence is rewritten to a shape that tier accepts (`| tee`, or the Write tool), preserving the fence's existing fail-closed semantics — in particular the Phase 0 staging path keeps its guard structure rather than collapsing to a single pipeline.
- [ ] AC3 — No error-handling arm instructs a run to read an artifact that the rewritten fence does not produce. `phase-0-setup.md:271`'s `acs.err` read is corrected or removed in the same change.
- [ ] AC4 — Local-tier-only fences are left executing as they do today; the change does not degrade the retrospective path to satisfy a cloud constraint.
- [ ] AC5 — The three shapes named under **Current Behavior** are each confirmed to execute on the cloud tier after the change, or the residual is recorded with the reason.

## Implementation Notes

A map, not a plan.

- The `acs-resolve` fence is the cheapest: strip the redirect and keep the `acs-rc` token, which is what actually made the outcome observable.
- The Phase 0 staging path is the delicate one — its three guarded stages exist so a partial diff cannot be mistaken for a whole one. A single `| tee` pipeline is what the agent improvised, and it is *not* an equivalent; preserving the staging likely means the Write tool or a helper rather than a one-liner.
- Prompt-surface edits route through a context-isolated writing-skills subagent with a `Writing-skills evidence:` marker per return. Add a `.changeset/*.md` carrying `bump: patch`.
- Consumer-repo constraint: these bodies ship verbatim, so no `lib/test/run.sh`, no `docs/internal/…`, and no citation of this repo's `CLAUDE.md`.

## Related

- #1047 — the PreToolUse shape guard is wired to no runnable tier. Adjacent (both concern denied shapes) but distinct: that issue is about a guard's wiring, this one about what shipped fences prescribe.
- Measurement source and the full request-level cost breakdown: `docs/internal/agent-context-economics.private.md` §9.6.

The01Geek:
Resolves #1721

Shipped agent-prompt fences prescribed command shapes the cloud harness silently refuses. A refused fence produces **no output at all** and burns the request, so a run recovers by improvising — and for the Phase 0 diff-staging path that improvisation dropped the fail-closed staging entirely.

Three distinct refusal causes were measured on a real cloud run, and the fix is per-cause, not a single sweep:

| Measured refusal | Cause |
|---|---|
| a compound staging chain (`git diff …`, `awk …`, `sed -n 'p' …`) | **multi-head** statement |
| `2>…/acs.err` — refused even for a target *inside* the allowed root | the **redirect construct** |
| `printf '%s\n' "$CLAUDE_CODE_SESSION_ID" > …` | a **variable expansion**, which survives removing the redirect |

## What changed

**The Phase 0 local-diff staging path** is now six ordered steps rather than one chain. A bare `git diff --name-status` producer carries no pipeline, so its exit status is genuinely `git`'s; the diff then streams through `tee` so its bytes never transit the orchestrator. Three equations guard it: the raw count against the producer's rows **plus its `T` rows** (a typechange is one row but two `diff --git` sections), the published count against raw minus logs, and the count of the **published file** against the streamed count — because `tee` keeps copying to stdout when its write fails, so a stream count alone admits a truncated cache. Any failure clears the cache and stops.

**Stderr is read from each invocation's own tool result** instead of being captured to a `.err` file, across the review root, three review phases, both fix-loop references, and the implement deferrals reference.

**One error-handling arm was dead, not merely awkward.** The acceptance-criteria resolver's failure arm told the run to read a file only the refused redirect could have created, so on any tier that refuses the redirect the diagnostic channel could never produce a cause. It now quotes the stderr the invocation itself returned.

**Two fences that routed on grepping a captured stderr file** now guard on each invocation's own exit status, read inline — never captured into a variable a later statement reads, which a runner that strips cross-statement variables leaves empty, routing every healthy run to the unrecognised arm. Where the status is genuinely ambiguous the run reads that call's stderr to classify it, and records an unrecognised shape as a failure rather than guessing.

**Deliberately unchanged:** the weekly retrospective skill, which no workflow dispatches and which therefore runs only on the interactive tier, where these redirects execute normally. It was not degraded to satisfy a cloud constraint.

## Scope carried to a follow-up

Three cloud-reachable occurrences are adjudicated but **not** rewritten, because this issue's remedy cannot reach them — the Write tool cannot participate in a shell read-loop, and cannot source a command's stdout. They are recorded with their reasons in `docs/internal/cloud-allowlist.md` and carried to #1734: `phase-3-agents.md`'s appends and captures, `phase-3-fix-loop.md`'s `mktemp` targets (the already-known probe-denied `/tmp` class), and the deferrals reference's jq-merge write to `"${AGG}.tmp"`.

No allowlist grant is added by this change.

## Verification

The rewritten fences are **executed**, not asserted about: `lib/test/run.sh` extracts the producer, the logs count and the `awk` filter from the shipped phase file between marker pairs and runs them against real fixtures — an unresolved base (fails non-zero and stages nothing), a section renamed *into* `.prflow/logs/` (the anchored-pattern case), and a real typechange commit. Each was mutation-checked: the shipped text passes and the regression it guards against fails.

**Known residual, recorded rather than papered over:** none of the three measured shapes is confirmed by a post-change cloud run — the probe workflow is the only channel that measures the matcher, and no run was made against this branch. `docs/internal/cloud-allowlist.md` records that residual and what each remedy rests on instead.

Five pin retirements this change makes necessary are authorized through a delta manifest, in the same shape as the issue-1633 fence migration.

Generated via /prflow:implement (v2.32.76, claude-opus-5, medium)

Writing-skills evidence: skills/review/phases/phase-0-setup.md — skill-loaded=yes (superpowers:writing-skills invoked before the edit); guidance-applied=yes (minimal correction addressing the observed failure, no added prose); pressure-scenario=no (the failure is a directly verifiable false completeness claim contradicted by the same file, not a discipline failure needing a baseline run); micro-tests=no (a factual enumeration fix, not behaviour-shaping wording).

Writing-skills evidence: skills/review/SKILL.md, skills/review/phases/phase-1-checklist.md, skills/review/phases/phase-4-verdict.md, skills/review-and-fix/references/loop-control.md, skills/review-and-fix/references/loop-exit.md — skill-loaded=yes (superpowers:writing-skills invoked earlier in this session, before the first prompt-surface edit); guidance-applied=yes (each edit rewords an existing sentence rather than appending, per the extension rule, and the review/SKILL.md edit was trimmed to hold the file under its reader-cap ceiling); pressure-scenario=no (these are factual corrections to false claims and a structurally unrenderable template slot, not discipline guidance needing a baseline run); micro-tests=no (no behaviour-shaping wording introduced; the one behavioural change is guarded by a new executable assertion in lib/test/run.sh).


The01Geek:
## Reception pass — deferral record (PR #1728)

Findings 1, 3 and both Suggestions were fixed in 520ae0f2a. The two below were not, each with its reason and its revisit condition.

**Finding 2 — the widened `ok || degraded` filing guard and the "no deferrals were filed" reflection.**
Not deferred: **already resolved at HEAD** before this pass ran. The reviewed head (fca97b19) predates commit f15045faa, which split that arm in `skills/implement/references/deferred-review-findings.md` into two texts selected by reading the sentinel's `filing=` field — `filing=` anything other than filed keeps the "no deferrals were filed this run" wording, and `filing=[filed]` records that a previously persisted aggregate was filed from and must not be re-triggered. Verified against HEAD; the contradiction the finding describes cannot be written any more.

**Finding 4 — the removed SP503 per-stage stale-cache-removal coverage.**
WHAT: not restored. The finding is factually correct — the old test executed the whole multi-stage shell fence under a faked PATH and asserted per failure stage that the stale cache and both candidates were removed; no equivalent executable test exists now.
WHY: the behaviour it covered no longer exists as an executable artifact. The Phase 0.2 local-diff path is no longer one runnable fence — it is six individually emitted steps plus three equations the agent evaluates, and the `rm -f` on failure is one clause of that agent-evaluated routing. Restoring per-stage coverage would mean re-introducing the single shell fence this issue exists to remove. This is the documented no-automated-coverage class for agent-executed prompt prose whose only reader is the runtime agent; the compensating control is the review pass, and the gap is narrowed rather than closed. What *is* executable was kept and extended this pass: the awk filter, the logs count, the producer's status and shape, the typechange equation, and — added here — the step-6 published-file count, whose three new assertions were mutation-checked (a step 6 re-pointed at the raw candidate prints 3 against the pinned 2).
REVISIT IF: the staging path is ever consolidated back into a single shell-executable fence, or a helper takes ownership of the staging so the removal logic has a testable surface again.

The01Geek:
# PRFlow Review — PR #1728



**Status:** ✅ Converged (fixes applied)
**Diff profile:** detect_all_audit + engine_self_modifying (full checklist + forced completeness critic)
**Run:** [View run](_(local run)_)
**Reviewed HEAD:** 6be63947cd68d545892bc306b3f5d2fe96c81b74 → fixes landed at 861bdd5d9
**Last updated:** Loop Exit

## Blueprint
- [x] Generate verification checklist (Phase 1) — 52 items (2 batches, deduped from 53)
- [x] Verify checklist (Phase 2) — 49 passed, 3 failed, 0 inconclusive (1 normalized wording-only)
- [x] Review agents (Phase 3) — 5 dispatched, all returned; `type-design-analyzer` gated out (`has_new_types` false)
- [x] Aggregate & verdict (Phase 4)
- [x] Run complete — everything this run owed

## Findings fixed this iteration

**Truthfulness defects (self-contradicting-diff class)**
- `docs/internal/cloud-allowlist.md` — "none of the three refusals **involves** a stdout redirect into `.prflow/tmp/**`" was false: row 1's pre-change staging chain carried three (`git diff … > …/diff.raw-candidate`, `awk … > …/diff.candidate`, `sed -n 'p' … > …/diff.patch`). Restated as the causal claim the evidence supports. *(The concurrent cloud review reached the same conclusion independently.)*
- `skills/review/SKILL.md`, `phase-1-checklist.md`, `phase-4-verdict.md` — "a later probe run did not reproduce" asserted a run that does not exist; the same PR records that no post-change cloud run was made.
- `lib/test/extract-command-shapes.py` — the STDERR arm's "flagged for ANY head" is contradicted by its own no-head early return, and probe row 19 was cited as supporting a gh exemption while the shipped fixture flags exactly row 19's form. Both restated; the arm's residuals (`<scratch-dir>/`, variable targets, multi-digit fds) now disclosed.

**Diagnostics the rewrite made unemittable** — raised independently by 3 of 5 review agents
- `loop-control.md`, `loop-exit.md`, `review/SKILL.md`, `phase-4-verdict.md` — the `<stderr-quote>` slot sat inside the fence whose stderr it quotes, so it could never be rendered: the `::warning::` would have carried a literal placeholder. The in-fence warning is now generic and the cause is emitted as a separate second warning after the tool result is read.

**Completeness-critic pass (forced by `detect_all_audit`)**
- Independent signal: tokenized every fenced block under `skills/` with the lint's own `_REDIR` + `_tokenize`, resolving each redirect target — structurally different from the audit's line-oriented grep.
- Uncovered member found: `skills/create-issue/references/issue-template.md:213`, whose target `"<main-root>/.prflow/tmp/…"` is placeholder-prefixed and so invisible to a pattern anchored at `.prflow/tmp`. Row added; the placeholder class documented beside the existing variable-target caveat.

**Guard that did not guard its own surface**
- `lib/test/run.sh` — IR6 is excluded from the command tier and the review arm table has no workspace-scratch arm, so **no profile scanned the review bundle** for the fd-2 capture this issue removed from it. Added a direct assertion over that bundle plus a non-vacuity control.

**Stale comments** — `run.sh` and `discover-deferral-manifests.py` both still described the retired `discovery=[failed]` arm.

**From the concurrent cloud review** — its `tee` wording inconsistency, and its field evidence that a stdout redirect into `.prflow/tmp/**` was observed refused on the current action version (recorded as superseding this section's optimism, hedged as one observation rather than a probe row).

## Pushed back (not applied)
- The AC1 table's `2 × 2>` count for `phase-3-fix-loop.md` was reported false; it is correct — lines 113 (`2>`) and 116 (`2>>`) both target `$PERSIST_ERR`.

## Deferred with recorded rationale
- The Phase 4.0.5 filing-guard widening (`{ok,partial}` → `{ok,degraded}`) — a deliberate, documented change; `file-deferrals.py`'s already-filed refusal bounds it. Raised for maintainer confirmation, not reverted mid-loop.
- Phase 0.2 publishing `diff.patch` before validating it — disclosed tradeoff central to this issue's premise.
- Additive fixture gaps (no extraction marker on steps 2a/2b/3; the `2>&1` row's discrimination; no space-separated/quoted stderr-target fixture).

## Verdict
**Converged after 1 iteration — fixes applied and pushed.** No Critical finding survived; every truthfulness defect found is corrected. Desk lints green (artifact regeneration, reference-size, ruff, shellcheck, stale-prose self-scan 0 STALE). The whole-suite gate is the CI reading for `861bdd5d9`.

Issue #919, opened 2026-07-29
Record the PreToolUse guard's post-merge probe evidence: dispatch the probe arm and fill the cloud-allowlist table (maintainer-only, split from #908)


The01Geek:
> **Maintainer-only, post-merge.** This is deliberately not implementable by an autonomous `/devflow:implement` run — see "Why this is separate" below. Split out of #908.

> **Post-#937 scope correction (2026-07-29).** PR #937 (issue #936) **withheld the automatic pull-request-triggered review tier**: `.github/workflows/devflow-review.yml` is deleted from `main`, `.github/workflows/devflow-runner.yml` is retained in the tree but has **no caller** (its only entry point was `devflow-review.yml`'s `review` job), and `install.sh` ships none of the three tier files. Consequence for this issue: the `matcher-probe.yml` half is unaffected and still measurable, but the *real-review-run* half is not — **AC3 is dropped** (recorded below under "Dropped acceptance criterion"), and reason 2 under "Why this is separate" is corrected. Everything else stands.

## Problem Statement

#908 wires the PreToolUse shape guard to fire: the `settings:` input on the review tier, the heartbeat and per-arm-count job outputs, the denied-command publication/rendering, and the `PreToolUse` probe arm in `.github/workflows/matcher-probe.yml`. All of that is *authoring* work and is dischargeable in-PR.

What #908 cannot do is **observe** the result. `docs/cloud-allowlist.md` carries a placeholder probe-evidence row (`_(pending post-merge dispatch)_`, added in #906) that stays empty until someone dispatches the probe and reads a real review run. This issue tracks that measurement.

## Background
Stripped down: #919 buys you **evidence that a safety net actually works**, not a new feature.

The background: on the cloud tiers, a command shape the matcher doesn't grant is *silently* refused — no error, no output. That's how you get a run that burns budget and ends with no verdict. #908 built a `PreToolUse` hook to catch that: intercept the Bash call, and say out loud "this shape would be denied, here's why."

But whether a hook registered through the action's `settings:` input actually fires in the real cloud action — and whether its reason string survives back to the transcript — cannot be established at the desk. Per the repo's own rule, matcher and hook semantics are only provable by dispatching `matcher-probe.yml` and reading a real run. #908 could *author* the guard; it couldn't *observe* it.

So #919's deliverable is one dispatch and a filled-in table row in `docs/cloud-allowlist.md`: fired or didn't, reason delivered or not. Three possible payoffs:

[...]

Issue #805, opened 2026-07-25
Stop denied command shapes from being re-emitted in cloud runs


The01Geek:
## Problem Statement

A DevFlow cloud run that emits a command in a shape the harness matcher denies gets a terse refusal naming the offending sub-part and nothing else, so the engine re-emits variants of the same denied shape instead of switching. In review run 30138268273 (PR #786) the engine recorded 18 permission denials; five of them were the same `/tmp` output redirect, emitted at transcript events 313, 319, 322, 390, and 541 — the last one 228 events after the first. Three more were the same `bash <path>` wrapper.

The engine already receives guidance against exactly this. The run's `Compose review prompt` step rendered, verbatim:

> `- **Denied — do not emit:** a leading VAR=value assignment or env-prefix … a >/>> redirect (or any other authoring) targeting /tmp …`
> `- **Hard rule: after two denials of a shape, switch to a permitted alternative above — never iterate variants of the denied shape.**`

The harness's own denial text named the remedy too: *"may only write to files in the allowed working directories for this session: '/home/runner/work/devflow-autopilot/devflow-autopilot'"*.

Guidance was present in the prompt and in the denial message, and the shape was still re-emitted five times. Adding a fourth copy of advisory prose to the prompt is the mechanism that has already been measured failing.

**The load-bearing assumption this issue rests on, stated with its falsifier.** A `PreToolUse` deny reaches the engine through `permissionDecisionReason`, which the cited documentation describes as "shown to Claude in the tool result as the reason the call was blocked" — the *same tool-result channel* the harness's own refusal already used. The assumption is that a per-call deny carrying a **specific permitted alternative for the matched arm**, delivered at the moment of the offending call, changes behavior where a generic refusal did not. It is **falsified if denied-shape re-emission persists at the run-30138268273 rate**, which is why this issue ships an outcome measurement rather than declaring the problem solved on the mechanism's existence. The supporting evidence is one observation, not a measured effect size: a probe on Claude Code CLI 2.1.220 in which a remediating hook deny cut repeats of one shape from five to two before the engine abandoned the shape — assumption, confirm before implementing.

Because that assumption may fail, the guard also carries a **repeat bound**, which is a control rather than a message: a second denial of the same arm within one run escalates the remediation to name the abandonment rule explicitly, and the run's escalation count is published, so a persistent re-emitter is visible in the check rather than only in an artifact.

[...]

PR #1965, opened 2026-08-25
Record the PreToolUse guard as retained-but-inert, and fail open when its interpreter is absent (#1047)

### LINKED ISSUE #1047: The PreToolUse shape guard is wired to no runnable tier, and its R1 remediation alternative has no recorded verdict
> **Framing correction.** This was first triaged as "register the guard on the implement tier — implement lacks what review has". That framing is **wrong** and is not what this issue asks for. The guard is inert on **every** tier: its sole registration point is a workflow with no caller. "Implement lacks what review has" is false, because review does not have it either — not in any form that can run.

## Problem Statement

`scripts/pretooluse-shape-guard.py` is a `PreToolUse`/`Bash` hook that intercepts a command whose any statement matches a probe-proven denied shape and returns a `deny` whose `permissionDecisionReason` names the permitted alternative, with an escalation suffix on a second denial of the same arm. It was built (#805, PR #906) as the compensating control for a measured failure: review run 30138268273 recorded five `/tmp`-redirect denials of the same shape, the last 228 transcript events after the first, with advisory prose against exactly that already in the prompt.

#908 wired its registration — a `settings:` input on `.github/workflows/devflow-runner.yml`'s review-tier action step. #937 (issue #936) then withheld the automatic pull-request-triggered review tier and deleted `.github/workflows/devflow-review.yml`, which was `devflow-runner.yml`'s only caller.

The result is a guard that is **wired to no runnable tier**. The body, its unit coverage, and its `#458` trusted-source hardening are all in the tree and green; the registration exists; and nothing can execute it. #919 named this as a newly surfaced consequence and explicitly deferred the wiring decision to a separate issue. This is that issue.

## Current Behavior — verified

Commands run at `4686555b` (`main`), output read.

- `Verified:` `grep -rn "settings:" .github/workflows/` returns exactly two real `settings:` inputs — `devflow-runner.yml:2406` (the guard; rationale comment at `:2358-2404`) and `matcher-probe.yml:1238` (an **ad-hoc probe hook** that touches a breadcrumb and returns `permissionDecision: allow`; it is not the guard). The other two hits are comment lines inside `devflow-runner.yml`.
- `Verified:` `grep -c "settings:" .github/workflows/devflow-implement.yml .github/workflows/devflow.yml` → `0` and `0`. Neither live tier registers any hook.
- `Verified:` `devflow-runner.yml` declares `on:` / `workflow_call:` only (lines 21-22).
- `Verified:` `grep -rn "uses: ./.github/workflows" .github/workflows/` exits 1 with zero output — no workflow in the tree calls any local reusable workflow, so `devflow-runner.yml` has no caller.

So the guard cannot fire in any run this repository can start, and — because `install.sh` ships none of the three withheld tier files — cannot fire in a consumer either, except in a consumer that installed the tier before #937 and still carries its own `devflow-review.yml`. That population keeps the guard live and keeps its exposure to #930/#920, per the retention rule in `CLAUDE.md`.

### What the implement tier actually gets today

- `Verified:` `.github/workflows/devflow-implement.yml:1398`, step `Surface execution diagnostics`, `if: always()`, gated on `.prflow.execution_diagnostics_enabled` (only the exact string `false` disables it; an empty read falls safe to enabled). It calls the vendored `surface-execution-diagnostics.sh`, whose `_publish_denials()` emits a job-log `::warning::` and publishes `permission_denials_count` — or the literal `unavailable`, never collapsed onto `0`.
- That step runs **after** `claude-code-action` returns. Nothing reaches the model while it is re-emitting a denied shape. It is a post-mortem, not a control.
- `Verified:` `grep -c permission_denials .github/workflows/devflow-implement.yml` → `0`. The count is produced by the helper, not referenced by this workflow.
- `Verified:` `grep -rn "describe-denial-count" .github/workflows/ scripts/ lib/` returns only prose references (each naming it as the *precedent* for the inline-shell-extraction convention). `scripts/describe-denial-count.sh`'s three-way clause selection — the repository's own reference implementation of "unknown is not zero" — is **dead code on every tier**: its caller was the deleted `devflow-review.yml`'s `finalize_check`.

## Desired Behavior

A decision, recorded, on whether the shape guard becomes live on a tier that runs — and if yes, a guard whose deny set is provenance-correct for the tier it polices.

Two options, both legitimate; this issue asks for one to be chosen and the losing one closed out explicitly rather than left implicit:

- **(a) Wire it onto a live tier** — `devflow-implement.yml` (the tier that actually spends budget on long autonomous runs) and/or `devflow.yml` (the manual review path that survived #936).
- **(b) Accept it as retained-but-inert** alongside the rest of the withheld tier, and say so in `docs/cloud-allowlist.md` and in the guard's own docstring, so a future reader does not mistake a registered-but-uncallable hook for a live control.

## The #805 exclusion rationale, quoted — and why it has strengthened

#805's acceptance criterion said, verbatim:

> The guard is registered on the **review tier only**. `.github/workflows/devflow-implement.yml` registers no guard in this change, because the implement tier's denied shapes are keyed to the separate `IMPLEMENT_RULES` set (`IR1`, `IR2`, `IR3`) and a separate probe baseline, and applying the review deny set there would police that tier by the wrong rules while leaving its own shapes uncovered.

and its scope note added: *"covering it is separate work with its own deny-set provenance to establish."*

That is an explicit **"not yet"**, not a "deliberately never" — which is what makes this issue admissible at all. But every clause of it has gotten **more** true since, so **this issue must not propose reusing the review deny set on the implement tier**:

- `Verified:` `IMPLEMENT_RULES` has grown from three members to five — `lib/test/extract-command-shapes.py:554` reads `IMPLEMENT_RULES = frozenset({"IR1", "IR2", "IR3", "IR4", "IR5"})`.
- `IR4` (a leading `cd`) is per `CLAUDE.md` and `docs/working-directory-contract.md` an **authoring** rule, explicitly *not* a claimed matcher refusal — a leading `cd` was observed **executing** on the review tier (run 30222310785). A runtime deny is terminal, so denying `IR4` at runtime would cost a shape the harness permits. It is exactly the `R2` exclusion, one tier over.
- `IR5` mirrors R3's `/tmp` arm but rests on a *proven-permitted alternative*, not on a measured denial of the arm itself on this tier. `docs/cloud-allowlist.md` states outright that `R1`, `R3` and `R4` are **not enforced on the implement profile at all, because their status there is unmeasured**.
- `Verified:` there is **no `_IMPLEMENT_ARM_TABLE`**. `lib/test/extract-command-shapes.py:544` defines `_REVIEW_ARM_TABLE`; `classify_arms()` at `:566` iterates that table and nothing else. `find_implement_violations()` at `:1103` walks `_fence_line_offsets(text)` — it is a **markdown-fence scanner over a whole file**, not a pure statement classifier a hook could call on a single Bash payload.

So option (a) for the implement tier is not a parameter change. It needs a new implement-side arm table with its own per-arm provenance, and a new pure statement-level classifier alongside the existing fence scanner. Deny-eligibility must be decided **per arm**, and an arm whose only backing is an authoring rule or an inference is deferred, never denied — the same discipline that excluded `R2` and `R3-heredoc` from the review deny set.

## Unestablished — do not treat any of these as settled

- ~~Whether the hook fires at all~~ **SETTLED by #919 (closed 2026-08-05T03:37:07Z).** It fires through `claude-code-action`'s `settings:` input: probe run `30956039324` (job `92149438739`) recorded **FIRED**, replicated identically across 8 consecutive observations, and independently corroborated by the deny/defer arms PR #1308 added — `pretooluse-deny-probe` (run `30967680822` / job `92185120507`, `FIRED-AND-DENIED`) and `defer-probe` (job `92185120496`, `DEFER-BLOCKED`). See *Harness hook-surface probe evidence (Part 2)* in `docs/internal/cloud-allowlist.md`. No longer unestablished.
- ~~Whether its `permissionDecisionReason` reaches the engine transcript~~ **SETTLED by #919, via Part 2.** The allow-decision arm (run `30956039324`) recorded `REASON-ABSENT`, but that is the *specified* behavior for an `allow` decision — the hooks reference shows the reason is delivered on `deny`/`ask` and ignored on `allow`/`defer` — not evidence the delivery path is broken. The `pretooluse-deny-probe` arm, which emits a real `deny` (the decision the shape guard itself emits), recorded **`REASON-DELIVERED`**. That is the decision the shape guard's `permissionDecisionReason` payload depends on, and it is confirmed delivered, not merely unmeasured.
- **Whether a hook `command` bypasses the tool-grant matcher.** Strongly implied — the review profile grants no `python3` head, yet the review guard's registered hook command is `... exec python3 "$_g"` — but never measured. If it does *not* bypass, option (a) on any tier is dead on arrival and needs a different delivery form.
- **Whether `settings:` MERGES with or OVERWRITES the base-restored `.claude/settings.json` hooks.** Flagged as an undocumented upstream caveat in `devflow-runner.yml:2394-2404`, with a live consequence already named there: under overwrite semantics the restored settings' three `Stop` hooks (`lib/efficiency-trace.sh --persist`, `lib/implement-stop-guard.sh`, `scripts/stop-hook-probe.sh`) plus its `enabledPlugins` block would silently stop applying. On `devflow-implement.yml` those `Stop` hooks are load-bearing, so this caveat is a **blocking** unknown for option (a) there, not a footnote.
- **Which implement arms would be deny-eligible.** Establishing that is the bulk of the work, and cannot be done from the current evidence: no implement-tier probe row exercises a leading assignment, a `/tmp` redirect, or an interpreter head.

## R1's remediation names an alternative that carries no recorded verdict — a second precondition of option (a)

The section above is a **provenance** problem about the *tier* a `REMEDIATION` message names. This is a different, sharper problem: R1's recommended *alternative* is itself asserted with no provenance at all, and two in-tree statements assert the opposite.

`Verified:` `scripts/pretooluse-shape-guard.py:205-208` — R1's remediation string:

> "devflow shape guard (R1): a leading VAR=value assignment or env-prefix (M=x cmd) is denied by the review-tier matcher. Permitted alternative: capture a command's output with VAR=$(cmd), or pass the value as an argument to the command; do not prefix the command with VAR=value."

`Verified:` `lib/test/run.sh:7037`, inside the `#484` `WITHHELD_IMPL` rationale block for the implement profile:

> "mktemp — invoked only from leading VAR=$(…) capture shapes the matcher refuses regardless of a head grant; granting it would be a no-op greening the guard on a still-denied path."

This names **exactly** the shape R1 recommends.

`Verified:` `CLAUDE.md:46`, the `run-parallel.sh` whole-suite bullet: "...because the matcher refuses caller-side assignments, redirects and background processes (#401/#455)."

`Verified:` `lib/test/extract-command-shapes.py`'s `classify_arms()` (`:566`), backed by `_assignment_violation()` (`:391-414`), returns no `R1` arm for a `VAR=$(cmd)` capture whose substitution spans the whole statement — the exemption is real in the shipped classifier, so nothing turns RED today for the shape R1 recommends.

### The coupled mirror

`Verified:` `docs/internal/cloud-allowlist.md:1308` carries the identical `R1` row, same alternative cell: *"capture a command's output with `VAR=$(cmd)`, or pass the value as an argument."* Any fix to the guard's `REMEDIATION` string is a **coupled edit** with this row — both change in the same commit, or the doc misleads a reader who never opens the guard source. (The issue body above cites `docs/cloud-allowlist.md`; the live path, verified to exist, is `docs/internal/cloud-allowlist.md` — `docs/cloud-allowlist.md` does not exist in the tree. Use the live path in any new work; the citation above is untouched.)

### The carve-out, checked and adjudicated

`docs/internal/cloud-allowlist.md:319` records **Shape 18** — `if VAR=$(granted-helper …)` command-substitution *condition* — as **PERMITTED**, review `probe` job run **30310938175** (2026-07-27), per `.github/workflows/matcher-probe.yml:276`. That is a real recorded review-tier verdict for *a* `VAR=$(…)` capture shape.

It does not settle R1's alternative, for a position reason, not a doubt about the verdict itself:

- Shape 18 tests a capture used **only as an `if`/`elif` condition** — `if VAR=$(granted-helper …); then`. R1's alternative is the general case: "capture a command's output with `VAR=$(cmd)`" as a stand-in for *any* leading-assignment use, most naturally a freestanding statement whose result is used afterward.
- `Verified:` the review `probe` job's own shape list (`.github/workflows/matcher-probe.yml:196-283`, Shapes 1-19) contains no freestanding `VAR=$(cmd)` capture row. The nearest neighbor, Shape 2 (`:201`), is a **literal**-value `M=probe02` on its own line followed by a separate `printf` statement — the denied shape restated, not R1's proposed alternative.
- `docs/internal/cloud-allowlist.md:679-683` discusses the analogous **implement**-tier carve-out (`IR3`, the phase-4 label-loop capture), not the review tier R1 governs, and states outright: *"`R1`, `R3` and `R4` are not enforced on the implement profile at all, because their status there is unmeasured."* `:735-753` ("Rows 8/9 — the non-label-capture disambiguators") goes further: even after the `#571` dispatch recorded verdicts for rows 8 and 9 (row 8 PERMITTED, row 9 DENIED — the table at `:718-723`), the document still treats the underlying capture-descent claim as *"an INFERENCE, not a measurement"* for that tier, because neither recorded row exactly replicates the fences' own capture shape. That is a different tier's classifier, but it shows this repository's own documentation treats a same-family capture carve-out as unsettled even where a probe row exists and returned a favorable verdict — the review tier's R1 alternative has no probe row at all in the position it is recommended for.

**Adjudication:** the two statements are not flatly contradicted by Shape 18 — they are **tier- and position-scoped, not settled**. Shape 18 proves a narrow position (condition-substitution) permitted; R1's alternative is broader and untested in the position it is actually recommended for. `run.sh`'s and `CLAUDE.md`'s blanket "the matcher refuses" statements are broader than the evidence too, in the other direction. Frame this precisely rather than as a flat contradiction: **R1's alternative is unproven for the position it is recommended in.**

### Why this is a precondition, not a nice-to-have

Registering the guard on any tier while R1's alternative lacks a recorded verdict for the position it recommends risks steering runs from one denied shape into another — worse than leaving the guard inert, because a wired-and-wrong remediation actively misdirects an agent that follows it. So settling this is a **second precondition of option (a)**, alongside the existing `#919` dependency recorded below.

### Sibling: #1514

`#1514` (open) — *"A recorded probe verdict says in-workspace `.prflow/tmp` redirects are PERMITTED; three runs report them refused"* — is a **sibling, not a blocker**, of this finding. Same failure class: a recorded shape verdict (or, here, the *absence* of one for the position that matters) diverging from what the guard or the prose asserts. Whatever adjudication method #1514 lands on for reconciling a probe verdict against contrary observed behavior should apply here too.

## #919 is closed — its measurement discharges the firing question, in option (a)'s favor

#919 closed **2026-08-05T03:37:07Z**. It said, verbatim, what its measurement would decide between:

> Whether to (a) wire the guard's `settings:` input onto `devflow.yml` / `devflow-implement.yml` so it is live on a tier that still runs, or (b) accept it as retained-but-inert alongside the rest of the withheld tier, is a **separate open decision**. This issue does not decide it and does not depend on it.

and its own stated decision rule: if the probe returns `NOT-FIRED`, wiring the `settings:` input onto a live tier buys nothing and option (b) is the only honest answer.

**The probe did not return `NOT-FIRED`. It returned `FIRED`** — primary record run `30956039324` (job `92149438739`), the 8th consecutive identical replication — and PR #1308's deny/defer arms (run `30967680822`) independently corroborate it: a hook `deny` is honored and its `permissionDecisionReason` is delivered to the transcript (`REASON-DELIVERED`), and `defer` blocks the tool rather than falling through (`DEFER-BLOCKED`). Per #919's own decision rule, that result selects **continuing toward option (a)**, not closing it out — the inverse of this section's premise before #919 closed.

**What that does and does not settle.** The firing / reason-delivery axis that made this section a *prerequisite* is answered, so the arm-table work below is no longer "building on an unmeasured mechanism" on that axis — that specific risk is retired. It remains gated on two axes #919 never measured (see "Unestablished" above): whether a hook `command` bypasses the tool-grant matcher, and — explicitly flagged **blocking** for `devflow-implement.yml` there — whether `settings:` MERGES with or OVERWRITES the base-restored `.claude/settings.json` hooks, under which the load-bearing `Stop` hooks would silently stop applying. It is also still gated on R1's remediation precondition (the section above this one), which #919 never touched and which this closure does not affect. **Net: #919 is discharged; the ticket is not yet fully unblocked.**

#919 was maintainer-only and post-merge by construction (no `gh` head is granted on any profile, and `matcher-probe.yml`'s `workflow_dispatch` resolves from the default branch) — which is why its dispatch could not have happened inside an autonomous run.

## Technical Context

> **Scope note:** the files below are known starting points, not the full list. Trace the change through the codebase before implementing.

**Change surface if option (a) proceeds**

- `lib/test/extract-command-shapes.py` — a new implement-tier arm table with per-arm provenance, and a pure statement-level classifier (the sibling of `classify_arms()`). The existing `find_implement_violations()` fence scanner is not that function.
- `scripts/pretooluse-shape-guard.py` — tier selection, and an implement-scoped remediation table. Its current `REMEDIATION` strings say *"denied by the review-tier matcher"* / *"granted by no review-tier profile"* (lines 122, 128, 133); emitting those on the implement tier would state a measurement that was never made.
- `.github/workflows/devflow-implement.yml` and/or `.github/workflows/devflow.yml` — the `settings:` input.
- `.github/workflows/matcher-probe.yml` — an implement-tier probe arm per candidate deny arm, since a shape proven on the review tier is unproven here.

**Explicitly NOT needed**

- No `lib/capability-profiles.json` regeneration. The `settings:` input is not one of the five generated `TOOLS` / `--allowed-tools` regions.
- `lib/review-profile.tokens` stays byte-identical — no grant is added to the review profile, so the security-boundary lock does not move.

**Two live suite pins must be reconciled in the same commit**

`lib/test/run.sh:42704` and `:42706` both assert `#908 AC1: devflow-implement.yml registers no settings input / PreToolUse guard`, expecting `yes` (the primary arm is a YAML extractor that fails closed to `extractor-failed`; the fallback arm is a `grep -qE 'settings:|PreToolUse'`). Registering a `settings:` input on `devflow-implement.yml` turns the required `lib + python tests` check RED until both are reconciled. That reconciliation is not incidental cleanup — those pins encode #805's tier-scope decision, so changing them **is** the decision, and it belongs in the same change as the rationale that overturns it.

**Adjacent stale prose (cheap to correct in the same change)**

- `scripts/pretooluse-shape-guard.py`'s module docstring still reads *"`.github/workflows/devflow-runner.yml` passes no `settings` input to the action"*. That clause has been false since #908. (Its companion clause — *"`.claude/settings.json` carries no `PreToolUse` key"* — is still true: `git show HEAD:.claude/settings.json` carries only a `Stop` key.)
- `docs/cloud-allowlist.md`'s **PreToolUse shape guard** section repeats the same superseded statement (*"Registration is not yet wired — the guard is inert at this revision"*, and *"no `settings` input on `devflow-runner.yml`'s review-tier action step"*). The **conclusion** — the guard is inert — happens to still be correct, but for an entirely different reason than the text gives, which is the worse kind of stale: it reads as accurate and hides the real cause.

**Environment / tier considerations**

- **Consumer vs this repo.** A consumer that installed the tier before #937 still has `devflow-review.yml` and therefore a *live* guard; `install.sh`'s `prune_stale_devflow_workflows()` is deliberately unchanged, so upgrading does not remove it. Any change to the guard body or its remediation strings reaches that population through the `prflow_version` vendor fetch, while a workflow change reaches it only through `install.sh`'s file copy — the documented skew hazard. A partially-upgraded consumer can therefore run a new guard body under an old workflow, or the reverse.
- **Cloud vs local.** The guard has no local-tier registration path in this repo: the committed `.claude/settings.json` carries no `PreToolUse` key, and per `CLAUDE.md` an agent may not add one. Any local-tier arming is an operator-side action.
- **Grant timing (#593).** `devflow-implement.yml` triggers on `issue_comment` only, so GitHub resolves the **workflow file** from the default branch. The entire workflow definition a run executes comes from `main`, never from the pull request. A `settings:` input a pull request ships is therefore inert for that pull request's own run.

## Acceptance Criteria

Split by dischargeability, because the verifiability boundary above is hard: registration is checkable in-PR only by **reading the committed file**; every firing or behavior claim is post-merge and needs a dispatch the implement tier cannot perform.

**Dischargeable in-PR (authoring; verified by reading the tree and by the suite)**

- [ ] The wiring decision is recorded with its rationale — option (a) with the tier(s) named, or option (b) with the retained-but-inert status stated where a reader will meet it.
- [ ] If (b): `docs/cloud-allowlist.md`'s guard section and the guard's own docstring state that the guard is registered on a callerless workflow and can fire in no run this repository starts; the superseded "registration is not yet wired" prose is removed. No workflow changes; the two `run.sh` pins stand unmodified.
- [ ] If (a): a per-arm deny-eligibility table for the target tier exists, each arm carrying its own provenance (probe row id or observed run denial), with authoring-only and inference-backed arms **deferred**, not denied — and the review deny set is not reused.
- [ ] If (a): the statement-level implement classifier exists as a pure function over one statement, distinct from the file-scoped `find_implement_violations()` fence scanner, with unit coverage over each arm and its negative controls.
- [ ] If (a): the target workflow's `settings:` input is committed, and `lib/test/run.sh:42704`/`:42706` are reconciled in the **same commit**, with the replacement assertion pinning the new intended state rather than being deleted.
- [ ] If (a): `lib/review-profile.tokens` is byte-unchanged and `python3 lib/generate-capability-profiles.py --check` is clean — i.e. no grant widening rode along.
- [ ] The guard's remediation strings never assert a tier-specific measurement that was not made for the tier the message is emitted on.
- [ ] Stale prose corrected: the guard docstring's `devflow-runner.yml` clause and `docs/cloud-allowlist.md`'s matching statement.
- [ ] R1's remediation string (`scripts/pretooluse-shape-guard.py`) and its mirror row in `docs/internal/cloud-allowlist.md:1308` no longer assert `VAR=$(cmd)` as a settled permitted alternative with no recorded verdict for the position it recommends (a freestanding leading capture) — reworded to mark it unproven, or narrowed to the one position that is actually probe-proven (Shape 18's `if`/`elif` condition-substitution, run 30310938175). Both are reconciled in the same change.

**Post-merge, maintainer-dispatched (NOT dischargeable by an autonomous run — no `gh` head is granted on any profile)**

- [x] **#919's probe result is recorded — `FIRED`, not `NOT-FIRED`.** Discharged on #919's closure (2026-08-05T03:37:07Z): primary record run `30956039324` (job `92149438739`), replicated 8 times, and independently corroborated by run `30967680822`'s deny/defer arms (*Harness hook-surface probe evidence (Part 2)*, `docs/internal/cloud-allowlist.md`). Per this criterion's own stated rule, a `NOT-FIRED` result would have selected option (b) and closed option (a); the actual result is the opposite, so option (a) stays open. The remaining post-merge bullets below are gated on the still-open merge-vs-overwrite unknown and on R1's remediation precondition, not on this one.
- [ ] If (a) proceeded: an implement-tier probe arm is dispatched per candidate deny arm, and each arm's PERMITTED/DENIED verdict is recorded in `docs/cloud-allowlist.md` with its run id, before that arm is promoted from deferred to denied.
- [ ] If (a) proceeded: one real run on the wired tier records whether the guard fired, and — under the merge-vs-overwrite caveat — whether the base-restored `Stop` hooks still applied.
- [ ] A review-tier `matcher-probe.yml` arm for a freestanding leading `VAR=$(cmd)` capture (not inside an `if`/`elif` condition) is dispatched, and its PERMITTED/DENIED verdict is recorded in `docs/internal/cloud-allowlist.md` with its run id, before R1's remediation — or any guard wiring under option (a) — cites that shape as proven.

## Dependencies

**No longer blocked by #919** — closed 2026-08-05T03:37:07Z, probe result `FIRED` (see "#919 is closed" above). The firing / reason-delivery question #919 existed to answer is discharged in option (a)'s favor. Option (a) remains blocked on two separate, still-open items that #919 neither measured nor its closure touches: the merge-vs-overwrite unknown for `settings:` against the base-restored `.claude/settings.json` hooks (flagged **blocking** in "Unestablished" above), and R1's remediation precondition (its own section above).

## Related (context only, not prerequisites)

#805 (the original guard, and the tier-scope rationale this issue revisits), #908 (the registration), #906 (the docs placeholder), #936 / PR #937 (the tier withholding that made the registration point uncallable). All are closed; none blocks this.

The01Geek:
Closes #1047.

## What this does

Two things, one of them a live consumer-facing bug found while working the ticket.

### 1. The `PreToolUse` guard's hook command failed CLOSED on a missing interpreter

The registered hook probed for the guard *script* and exited 0 when it was absent, but ended in a bare `exec python3` — which exits **127** on a host with no `python3` on `PATH`. That is routine on a **self-hosted Windows runner**, which is why this repo ships `scripts/provision-python3-shim.sh` at all.

`scripts/pretooluse-shape-guard.py`'s own `main()` records that *"a non-zero PreToolUse exit is a BLOCK, not a fall-through"*. So for a consumer still running the pre-#937 review tier on such a host, a missing interpreter did not degrade quietly — it blocked **every Bash tool call** in the run.

The command now carries `command -v python3 >/dev/null 2>&1 || exit 0`, so a missing interpreter fails open exactly as a missing script already did.

Covered by a **behavioral** test, not a substring pin — a probe written *after* the `exec` would satisfy any grep for it while changing nothing. Verified discriminating: GREEN on this branch, RED against `main`'s file.

### 2. The wiring question is recorded as decided — option (b), retained-but-inert

#805 registered the guard review-tier-only and called it an explicit *"not yet"*; #919 put the decision out of its own scope; #1795 was closed unimplemented naming #1047 as its home. This records it: **the guard stays registered on the callerless `devflow-runner.yml`, and no live tier registers it.** Rationale, the evidence that would reopen option (a), and why deletion was refused are in `docs/internal/cloud-allowlist.md`, with a pointer from the guard's docstring.

## Correction to the issue's second half

**#1047's R1 claim does not hold, so this PR does not act on it.** The issue says R1's `VAR=$(cmd)` alternative has "no recorded verdict". It has one — `lib/test/extract-command-shapes.py:400` cites **real-run evidence: run 29105381021** seeding its progress comment through exactly a `WP=$(vendored-path create …)` call. That run is the **review-tier** run on PR #397, i.e. the same tier R1's message names ("denied by the review-tier matcher"), and an observed run execution is the evidence class the guard's own deny set already accepts.

The two statements the issue cites as contradicting it are both **implement**-tier, and the tree already draws that line deliberately (`extract-command-shapes.py:686`: *"a shape proven on the review tier is UNPROVEN here"*; `skills/implement/SKILL.md:42`: *"Unproven — fail closed"*).

So the review-side R1 remediation, both docs rows, the five `agents/**` bodies and `skills/review/SKILL.md` are **left unchanged — they are correct as written.**

What *was* real is the inverse, and is fixed here: `lib/test/run.sh`'s `WITHHELD_IMPL` rationale asserted a *measured* matcher refusal of leading `VAR=$(…)` captures on the implement tier. That was never measured — the only implement row that measured a capture (I6) came back DENIED while confounding three properties. The conclusion it supports (do not grant `mktemp`) is unchanged; only the stated reason is corrected. `lib/test/extract-command-shapes.py`'s own "proven-permitted" is likewise softened to "observed-permitted" to match the careful wording that same file already uses at :686.

`CLAUDE.md:60` was considered and **deliberately left alone**: on its natural reading "caller-side assignments" means the env-prefix form, which genuinely is denied, so an edit there would be defensive prose against a misreading.

## Verification evidence

- `lib/test/run-shard.sh monolith` — **9844 passed, 1 failed**; the single failure was this PR's own new assertion reading `main`'s workflow after the checkout was switched to `main` mid-run (the RED half of the pair, firing correctly). Re-verified GREEN against the branch tree afterwards.
- `ruff check` over tracked `*.py` — clean.
- `shellcheck --severity=warning -e SC1091 --extended-analysis=false lib/test/run.sh` — clean.
- `actionlint .github/workflows/devflow-runner.yml` — 3 pre-existing `info` findings, identical count on `main` and on this branch; none in the edited block.
- `lib/test/lint-reference-size.py`, `lint-tree-enumeration.py`, `lint-carveout-guard.py`, `lint-shipped-pruned-path.py` — all rc=0.
- Hook command behavior, measured directly: `python3` absent → **rc 0** (fails open); `python3` present → **rc 9** from a stub guard (interpreter still reached, so the probe is not a no-op).

**Writing-skills evidence: not applicable — no `skills/**` file is touched by this PR.**


The01Geek:
## Disposition of the review on `9cd8457cd`

**Important 1 — environment-fragile present arm: FIXED.** Confirmed by direct reproduction before touching anything: running the extracted hook command from the scratch workspace returned `rc=9` with `CLAUDE_PROJECT_DIR` unset and `rc=0` with it set to the repo root, so the arm asserting `9` would have gone RED exactly as described. Root cause is as stated — `git rev-parse` fails from the `mktemp` workspace, so `_r` falls back to `${CLAUDE_PROJECT_DIR:-.}` and `_g` resolves to the real guard, which exits 0 on empty stdin.

Fixed by pinning `CLAUDE_PROJECT_DIR="$_1047_WS"` in **both** arms, not just the present one — the absent arm had the same escape hatch and was passing only because `PATH` neutralisation masked it.

Re-verified across the full matrix rather than the one failing cell:

| | branch (with probe) | mutant (main, no probe) |
|---|---|---|
| `CLAUDE_PROJECT_DIR` unset | all PASS | RED on fail-open arm (127≠0) |
| `CLAUDE_PROJECT_DIR` set | all PASS | RED on fail-open arm (127≠0) |

The mutant column is the mutation check: the test still goes RED for the reason it pins, in both environments, so the fix did not make it vacuous. The mutant was read via `git show origin/main:` into a scratch path and never entered the working tree.

**Suggestion 1 — false changeset line: FIXED.** Verified against the diff and the finding is correct. The diff makes **one** implement-tier refusal→unmeasured correction (`run.sh` `WITHHELD_IMPL` `mktemp`) and **two** review-tier `proven-permitted`→`observed-permitted` edits in `extract-command-shapes.py` — a file the sentence called an unchanged reference while the diff edits it. Reworded to describe what the diff does. For the record, the sentence was a leftover: it described a `CLAUDE.md:60` correction I deliberately dropped mid-session and never reflected back into the changeset.

**Suggestion 2 — comment provenance: FIXED.** Trimmed to the prevention-only standard. The Windows derivation and the block-on-nonzero provenance now live only in the decision record in `docs/internal/cloud-allowlist.md`, which already carried both.

**Suggestion 3 — `extractor-failed` vs a `host-capability` skip: DECLINED, recorded inline.** PyYAML is a suite *requirement*, not a host capability, so a skip here would manufacture a skipped check on a dependency the suite guarantees — and a skipped check is never a clean pass. The repo's own uniform pattern agrees: the adjacent `#908` extractor fails closed with this identical `extractor-failed` sentinel (`lib/test/run.sh:51179`, mirrored at `:50909`). Converting mine would make the file less consistent, not more. Recorded as a comment at the call site so the next pass sees the reasoning instead of re-raising it.

**On the two disclosed process notes** (neither contested): the acceptance-criteria resolver being denied on this tier, and the `Verification evidence:` marker literal. The PR body carries a `## Verification evidence` section; this PR maintains no workpad, so the PR description is the documented fallback channel.

Local verification after the fixes — `shellcheck` rc=0, `ruff` clean, `coverage_map_guard.py` rc=0, `regenerate-artifacts.py` `all artifacts reconciled — exit 0`.

Issue #1745, opened 2026-08-18
Sweep the remaining "DevFlow" mentions to "PRFlow", excluding the frozen identifier set


The01Geek:
## Problem Statement

The consumer-facing `devflow` → `prflow` rename (issue #1002, Tiers 1–4) deliberately left a large residue, and the tree still reads inconsistently: `DevFlow` prose beside `PRFlow`, `DEVFLOW_*` env vars, `/devflow:` spellings and `devflow-*.yml` filenames, with no single place that says which occurrences are finished renames, which are permanent, and which are still owed. A blanket search-and-replace is not the shape of this work — `lib/rename-map.json` freezes many of these spellings, and renaming a frozen one does not move a setting, it deletes it, usually silently.

## Current Behavior

Measured on `e0c35a30f` (2026-08-17) — a past-time snapshot, not a figure to keep current: 1,049 case-sensitive brand-cased `DevFlow` occurrences in 168 files; ~1,221 `/devflow:` alias mentions; 229 distinct `DEVFLOW_*` identifiers; 340 references to `docs/internal/DEVFLOW_SYSTEM_OVERVIEW.md`; the largest single files are `lib/test/run.sh` (236) and `CHANGELOG.md` (111). Nothing tells a reader which bucket any given occurrence falls into, and nothing stops a new brand-cased `DevFlow` from landing in ordinary prose.

## Desired Behavior

Every brand-cased `DevFlow` occurrence in the tracked tree is classified into a machine-checkable bucket set recorded as data and reconciled against the tree by a check that fails closed. Every occurrence in the "ordinary brand prose, safe to rename" bucket reads `PRFlow`. A new brand-cased `DevFlow` outside a declared frozen or permanent bucket turns the suite RED, so the residue cannot re-accumulate.

## User Impact

Contributors and consumers stop reverse-engineering, per occurrence, whether a spelling is a leftover or load-bearing. The rename reaches a stated, enforced end state instead of an indefinite half-migrated one.

## Technical Context

> **Scope note:** The files and details below are the known starting points, not the full list. Before implementing, trace the change through the codebase to find every affected call site, consumer, and layer — this issue maps the work, it does not bound it.

[...]

Issue #1132, opened 2026-08-02
Single-source the flight scope that discharges the Phase 4.3 completion-evidence gate


The01Geek:
## Problem Statement

The Phase 4.3 completion-evidence gate added by #1119 (issue #1087) admits a verification flight only when it is current, in the terminal `passed` state, carries an integer-`0` `exit_status`, and reports an **empty** `skipped_checks` population. The prompt surfaces that tell a run *which flight* to launch carry two statements about the required scope that do not agree with each other, so which one a run follows decides whether it completes. On 2026-08-02 the two cloud implement runs that reached that phase after the gate landed read the surfaces differently: one recorded a focused-module flight and completed; the other judged that only a zero-skip full-suite flight was honest evidence, did not obtain one inside the runner's per-command execution ceiling, correctly refused to launder a host-capability skip into a clean pass, and dead-ended Blocked at maximum cost with the work already finished and reviewed.

## Current Behavior

`.prflow/prompt-extensions/implement.md` states, one line apart in the same section:

- line 187 — "A focused result discharges intermediate iteration only, never the final completion gate."
- line 194 — "The final full-suite obligation binds the cloud tier too: a cloud completion claim rests on that in-env `lib/test/run-parallel.sh` (or a covering focused module) plus every required lint gate, run in the cloud run's own environment — not on any CI result the run never saw."

The first sentence excludes a focused result from the final gate. The parenthetical in the second admits one. Nothing downstream resolves the tension: `scripts/check-completion-evidence.py` is semantic-free about scope — its `_validate_implement_record` checks the record's shape, state, exit status and skip population, and reads `suite_summary.command` only as "a nonempty string". `skills/implement/phases/phase-4-documentation.md`'s step 1 likewise says to run "the project's already-allowlisted verification command **unchanged**" without deciding which command that is at this phase.

The observed consequence, two runs, same policy, same tip:

- Issue #990, run 30739734583 (created 2026-08-02T08:25:07Z, concluded `success`): recorded the flight over the focused `review-trigger-helpers` module at its 628-assertion floor and reached `🎉 Complete`. Its own workpad note records that `Full lib/test/run-parallel.sh exceeded the 10-min shared-host budget (a wall-clock timeout, not a failure; CI isolates each shard per runner) — pushed to trigger CI as the post-PR merge gate.`
[...]

PR #2002, opened 2026-08-26
perf: fail-fast the two cheap lints before the whole-suite gate launches

The01Geek:
## What

`lib/test/run-parallel.sh` now runs the two sub-second, read-only cheap lints — the #1595 reference-size ceiling and the #1745 brand-baseline sweep — as part of its pre-launch checks, on **both** the coordinator's own flow and the standalone `--preflight` route the #1132 shard-decomposition path names before its shard loop.

## Why

Both lints are `lib/test/run.sh`-resident, so before this nothing cheaper than a whole coordinator pass caught either.

Measured from cloud implement run [32929107360](https://github.com/The01Geek/prflow/actions/runs/32929107360): its Phase 4.3 completion gate ran `run-parallel.sh` for **12.6 min** and went RED on exactly two things — `skills/review-and-fix/references/shadow-review.md` over the reference-size ceiling, and a brand-baseline drift. Both were one-line fixes, and the mandatory relaunch cost a further **12.5 min**. That is 25 of that run's 44-minute Phase 4.

The gate reaches the same verdict in **~0.6s** on the real tree.

## Verdict contract

Deliberately identical to the existing `_artifact_preflight`, and single-sourced the same way (one function, both routes, so `--preflight` and the coordinator cannot diverge):

- **Fail closed** on a positively-attributed finding.
- **Fail open** on anything that leaves the check unusable — a crash, an exit-2 record error, an absent or denied helper.

The refusal comparand is each lint's own **completion sentinel** (`<tool>: audited …`), *not* its exit code: a Python traceback exits 1 exactly as a finding does, so keying on the code alone would turn every crash into a launch refusal. The sentinel is matched at the **start of a line**, so the same text quoted inside an indented row diagnostic stays data — the negative control the artifact preflight already carries, ported here.

Each gate takes an injectable override in the established `DEVFLOW_ARTIFACT_PREFLIGHT` idiom (`DEVFLOW_REFERENCE_SIZE_PREFLIGHT`, `DEVFLOW_BRAND_SWEEP_PREFLIGHT`), using `-` rather than `:-` so an explicitly-empty value disables that gate while an unset one takes the bundled default.

## One deliberate deviation from the brief

The brief asked that the new checks not be "distinguishable only by absence of output". **A clean gate is still silent**, because that is a *settled* contract in this file (recorded in `_artifact_preflight`'s header, raised and disposed on PR #1294) that the prompt-extension prose depends on. Changing it would have broken the documented silence-is-clean reading for the pre-existing preflight too. The intent is met the other way round: every non-clean outcome names itself and its remedy on stderr, and the exit status is the machine channel.

## Dogfooding note

On its first real-tree run the gate immediately caught a genuine brand-baseline finding — introduced seconds earlier by this PR's own test fixture, which quoted the brand literal. Caught in 0.6s instead of 12 minutes, which is precisely the point.

## Verification

- `lib/test/run-module.sh parallel-suite-runner` — **202 passed, 0 failed** (34 new assertions; written RED-first, confirmed at **188 passed / 14 failed** before the fix)
- `lib/test/run-module.sh regenerate-artifacts` — **402 passed, 0 failed** (the coupled module that drives the real preflight end-to-end through this coordinator)
- `python3 lib/test/lint-reference-size.py` — clean; `python3 lib/test/lint-brand-devflow-sweep.py` — clean
- `shellcheck --severity=warning -e SC1091` on both changed shell files — clean
- `lib/test/run-parallel.sh --preflight` on the real tree — exit 0, silent, 0.6s

Per the local-tier ladder this run does not gate on a local full-suite pass; CI on this pushed head is the completion signal.

No `skills/**` file was touched, so no `Writing-skills evidence:` line is owed.

The module's `minimum_assertions` floor is left at 104 in both coupled literals — assertion counts are a floor, and adding assertions does not require raising it.

## Review reception (dc6db692c64a1ff7583ae543de8dc431a262596c)

Verification evidence: local/interactive tier; no whole-suite launch and no CI reading — GitHub Actions is under an outage, so the local-tier whole-suite completion gate for this head is **undischarged** and must be re-read once Actions recovers. Focused in-environment results on `dc6db69`: `lib/test/run-module.sh parallel-suite-runner` — 210 passed, 0 failed; `lib/test/run-module.sh regenerate-artifacts` — 402 passed, 0 failed; `shellcheck --severity=warning -e SC1091` on both changed shell files — rc 0; `python3 lib/test/lint-reference-size.py` and `python3 lib/test/lint-brand-devflow-sweep.py` — both clean. Mutation check (route b, restored): breaking the bundled default path turned `cheap-lint gate real: the bundled default is not silently inconclusive` RED for the reason it pins.


The01Geek:
**Local review (GitHub Actions outage — cloud review tier unavailable)**

Run locally against `658c8458f11dc172858cb17ce8e528cf5983ff1e` (PR head) in an isolated worktree. No commits, no pushes, no formal review submitted.

## Verdict: REJECT — 1 Important, 3 Suggestions, 0 Critical

The change itself is sound, well-motivated, and unusually well-tested. The single Important finding is a two-line fix in the new test arm; everything else is advisory.

---

## Verification evidence (this review)

| Check | Result |
|---|---|
| `lib/test/run-module.sh parallel-suite-runner` | **210 passed, 0 failed** (PR body says 202; the base merge added 8 — cosmetic) |
| `lib/test/run-module.sh regenerate-artifacts` (coupled) | **402 passed, 0 failed** |
| `shellcheck --severity=warning -e SC1091 lib/test/run-parallel.sh` | clean (rc 0) |
| `python3 lib/test/lint-reference-size.py` | clean — `audited 82 of 327 files [whole-tree]` |
| `python3 lib/test/lint-brand-devflow-sweep.py` | clean — `audited 1146 of 1146 files` |
| CI (`lib + python tests`, `lint (shellcheck + actionlint + ruff)`) | **UNESTABLISHED** — see below |

**CI test evidence is MISSING, not passing.** `gh pr checks 2002` reports only a `skipping` Mintlify row; neither required `ci.yml` job has registered for this head. Actions is under an outage, so no run was read and none was waited on. Absence of a failure is not a pass — I am not citing any CI conclusion.

Consequence for the PR body's claim *"Per the local-tier ladder this run does not gate on a local full-suite pass; CI on this pushed head is the completion signal"*: that completion signal does **not** currently exist. Under `CLAUDE.md` rung 1, a head with no run is unestablished, so the local-tier whole-suite gate is undischarged for `658c845`. This needs re-reading once Actions recovers — it is not something this PR can fix.

---

## What I verified about the design

The core claim holds up. Both lints print their completion sentinel on the finding path and return 1:

- `lint-reference-size.py` prints findings, then `f"{_TOOL}: audited {len(covered)} of {len(candidates)} files [{scope}]"`, then `return 1 if findings else 0`.
- `lint-brand-devflow-sweep.py::cmd_check` prints `lint-brand-devflow-sweep: audited N of N files` **before** its two-space-indented finding rows, then `return 1 if findings else 0`.

So the fail-closed arm is genuinely reachable, and the `RecordError` (`return 1`, no sentinel) and bucket-record (`return 2`, stderr only, no sentinel) paths genuinely fail open — matching the stated contract exactly. The start-of-line match is the right discriminator given brand-sweep indents its rows two spaces. The negative controls (`cl-quoted-rsz`, `cl-crash-rsz`, `cl-exit2-rsz`) and the real-helper coupling pin (b) are the right tests, and pin (b) closes the stub-drift gap properly.

`-` vs `:-` for the overrides is correct and consistent with `DEVFLOW_ARTIFACT_PREFLIGHT`. Ordering (artifact preflight first, short-circuit on first cheap-lint refusal) is defensible and covered by an assertion. `set -u` with no `set -e` means `[ "$rc" -eq 0 ] && return 0` is safe. Both routes share one function, so `--preflight` and the coordinator cannot diverge.

---

## Findings

### Important

**1. The real-helper arm (a) couples the module's verdict to unrelated working-tree state, producing a misattributed RED.**

`lib/test/modules/parallel-suite-runner.sh`, arm (a) — the "bundled default resolves and runs" block — invokes the real coordinator from the real checkout with **both cheap-lint overrides unset**, then asserts:

```
assert_eq "cheap-lint gate real: the bundled default resolves and runs (exit 0)" "0" "$PSR_RH_RC"
assert_eq "cheap-lint gate real: the bundled default launches the shard" "yes" ...
```

Both of those are now false whenever the developer's tree carries *any* reference-size or brand-baseline finding — which is exactly the state a developer is in while fixing one. The gate does its job (exit 2, no shard), and the module reports two failures whose names point at "the bundled default", not at the tree. `CLAUDE.md` states there is no known-flake set and every FAIL is a real failure to diagnose, so this costs a real triage cycle in the wrong place.

The third assertion — `is not silently inconclusive` — is the one that actually discriminates default resolution, and it is tree-independent. Suggested fix: keep that assertion, and drop or invert the two tree-dependent ones (e.g. assert only that the run did *not* warn inconclusive, and that it either launched the shard **or** refused by name — both prove the default resolved).

### Suggestions

**2. Arm (a) leaves a run root behind in the developer's checkout.**

Running the module left `.prflow/tmp/parallel-suite/run-73282-0` in this worktree. The module's `rm -rf "$PSR_ROOT"` covers its own scratch root but not the real coordinator's, since arm (a) runs from `$LIB/..`. It is gitignored and harmless per-run, but it accumulates one directory per module invocation, and a stale PID-keyed run root under that path is precisely the artifact the "stop a suite process by recorded PID" triage procedure has to reason about. Consider capturing and removing it, or pointing arm (a)'s run root elsewhere.

**3. Comment prose exceeds this repo's prevention-only comment standard.**

`run-parallel.sh` gains a ~25-line block comment and the module gains several multi-paragraph narrative comments. `CLAUDE.md`'s comment discipline binds library/script source **and test files**, caps a prose comment at three physical source lines, and moves derivation, worked example, provenance, and design narrative out. The contract-header carve-out covers the specification statements (the verdict contract, the sentinel-not-exit-code rule, the `-` vs `:-` idiom) — but the surrounding provenance and narrative are outside it: the measured-run derivation, "this is the 12.5-minute relaunch the run above paid for", "which is the whole point", and the two-exposure essay above the real-helper arms. The specifications are worth keeping inline; the story belongs in the issue and in `docs/internal/`.

**4. `CLAUDE.md`'s `--preflight` description now under-describes the remedy set.**

The shard-decomposition bullet describes `--preflight` as "read-only, sub-second, same verdict contract as the coordinator's own pre-launch check — regenerate and commit any drifted artifact it names". The verdict-contract half stays accurate by deferring to the coordinator, but the remedy clause now only covers one of the three things `--preflight` can refuse on. A run that gets a reference-size or brand refusal there is pointed at "regenerate the artifact", which is the wrong action. Worth a one-clause update.

### Advisory (non-blocking, repo prompt-extension clause)

**5. No `Verification evidence:` marker on any durable surface.** `closingIssuesReferences` is empty, so there is no issue workpad, and the PR description carries no `Verification evidence:` marker. Tier classification: **local/interactive** (no `prflow:checkpoint gha:` row exists to read, because no workpad exists). Non-blocking and it does not affect the verdict.

### Gates checked and clear

- **Prompt-surface edit routing gate:** the diff touches `.changeset/`, `lib/test/run-parallel.sh`, and `lib/test/modules/parallel-suite-runner.sh` — none of the trigger globs. No `Writing-skills evidence:` line is owed; the PR body says so correctly.
- **Wording-only pin policy:** the new assertions read *process output* of `run-parallel.sh` and of the real lints, not file text. These are executable-boundary checks, not prose pins, so the issue #810 declaration requirement does not attach.
- **Shipped-prose cost audit:** no `skills/**` or `agents/**` prose added. N/A.
- **`minimum_assertions` floor left at 104:** correct — assertion counts are a floor and adding assertions does not require raising it.
- **Consumer-repo gotchas:** the `$cmd` word-split (`# shellcheck disable=SC2086`) means a `$SCRIPT_DIR` or override path containing a space breaks the invocation. This is repo-internal `lib/test/` tooling and matches the established `DEVFLOW_ARTIFACT_PREFLIGHT` idiom, so I am not raising it as a finding — but it is worth knowing if this pattern is ever copied to a consumer-facing surface.

---

## Summary

Ship it once finding 1 is addressed and CI can actually be read for the head. The perf argument is well evidenced (25 of 44 minutes of Phase 4 on the cited run, versus ~0.6s here), the verdict contract is correctly modelled on the existing preflight, and the sentinel-not-exit-code discriminator is the right call with the right negative controls.

The01Geek:
**Reception pass — `658c8458f` → `dc6db692c64a1ff7583ae543de8dc431a262596c`**

All four findings verified against the code before acting. Nothing refuted — one severity note below.

### 1 (Important) — real-helper arm (a) coupled to working-tree state — **accepted, fixed**

Confirmed. Arm (a) invokes the real coordinator from `$LIB/..` with both overrides unset, so `_cheap_lint_preflight` resolves the bundled defaults and audits the developer's own tree. On a tree carrying any reference-size or brand finding the gate correctly exits 2 and launches nothing, and both `exit 0` and `launched shard alpha` go false — two FAILs named after "the bundled default", pointing at the wrong subject, in a repo with no known-flake set.

Fixed as suggested, keeping the discriminating half and making the rest tree-independent:

- kept `... is not silently inconclusive` (the assertion that actually pins default resolution);
- replaced `launches the shard` with `launched the shard or refused by name` (either outcome proves the default resolved and reached a verdict);
- replaced `exit 0` with rc in `{0, 2}` — 0 to proceed, 2 being the gate's own refusal.

Mutation-checked (route b, mutation restored, tree re-verified clean): pointing the bundled default at a nonexistent filename turned exactly `cheap-lint gate real: the bundled default is not silently inconclusive` RED — 209 passed, 1 failed — and nothing else. So the discriminator still discriminates after the change.

### 2 (Suggestion) — run root left in the checkout — **accepted, fixed**

Confirmed: the coordinator's `RUN_ROOT` is allocated under `$REPO_ROOT/.prflow/tmp/parallel-suite`, and `REPO_ROOT` resolves from `SCRIPT_DIR`, so arm (a) allocates in the real checkout while the module's `rm -rf "$PSR_ROOT"` only covers its own `mktemp` root.

Arm (a) now parses the coordinator's own `run-parallel: retained logs: <root>/logs` line with a bash-builtin `read`/`case` loop and removes the parent, guarded by a `*/.prflow/tmp/parallel-suite/run-*/logs` pattern so the `rm` structurally cannot escape that parent. Verified empirically: after a full module run, `.prflow/tmp/parallel-suite/run-*` matches nothing.

### 3 (Suggestion) — comment prose over the prevention-only standard — **accepted, fixed**

Trimmed in both files. Kept inline, as the contract-header carve-out allows: the verdict contract (refuse on attribution / fail open on unusable), the sentinel-not-exit-code rule with its start-of-line match, the coupled-contract note, and the `-` vs `:-` idiom — the last moved down beside the overrides it governs, written as the prohibition. Removed: the measured-run derivation, "this is the 12.5-minute relaunch the run above paid for", "which is the whole point", the two-exposure essay, and the synthetic-tree design narrative (kept its one prevention clause). Net −21 lines in `run-parallel.sh`, and the module's block comments are now prohibition-shaped and at or under three lines each.

### 4 (Suggestion) — `CLAUDE.md` `--preflight` remedy clause — **accepted, fixed**

Confirmed: the bullet's remedy clause named only artifact regeneration, while `--preflight` can now refuse on three things. Rewritten to `apply the remedy the refusal itself names: regenerate and commit a drifted artifact, trim a file the reference-size ceiling names, or classify a brand-sweep occurrence`. Grepped `lib/`, `scripts/`, `skills/`, `.prflow/`, `docs/` for the old phrase — no pin asserts it, so the reword is unblocked. Made directly under the autonomous-run `CLAUDE.md` carve-out rather than through the revise skill, which would tail-call and stall.

### 5 (Advisory) — no `Verification evidence:` marker — **accepted, recorded**

`closingIssuesReferences` is empty, so per the reception prompt extension the durable surface is the PR description. Added there, and it records the gate as **undischarged** rather than implying otherwise.

### Severity note (calibration, not pushback)

Finding 1 fails *closed* — the gate refuses and launches nothing, a visible stop with no wrong output and no skipped guard. Its real cost is misattribution and a wasted triage cycle, not a corrupted result. Important is still the right grade given the no-known-flake rule; recording the fail-direction rather than restating the label.

### Verification (in-environment; CI not cited)

Actions is still under an outage. Per this repo's fix-loop rule, no CI conclusion is cited for this pass's own progress, and no run was waited on. On `dc6db69`:

| Check | Result |
|---|---|
| `lib/test/run-module.sh parallel-suite-runner` | 210 passed, 0 failed |
| `lib/test/run-module.sh regenerate-artifacts` | 402 passed, 0 failed |
| `shellcheck --severity=warning -e SC1091` (both changed `.sh`) | rc 0 |
| `python3 lib/test/lint-reference-size.py` | clean — audited 82 of 327 files |
| `python3 lib/test/lint-brand-devflow-sweep.py` | clean — audited 1146 of 1146 files |
| leftover run root after a module run | none |
| CI required checks | **unestablished** — not read, not waited on |

Your observation stands unchanged: the local-tier whole-suite completion gate for this head has no CI reading and is undischarged until Actions recovers. That is now stated on the PR body rather than left implied by the earlier "CI on this pushed head is the completion signal" sentence.

PR left in draft; not marked ready and not merged.

Issue #1655, opened 2026-08-12
Name the running plugin version, model, and effort in the implement PR provenance line


The01Geek:
## Problem Statement

When a PRFlow-authored pull request turns out to be wrong, there is no way to tell what produced it. The PR body's provenance line names the command and nothing else, so a maintainer triaging a bad PR in a consumer repository cannot tell which plugin version, which model, and which reasoning-effort setting were in play. The most common triage question — "is this repository running an outdated PRFlow?" — cannot be answered from the artifact itself.

## Current Behavior

The `/prflow:implement` Phase 3.1 draft-PR create fence writes a fixed provenance line into the new PR body, identifying the command and the issue argument. It carries no version, no model, and no effort setting.

Two properties shape the work:

- The line is written only on the create arm. A resumed run that adopts an existing PR does not rewrite the body.
- The line survives Phase 4.2's PR-description regeneration only by accident: `prflow:pr-description` treats a body carrying no markers as pre-marker content and re-emits it. Nothing names the provenance line, and nothing asserts it survives.

## Desired Behavior

A `/prflow:implement` run's draft PR body carries a single-line provenance statement naming the plugin version that executed the run, together with the model and reasoning effort when those can be established:

```
Generated via /prflow:implement (v2.32.58, claude-opus-5, high)
```

The line carries no backtick, so no shell-active construct can reach the double-quoted `--body` it is substituted into. A value that cannot be established is omitted rather than guessed, so a run with no readable source renders the version alone:

```
Generated via /prflow:implement (v2.32.58)
```

A cloud run has no model or effort source and renders the version alone by that same omission rule. The issue argument is dropped from the line; the body's separate `Resolves #N` line already carries the ticket reference. A repository that does not wish to publish its model and effort configuration switches that clause off through configuration and still gets the version.

## User Impact

A maintainer triaging a PRFlow-authored pull request reads the plugin version, model, and effort from the PR body itself, without opening an Actions run and without asking the reporter. A maintainer sweeping many repositories can tell which are running an outdated plugin build. An operator who does not want their model configuration published turns that clause off without losing the version.

## Technical Context

> **Scope note:** The files below are the known starting points, not the full list. Before implementing, trace the change through the codebase to find every affected call site, consumer, and layer.

[...]

Issue #1611, opened 2026-08-11
Accept a CI-derived completion-evidence record so a local implement run following #1607's tier ladder can write Complete


The01Geek:
## Problem Statement

Issue #1607 made a CI reading this repository's **local/interactive** whole-suite completion gate. It changed prose only. The mechanical gate that decides whether a `--status Complete` write lands did not move, and it cannot express the new signal — so a local `/prflow:implement` run that follows the tier ladder correctly is refused at its own terminal write, with no honest way to satisfy the refusal.

The maintainer hits this on every local implement run in this repository. The run does the right work, establishes a green required check for the commit it pushed, and is then told its completion evidence is missing. The only ways past are to run a suite the tier ladder says is not the gate, or to record a flight that misdescribes what was actually verified.

## Current Behavior

`scripts/workpad.py`'s `_terminal_complete_gate` runs `_completion_evidence_verdict` first, before any PATCH, on every `--status Complete` write. That member is tier-agnostic — it carries no tier discriminator — and recognises exactly one marker family.

Verified: `scripts/workpad.py` carries `_COMPLETION_MARKER_RE = re.compile(` followed by `    r'<!--\s*(?:pr|dev)flow:checkpoint\s+completion-verification:([^\s]+?)\s*-->'`.

Verified: `scripts/workpad.py`'s `_completion_evidence_verdict` refuses with `"refusing to finalize Status: Complete — no completion "` / `"verification-flight marker present [missing-evidence]. Record one with "` when no such marker is present.

The single marker's payload is a verification-flight key, and the record it names is validated by the implement-completion context, which requires a passed in-environment suite run.

Verified: `scripts/check-completion-evidence.py`'s `_validate_implement_record` raises `TOK_MISSING` with `"suite_summary.command is missing or not a nonempty string"`, raises `TOK_NOT_PASS` with `f"flight state is {state!r}, not {FLIGHT_STATE_PASSED!r}"`, raises `TOK_NOT_PASS` with `f"suite_summary.exit_status is {exit_status}, not 0"`, and raises `TOK_SKIPPED` with `f"flight records {len(top_skips)} skipped check(s); implement completion "` — with `_check_stale_candidate` raising `TOK_STALE` with `"verification evidence predates the shipped candidate "` on an identity mismatch.

None of the three writers can read a GitHub Actions conclusion, and the flight ledger structurally refuses to record one: a declaration that admits any external dependency is rejected before a key is derived.

[...]

Issue #1289, opened 2026-08-04
shard-tally combine's --expect is caller-supplied and never reconciled against the true shard partition


The01Geek:
## Problem Statement

`lib/test/shard-tally.py combine` fails closed when it is handed **fewer** tally directories than `--expect` names, but `--expect` is a caller-supplied number that is never reconciled against the true partition from `lib/test/run-shard.sh --list-shards`. A recombination over a subset of shards therefore prints a summary that is byte-shaped exactly like a complete one and exits 0. Nothing mechanically distinguishes "all five shards recombined" from "one shard recombined and the caller said it expected one".

This is **pre-existing** — the `--expect` contract has had this shape since the guard was added, and no recent PR changed it.

## Current Behavior

The guard is a floor against the caller's own claim (`lib/test/shard-tally.py:280-289`):

```python
if len(dirs) < args.expect:
    ... "a shard is missing; refusing to report a green gate over an incomplete run"
    return 1
```

`--expect` is `required=True` (`:415-419`) with the explicit rationale at `:411-414` that an optional guard is one an invocation can silently disable by omission. That reasoning is sound and the flag is doing what it was built to do — the residual is that the number itself is unvalidated, so `--expect 1` disables the guard just as effectively as omitting it would have.

Observed, against a retained run root in this checkout:

```
$ lib/test/run-shard.sh --list-shards
monolith
python-pool
modules-pin
modules-large
modules-rest

$ python3 lib/test/shard-tally.py combine --expect 1 \
    .prflow/tmp/parallel-suite/run-3814-0/tally/modules-large
2808 passed, 0 failed

shard-tally combine: 1 shard(s): modules-large
rc=0

$ python3 lib/test/shard-tally.py combine --expect 2 \
    .prflow/tmp/parallel-suite/run-3814-0/tally/modules-large \
    .prflow/tmp/parallel-suite/run-3814-0/tally/modules-pin
2851 passed, 0 failed

shard-tally combine: 2 shard(s): modules-large, modules-pin
rc=0
```

The leading `N passed, M failed` line is the same contract a whole-suite result is read from. The trailing `shard-tally combine: N shard(s): <names>` line does name the members, so the information is present — but it is one prose line a reader has to notice and compare against a partition they have to already know, not an enforced check.

## Why it is currently harmless, and why it still matters

[...]

Issue #742, opened 2026-07-23
Adopt ruff 0.16.x and work through the current findings, then advance the CI pin


The01Geek:
> **Status (2026-08-25): PR 1 of 2 has merged.** [PR #1964](https://github.com/The01Geek/prflow/pull/1964) (`5fc1f6591550`) created the repo's first `ruff.toml` with the four documented-convention ignores. This issue is now **PR 2**: the mechanical fix pass, the remaining rule rulings, and the pin advance. Scope below is settled — refine figures, not approach.

## Where this stands

- `ruff.toml` **exists** at the repo root, carrying `PLW1510`, `BLE001`, `EXE001`, `S110` with one-line rationales. The "config surface must be created" note from the original body is discharged.
- CI still pins `ruff==0.15.*`, which resolves to `0.15.22` today. The repo is **green** on that pin. Nothing is broken; there is no deadline.
- Latest ruff is **0.16.4**.

**Since this issue was filed, issue #1621 changed the risk profile.** `lib/test/run.sh` now runs `ruff check` over the `git ls-files '*.py'` set inside the `monolith` shard, so ruff findings redden **`lib + python tests` — the required check** — not just the advisory `lint` job. A bad pin advance therefore blocks merges repo-wide, and the pin advance must land atomically with the fixes.

## Measured on merged `main` (`5fc1f6591550`), ruff 0.16.4, 145 tracked `.py`

Re-derive before starting; these decay. `git ls-files '*.py' | xargs ruff check --statistics`.

| | |
| --- | ---: |
| findings, config live | **1500** across 42 rules |
| cleared by `--fix --unsafe-fixes` | **1519** (fixes cascade) |
| **residual needing judgment** | **191** across 17 rules |
| files touched by the mechanical pass | **114 of 145** |
| changed lines | **~3,492** |

**The `RUF100` interaction is the non-obvious part.** Ignoring `BLE001` made 63 live `# noqa: BLE001` directives redundant, so `RUF100` rose 170 → 226. The mechanical pass auto-fixes them, but that means the diff strips those directives — expect it, and check nothing depends on their presence.

**The exec-bit hazard did not reproduce.** `ruff --fix` dropped executable bits on PR #741, but on 0.16.4 across all 145 files, **zero** changed mode. Re-verify rather than assuming either way.

## Rule dispositions (settled 2026-08-25 — implement these, do not re-litigate)

Extend `ruff.toml` with these, each carrying a one-line rationale in the style PR #1964 established:

[...]

PR #2005, opened 2026-08-26
fix: require whole-suite evidence at the implement completion gate

The01Geek:
## Why

Cloud implement run [32957163134](https://github.com/The01Geek/prflow/actions/runs/32957163134) (issue #742) reached `🎉 Complete` and published PR #1997 **having run no test suite at all**. Its Phase 4.3 verification flight was claimed and finished `--result passed` declaring:

```
git ls-files '*.py' | xargs python3 -m ruff check --no-force-exclude
```

`scripts/check-completion-evidence.py` enforced only that `suite_summary.command` was a *nonempty string* and that `skipped_checks` was empty, so a ruff invocation satisfied the whole-suite completion gate.

This was found while analysing where Phase 4 time goes across five uninterrupted implement runs. It is why that run's P4 was 10.3 min against 17–44 min for the four compliant ones.

## What changed

**(A) The declared command must be whole-suite evidence.** `check-completion-evidence.py` gained `classify_whole_suite_command`, which requires some segment of the declared command to *invoke* one of:

- `lib/test/run.sh` — the serial full suite
- `lib/test/run-parallel.sh` — the shard coordinator
- `lib/test/shard-tally.py combine … --require-shards …` — a partition-reconciled recombination

Refused as `missing-evidence`: a focused module, a single shard, a focused Python file, a lint, a runner named as an *argument* rather than invoked, and an unreconciled `combine` (whose count floor alone can be met by a subset, per #1289).

It sits in the same first-failing-class position the nonempty check already occupied and mints **no ninth token**. Accepted spellings include `./`-prefixed and absolute paths, a `bash`/`python3` wrapper head, a leading `VAR=` assignment, and any pipeline or list segment — so `cd <root>; lib/test/run-parallel.sh` and `lib/test/run-parallel.sh 2>&1 | tail -25` both classify correctly.

**The writer is deliberately unchanged.** Flights coordinate focused and shard executions too (`CLAUDE.md`: consult the flight before *any* suite execution — whole-suite, shard, or focused), so `finish --result passed` on a focused run must stay valid. Only the completion gate requires a whole-suite result. The existing writer/reader parity test's fixture was reconciled in this same change and now carries a comment recording that the command is deliberately *not* a parity dimension.

The CI-derived marker family (`--record-completion-evidence-ci`, issue #1611) is untouched — it is a separate, legitimate local-tier path with its own validation.

**(B) `verification-flight.py finish --from-runner-log <path>`.** Terminal evidence can now be derived from the log a runner retained — every runner names its retained log's absolute path on exit — instead of hand-authoring a `--summary-file`. Two of the five runs analysed wrote that JSON by hand after reading `--help` inline; that improvisation is what made (A)'s bypass reachable.

The log's own verdict is the authority. For the coordinator the `aggregate CLEAN`/`FAILED` marker decides the result, **not** the tally — the coordinator returns non-zero for a shard that did not complete even when the recombined tally reads clean. A `--result` the log contradicts is refused **non-terminally**, exactly like the existing exit-status backstop: the flight stays `running` and re-finishable, so the truthful `finish` can still land.

Adversarial input-shape matrix, each with its own named refusal reason rather than a silent misread:

| Shape | Reason |
|---|---|
| absent / unreadable file | `runner_log:<ExcClass>` |
| empty, no marker, binary-ish | `runner_log_unrecognized` |
| marker present, tally truncated away | `runner_log_no_tally` |
| `--result passed` over a failing log | `runner_log_contradicts_result` |
| clean-looking tally under `aggregate FAILED` | `runner_log_contradicts_result` |
| both `--from-runner-log` and `--summary-file` | `summary_source_ambiguous` |

A skip population survives derivation into `skipped_checks` — including a counted-but-unitemized one — so it reaches the reader's no-skip policy rather than being laundered into a clean pass.

**Prose.** `skills/implement/phases/phase-4-documentation.md` §4.3 stated the old contract verbatim ("The summary's nonempty `command` … enforced by …") and its fence taught the hand-authored `--summary-file` path. Both are corrected in this change; the file grew 270 bytes to 60,088, still under the #1595 ceiling (`lint-reference-size.py` green).

## Expected consequences — please read

- **(A) will raise the `Blocked` rate.** Runs that currently slip through on non-whole-suite evidence will now be refused at the `--status Complete` write. That is the intent, but expect a step change rather than a trickle.
- **Every published P4 timing figure understates the true cost.** The 10.3-minute P4 that motivated this was a gate bypass, not a fast path. With the hole closed, the honest P4 baseline is the ~28.6 min mean over the four compliant runs, not the 24.9 min five-run mean.

## Verification

Local focused runs (per the tier rule, the whole-suite gate for this PR is CI on the pushed commit, not a local aggregate):

- `lib/test/test_verification_flight.py` — **134 passed, 0 failed**, including 10 new `TestFinishFromRunnerLog` cases
- `lib/test/test_python_scripts.py` — **6112 passed, 29 failed**; all 12 new `#742` assertions green. The 29 are the known `#1655` local-only environment failures, unchanged from the pre-change baseline on this host (which was 29 + my 7 RED assertions = 36)
- `lib/test/run-module.sh harness-python-guards` (the coverage-map owner of `verification-flight.py`) — **45 passed, 0 failed**
- `lint-reference-size.py`, `lint-shipped-pruned-path.py`, `lint-skills-glob-guard.py`, `lint-anchor-fallback-arm.py`, `lint-worktree-fence-shapes.py` — all rc 0
- `ruff check` clean on both changed scripts and on the flight test file

RED-first: the 7 negative `#742` assertions and all 10 `TestFinishFromRunnerLog` cases were written first and confirmed failing (the 5 positive `#742` cases passed before *and* after, which is the discriminating boundary — the change must not widen what already counted as whole-suite evidence).

Writing-skills evidence: skill-loaded=yes (`superpowers:writing-skills` invoked before the `skills/**` edit); guidance-applied=yes (form matched to failure type per *Match the Form to the Failure* — the failure is prose stating a contract that is now false, so the fix is a positive restatement of the actual contract, not a prohibition, and it keeps the instruction-plus-one-consequence shape); pressure-scenario=no (the edit is a factual contract correction to a phase reference, not a discipline rule — there is no agent rationalization for a scenario to discriminate, and the baseline "failure" is a false sentence, not a behavior); micro-tests=no (the skill's own checklist marks wording micro-tests N/A for reference content, and this sentence's correctness is verified against the code it describes rather than by wording variance).


The01Geek:
**Local review (GitHub Actions outage — cloud review tier unavailable)**

**Verdict: REJECT** — 3 Critical, 6 Important, 7 Suggestions.

Run by the shared PRFlow review engine on a local, read-only checkout: no push, no commits, no branch changes, and no formal GitHub review — this is a plain comment. Reviewed HEAD: `e8332372ddfe2e9e42c5f4a243c038869d4f0a38`. Diff classification: `engine_self_modifying` → full checklist. Phase 3 roster: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer.

**Test evidence: MISSING / unestablished.** `gh run list --commit e8332372ddfe2e9e42c5f4a243c038869d4f0a38` returns no run at all, and neither required check (`lib + python tests`, `lint (shellcheck + actionlint + ruff)`) appears in the rollup — only an unrelated `Mintlify Deployment / skipping` row. That is UNKNOWN, not green. Per the tier rule this PR's own Verification section invokes, the whole-suite gate for this branch is a CI reading on the pushed commit; no such reading exists yet, and no CI conclusion is claimed anywhere below.

Every finding was reproduced against PR-head bytes, not inferred from the diff.

---

## Critical

### C1 — `--from-runner-log` cannot consume any log this repository's runners actually retain

The change makes this the *prescribed* completion path: `skills/implement/phases/phase-4-documentation.md` now says "Finish from the runner's own retained log — `--from-runner-log <path>`, the absolute path the runner named on exit", and the fence drops `--summary-file` for it. Traced to the producers, no path through that instruction reaches a pass:

- **Coordinator.** `lib/test/run-parallel.sh:597` prints `run-parallel: retained logs: $RUN_ROOT/logs` — a **directory** (`:328` `mkdir -p "$RUN_ROOT/logs"`; `:478` `shard_log="$RUN_ROOT/logs/$shard.log"`). Passing it to `_read_runner_log` raises `IsADirectoryError` → `runner_log:OSError`. The `run-parallel: aggregate CLEAN` line (`:602`) — the sole operand that yields `command == "lib/test/run-parallel.sh"` — exists only on the coordinator's console stdout; nothing tees it to a file.
- **Per-shard log.** `run-shard.sh` prints `run-shard.sh: retained log: <abs>` (`:142`) *after* `cat "$LOG_FILE"` (`:136`), so that marker is **not inside the file it names**. The file is the redirected `run.sh` / `run-module.sh` output; feeding it in derives `lib/test/run-shard.sh`, which the new classifier then refuses as not whole-suite.
- **Serial `lib/test/run.sh`.** Measured against PR-head bytes:

```
serial full-suite log ('6112 passed, 0 failed, 2 skipped' + SKIP lines)
  -> REFUSED runner_log_unrecognized
```

This is the CLAUDE.md guard-operand rule exactly: every comparand the new recognizer reads traces back to no producer that emits it into a retained file. `TestFinishFromRunnerLog` cannot catch it because `PARALLEL_CLEAN` is a hand-written string whose shape no retained file has. Net effect: the completion gate can still only be discharged by hand-authoring `--summary-file` — the improvisation this PR set out to remove — while the shipped skill body directs the agent down a path that refuses it.

Fix direction: either have `run-parallel.sh` tee its own stdout to a retained coordinator log and name *that* file (a coupled producer change in this PR), or correct all three prose sites to instruct the caller to capture the coordinator's own merged output and pass that file — and add a `lib/test/run.sh` marker so the serial suite is recognized as itself.

### C2 — The classifier accepts the `monolith` shard, spelled with its own env prefix

`_segment_tokens` drops leading `VAR=value` tokens before classification. Measured:

```
'DEVFLOW_SKIP_SUITE_MODULES=1 lib/test/run.sh'                            -> True
'DEVFLOW_SKIP_PYTHON_POOL=1 bash lib/test/run.sh'                         -> True
'DEVFLOW_SKIP_SUITE_MODULES=1 DEVFLOW_SKIP_PYTHON_POOL=1 lib/test/run.sh' -> True
'lib/test/run-parallel.sh --preflight'                                    -> True
'lib/test/shard-tally.py combine --tally a.json --require-shards monolith' -> True
```

The third line is verbatim the `monolith` shard: `lib/test/run-shard.sh:113-114` prints `monolith shard — bash lib/test/run.sh (DEVFLOW_SKIP_SUITE_MODULES=1 DEVFLOW_SKIP_PYTHON_POOL=1)` and runs exactly that. CLAUDE.md states `monolith` "**never discharges a completion gate**", and this PR's own new test asserts `lib/test/run-shard.sh monolith` → `missing-evidence`. The same run, spelled two ways, now gets opposite verdicts — one env prefix away from the failure class the change exists to close. `--preflight` is documented as "read-only, sub-second" and runs no tests at all; `--require-shards monolith` reconciles against a one-shard "partition".

The recombination arm already shows the right shape (it inspects `rest`). The runner arm should too: inspect the stripped assignments rather than discarding them, and refuse non-executing flags — each on the `False` arm with a named detail, so the breadcrumb tells the run what to re-run.

### C3 — A module log's `K skipped` is silently dropped

`lib/test/run-module.sh:609` emits `Module <id>: N passed, M failed, K skipped`. `_RE_MODULE` captures only groups 1–3; `_RE_TALLY` is anchored on a leading digit so it never matches a `Module …` line; the fallback is `tallies = [(module.group(2), module.group(3), None)]` → `skipped_count = 0`. Measured:

```
'Module regen: 40 passed, 0 failed, 1 skipped'
  -> command=lib/test/run-module.sh regen  passed=40 failed=0 skipped_checks=[] result=passed
```

`shard-tally.py`'s `_MODULE_SUMMARY` already carries that optional `skipped` group, so the producer format is documented and this parser diverges from it. This is precisely the laundering the code's own step-3 comment claims to prevent ("dropping it would launder a skip into a clean pass"), and `test_module_log_derives_its_module_command` uses a skip-free fixture, so it is uncovered.

---

## Important

### I1 — `lib/test/run.sh` is accepted by the gate but unrecognized by the derivation, and can be misattributed

`WHOLE_SUITE_RUNNERS` includes `lib/test/run.sh` and the new `#742` table asserts it passes, but `derive_summary_from_runner_log` has no `run.sh` arm. Worse than a plain refusal: a `run.sh` log contains `Module <id>: …` lines (`run-module.sh:609/612`, re-emitted through `run-shard.sh`), so the `elif module:` branch matches the **first** such line and derives `command = "lib/test/run-module.sh <first-module-id>"` while attaching the whole suite's last tally — a genuine full serial suite both misattributed and then refused as "a focused module". The two halves of this PR disagree about the serial full suite, and nothing tests the seam.

### I2 — "last bare tally wins" overrides the authoritative tally line

`tallies = _RE_TALLY.findall(text)` runs first; the `Module` line is consulted only if no bare tally exists at all. `shard-tally.py`'s `_parse_log` deliberately does the opposite for the modules tier and documents why: a module log's captured output "contains many `N passed, M failed` and `  SKIP  ` lines". So a nested sub-suite's `12 passed, 0 failed` appearing last yields `failed = 0` → `derived_result = "passed"` for a module run that failed. The completion gate refuses the module *command*, but the flight is a **reuse coordinator** — a later run reusing that clean flight skips re-executing a suite that failed. Marker precedence has the same inversion: `_RE_SHARD` is the outermost runner's marker and should be tried before `_RE_MODULE`.

### I3 — The decisive coordinator marker is on stderr, and the stream requirement is unstated

`run-parallel.sh:602` writes `aggregate CLEAN` to stdout; `:604` writes `aggregate FAILED …` to **stderr** (as do the `MISSING`/`SHARD_RCS` diagnostics at `:551`/`:555`/`:566`). A log captured with `tee` or `>` therefore carries the CLEAN marker on a passing run and *no aggregate marker at all* on a failing one. Today that lands on `runner_log_unrecognized` — fail-closed, but incidentally so, and one concatenation away from I2's misattribution. It also means `test_failing_log_contradicting_a_passed_result_is_refused` asserts a shape a real single-stream capture never produces. State the merged-capture requirement in the flag help and the Phase 4 fence, and give "a `run-parallel:` line present but no `aggregate` line" its own named reason so the caller is told they lost stderr.

### I4 — The new shipped prose states an incomplete accepted set

The replacement sentence says the gate "requires the recorded `command` to be a whole-suite result (the coordinator, or a `--require-shards` recombination)". It omits `lib/test/run.sh`, the serial full suite, which `WHOLE_SUITE_RUNNERS` explicitly accepts and this PR's own test pins as passing. This is agent-executed prose telling an implement run what will be accepted; understating the set pushes runs off a valid path. It also carries two consequence clauses where the instruction-plus-*one*-consequence rule allows one. Given C1 the sentence needs rewriting regardless — make it name a log path that actually works.

### I5 — Skip itemization is scanned globally, not positionally

`_RE_SKIP_LINE.findall(text)` collects every `  SKIP  ` line anywhere in the log. `shard-tally.py`'s `_parse_log` deliberately counts only the lines *after* the real summary, stating the reason inline (fixtures emit `  SKIP  ` lines earlier in the capture). It also swallows `summary.sh`'s four diagnostic placeholders — `  SKIP  (detail unavailable …)`, `  SKIP  (skip tally unavailable …)`, the itemization-shortfall line, and `shard-tally`'s omitted-list line. Measured, a `0 skipped` tally beside one placeholder yields a non-empty `skipped_checks`, which the completion reader then refuses. Over-counting fails closed, so this is Important rather than Critical — but it refuses a clean run over a cosmetic breadcrumb. Relatedly, the captured `[<kind>]` (`blocking-gate` vs `host-capability`, load-bearing per CLAUDE.md) is discarded into `kind: "unparsed"` though the producer emits it in a fixed position.

### I6 — Two new comments state facts the shipped code refutes

- `_RE_MODULE`'s comment claims the `Module <id>: …` line "is also the only place the module id appears in its own log". `run-module.sh:654` also emits `Module %s: heavy units REQUESTED bounded …` with `$MODULE_ID`, and `:706` prints `Log: %s` whose path embeds it. False twice over, and inert — the regex works regardless.
- `_RE_TALLY`'s comment calls it "The tally line every runner and the recombination print", but `run-module.sh` prints `Module <id>: N passed, M failed`, which the `^(\d+)\s+passed` anchor cannot match — which is exactly why the `if not tallies and module:` fallback four lines below exists. A reader trusting the comment would conclude that fallback is dead code and delete it.

---

## Suggestions

1. **Wrapper unwrap runs after assignment stripping**, so `env DEVFLOW_BASH=… lib/test/run-parallel.sh` leaves the assignment as the head. Measured `False` — fail-closed, but a spurious refusal with a misleading detail. Re-run the strip after unwrapping.
2. **`except ValueError: continue` produces no breadcrumb.** `_SEGMENT_SPLIT` is applied before `shlex`, so a `;` or `|` inside a quoted operand shreds a legitimate command into unbalanced fragments that are silently dropped; the caller sees only the generic detail. Append `(N segment(s) unparseable)` to the returned detail.
3. **`--result timed_out` / `cancelled` skip the contradiction check** (`if args.result in ("passed","failed")`), so a clean log records `exit_status: 0, passed: 1234` on a flight the caller declared never completed. Not a false green today, but an internally contradictory record.
4. **Partial skip itemization is reconciled in one direction only.** `if not skip_items and skipped_count` pads a fully unitemized population; 2 `SKIP` lines under a `5 skipped` tally keeps 2 and drops 3 with no disagreement check — the one `summary.sh` and `shard-tally.py` both implement.
5. **`combine` and `--require-shards` are presence-checked only** — `combine` is accepted anywhere in the argument list, and `--require-shards` with any value. Anchor `combine` to `rest[0]` and note the residual beside the existing narrowing disclaimer.
6. **Nothing binds the supplied log to *this* flight or *this* tree.** Only `summary["runner_log"] = str(path_str)` is recorded, so any readable file with a CLEAN aggregate satisfies the derivation — an older tree state, a sibling worktree's run root, or a file the agent wrote itself. The hand-assertion has moved rather than been removed. Record mtime and size at minimum, or state the residual the way `classify_whole_suite_command` states its own; the header's "never guesses" currently overstates what is established.
7. **Comment discipline (prevention-only, three physical lines).** These are derivation/provenance over the cap, naming no wrong change they prevent: the ~15-line "Runner-log derivation" banner in `verification-flight.py`; `classify_whole_suite_command`'s second docstring paragraph (run id and design narrative — the first paragraph is the carved-out specification statement); the three-line `#742` block at the call site, which restates that provenance a third time (also a restatement-rule breach); `TestFinishFromRunnerLog`'s docstring middle paragraph; and the five provenance lines above `_cce_tok_for_command`, whose assertion labels already carry `#742`. Also: `_head_invokes`'s docstring example (`my-run.sh` vs `run.sh`) describes a comparison the function is never asked to make — every `target` is a multi-segment path. State the property instead. The comments worth keeping as written are the `_WRAPPER_HEADS` unwrap note, the unreconciled-combine `continue` note, the `skip_items` non-laundering note, the `summary_source_ambiguous` note, and the parity-dimension note in the exit-status parity test.

---

## Gate results

- **Prompt-surface edit routing evidence gate — DISCHARGED.** The diff touches `skills/implement/phases/*.md`, so the gate fires. `closingIssuesReferences` is empty (arm 1 unavailable), and the PR description carries the marker with all four slots stated: `skill-loaded=yes`, `guidance-applied=yes`, `pressure-scenario=no`, `micro-tests=no`. A `no` is a fully discharging disposition.
- **Verification-evidence marker advisory (non-blocking) — FIRES.** Tier classification: local/interactive (no workpad, no linked issue). Neither surface carries a `Verification evidence:` marker. Informational; it does not contribute to the verdict.
- **Phase 0.6 stale-prose lint** — gated off (`prflow_review.stale_prose.enabled` is `false`).
- **Phase 0.3.6 blocker recheck** — not applicable (no prior review on this PR).
- **Reference size** — `phase-4-documentation.md` is 60,088 bytes at the PR head, under the 61,750-byte ceiling. Claim verified; roughly 1.6 KB of headroom remains.

## What is right about this change

The diagnosis is correct and valuable. The RED-first discriminating boundary is stated properly — the 5 positive cases passing before *and* after is the right control. The `#742` table is DAMP with labels that name each refused shape, no wording-only pins were introduced, and the exit-status parity fixture reconciliation is correct with its coupling named in the comment. Leaving the writer unchanged is the right call and is well argued. Existing `suite_summary` fixtures elsewhere in the tree already use `lib/test/run.sh`, so there is no collateral breakage.

The verdict turns on two things: the derivation is verified against a log format no producer emits, and the classifier admits the one invocation the project has already ruled out. Both are fixable without changing the design.

## Verification evidence for this review

Local, read-only. Reviewed HEAD `e8332372ddfe2e9e42c5f4a243c038869d4f0a38`. CI conclusion: **unestablished** — no workflow run exists for that SHA (GitHub Actions outage). Findings C2, C3, I5 and Suggestion 1 were reproduced by executing the PR-head copies of `scripts/check-completion-evidence.py` and `scripts/verification-flight.py`. C1's and I6's producer claims were read from `lib/test/run-parallel.sh`, `run-shard.sh`, `run-module.sh` and `summary.sh`, which this PR does not modify.

The01Geek:
## Review reception — round 1

Pushed `05003dffb1b6cfd80cae90bb4ed3c25478543231`. Every finding was reproduced against PR-head bytes before it was acted on. **All three Criticals and all six Importants reproduced and are fixed; none were refuted.** Suggestions 1–6 are fixed, Suggestion 7 partially.

GitHub Actions is in an outage, so no CI reading exists for this head. Verification below is in-environment focused runs only, and this pass makes no completion claim on it.

### Critical

**C1 — `--from-runner-log` could consume no log the runners retain. Accepted; fixed at the producers.**

Reproduced all three limbs: `run-parallel.sh:597` names `$RUN_ROOT/logs`, a directory, and nothing teed the `aggregate` line anywhere; `run-shard.sh` printed its `retained log:` marker *after* `cat "$LOG_FILE"`, so the marker was not inside the file it named; a serial `run.sh` log carries no self-identifying marker at all.

Fixed by taking your first fix direction — a coupled producer change — rather than only correcting the prose:

- `lib/test/run-parallel.sh` now writes its terminal block to `$RUN_ROOT/coordinator.log` and names it as `run-parallel: retained coordinator log: <abs>`. The `tee` on the recombination carries the tally; a `_coord_line` helper writes the roster, retained-log, elapsed and `aggregate` lines to both the console and that file. It writes per-line rather than tee-ing the whole block on purpose: `aggregate FAILED` and the `MISSING`/`SHARD_RCS` diagnostics stay on the console's **stderr**, while still reaching the file — so a failing run derives as a failure instead of as an unrecognized log. An unwritable run root degrades to a named stderr breadcrumb, not a crash.
- `lib/test/run-shard.sh` appends the marker into `$LOG_FILE` before the `cat`, so console output is unchanged and the marker is now inside the file. (The line is neither a tally nor a `SKIP`, so `shard-tally.py`'s positional parse is unaffected.)
- `lib/test/run.sh` emits `run.sh: serial suite complete (skip-suite-modules=<0|1>, skip-python-pool=<0|1>)`. The selectors are in the marker deliberately — both set is verbatim the `monolith` shard, so a marker without them would have re-opened C2 through a second door: the derivation would have read a reduced-population run as the full serial suite.

End-to-end proof, in `lib/test/modules/parallel-suite-runner.sh` against the real coordinator (not a fixture): the clean run names a coordinator log, that file exists, it carries `aggregate CLEAN` and the recombined tally, a failing run's file carries `aggregate FAILED`, and `derive_summary_from_runner_log` on that exact file returns `lib/test/run-parallel.sh|0`. Mutation-checked: suppressing the coordinator-log write turns 3 of those 6 RED, including the end-to-end one.

**C2 — the classifier accepted the `monolith` shard spelled with its env prefix. Accepted; fixed.**

Reproduced verbatim, including `--preflight` and `--require-shards monolith`. `_segment_tokens` now **returns** the stripped assignments instead of discarding them, and the runner arm refuses a `run.sh` carrying a set `DEVFLOW_SKIP_SUITE_MODULES` / `DEVFLOW_SKIP_PYTHON_POOL`, and refuses `--preflight` / `--list-shards` / `--help` / `--dry-run` — each on the `False` arm with a named detail naming the selector or flag, so the breadcrumb says what to re-run. An explicit `=0` is not a reduction and still passes. Mutation-checked: neutering `_reducing_assignments` turns 3 assertions RED.

`--require-shards <single-shard>` is **not** fixed and is deliberately deferred: the true partition is only knowable by running `--list-shards`, and hardcoding a shard count or name here would drift the moment the partition changes. The residual is now stated in `classify_whole_suite_command`'s own docstring beside the other one, per your Suggestion 6 framing. Revisit if the partition ever gains an offline declaration the classifier can read.

**C3 — a module log's `K skipped` was dropped. Accepted; fixed.**

Reproduced exactly: `Module regen: 40 passed, 0 failed, 1 skipped` yielded `skipped_checks=[]`. `_RE_MODULE` now carries the optional fourth group and the module arm reads it. Two tests cover it — one with itemized `SKIP` lines, one with the count and no itemization (your measured shape). Mutation-checked: restoring `None` for that group turns both RED.

### Important

**I1 — `run.sh` accepted by the gate, unrecognized by the derivation, misattributed. Accepted; fixed.** The `run.sh` marker from C1 closes the seam, and the derivation now identifies runners **outermost first** (aggregate → shard → run.sh → module), so a serial log's incidental `Module <id>:` line no longer wins. Covered by `test_serial_run_sh_log_is_recognized_as_itself` and `test_a_shard_log_is_attributed_to_the_shard_not_its_driver`.

**I2 — "last bare tally wins" overrode the authoritative line. Accepted; fixed.** A module log is now scored by its own `Module` line's groups, never by a driven fixture's trailing tally. Marker precedence inverted as you described. `test_a_module_log_is_not_scored_by_a_nested_sub_suites_tally` pins the `3 passed, 2 failed` + trailing `12 passed, 0 failed` case at `failed=2`.

**I3 — the decisive marker is on stderr and the stream requirement was unstated. Accepted; fixed twice over.** The producer change means the coordinator writes the verdict to the file itself, so the common path no longer depends on the caller's redirection. For a caller-made capture, `runner_log_no_aggregate` is now its own named reason ("coordinator lines are present but no aggregate line is…"), and the `--from-runner-log` help states the merged-capture (`2>&1`) requirement. The `PARALLEL_CLEAN` fixture now matches the real producer including the coordinator-log line.

**I4 — the shipped prose stated an incomplete accepted set. Accepted; rewritten.** It now names the coordinator, `lib/test/run.sh` over its **full population**, and a `--require-shards` recombination; names log paths that exist (`retained coordinator log`, a shard's or module's `retained log`); and carries one consequence clause. `phase-4-documentation.md` is 60,200 bytes, still under the 61,750 ceiling — `lint-reference-size.py` green.

**I5 — skips scanned globally, placeholders counted, `[kind]` discarded. Accepted; all three fixed.** Itemization is now positional from the end of the tally that announced it; `_RE_SKIP_LINE` excludes `summary.sh`'s four parenthesised diagnostics; `_RE_SKIP_TAIL` parses `[blocking-gate|host-capability] — <reason>` into the real `kind` field. Your Suggestion 4 is folded in: the reconciliation is now two-directional — a short itemization pads, and an over-itemization appends an explicit disagreement member rather than silently dropping the surplus.

**I6 — two comments the shipped code refutes. Accepted; both corrected.** Confirmed both: `run-module.sh` also emits `Module %s: heavy units REQUESTED…` and `Log: %s`, and `run-module.sh`'s tally is not `_RE_TALLY`'s shape. Both comments rewritten to state the property that is actually load-bearing (the third group must be read; the module arm reads its own groups rather than falling through), so the fallback can no longer read as dead code.

### Suggestions

1. **Accepted.** Assignments are re-stripped after the wrapper unwrap; `env DEVFLOW_SKIP_PYTHON_POOL=1 lib/test/run.sh` is now refused for the right reason instead of incidentally.
2. **Accepted.** `_segment_tokens` returns the unparseable-segment count (returned, not yielded — a trailing unparseable segment yields nothing and would lose it), appended to the detail as `(N segment(s) unparseable)`.
3. **Accepted, with a different fix.** Rather than extending the contradiction check, `--from-runner-log` now refuses a non-verdict `--result` outright (`runner_log_result_not_a_verdict`): a run that did not complete has no runner log to derive from, so the contradictory record cannot be written at all.
4. **Accepted** — folded into I5 above.
5. **Partially accepted.** `combine` is now anchored to `rest[0]`. The `--require-shards` value check is the deferred residual described under C2, and the docstring states it.
6. **Accepted.** `runner_log_size` and `runner_log_mtime` are recorded, and the header no longer claims it "never guesses" — it now states plainly that the log is not bound to this tree or this flight, as does the flag help.
7. **Partially accepted.** Trimmed the ~15-line derivation banner, `classify_whole_suite_command`'s narrative paragraph, the `#742` call-site block, the test-class docstring middle paragraph and the `_cce_tok_for_command` provenance lines; `_head_invokes`'s docstring now states the path-boundary property instead of the `my-run.sh` example. The five comments you named as worth keeping are unchanged.

### Verification

In-environment focused runs on `05003dffb`:

- `lib/test/run-module.sh parallel-suite-runner` — 179 passed, 0 failed (was 173; +6 new `#742` assertions)
- `lib/test/run-module.sh harness-python-guards` — 45 passed, 0 failed
- `lib/test/run-module.sh implement-contract` — 14 passed, 0 failed
- `lib/test/test_verification_flight.py` — 142 tests, OK (was 133)
- `lib/test/test_python_scripts.py` — 6118 passed, 29 failed; all 29 are the known environment-only `#1655` failures, unchanged from the pre-fix baseline on this host
- `lib/test/lint-reference-size.py` — green
- `shellcheck --severity=warning -e SC1091` over all non-`lib/test` `.sh` plus the four touched runners/modules — clean
- `git ls-files '*.py' | xargs ruff check` — 54 findings, byte-identical to the count at `e8332372d` (all pre-existing)

`lib/test/modules/coverage-map.json` regenerated via `coverage_map_guard.py --fix` for the new `#742` module label. Module assertion floors unchanged — the new assertions only raise the count above an existing floor.

CI is **unestablished** for this head, per the Actions outage; per the fix-loop rule this pass does not cite CI for its own progress, and no whole-suite completion claim is made here.

The01Geek:
## Reception pass on review `5033238275` (reviewed head `e833237`)

That review was submitted at 17:24:05Z against `e833237`; commit `05003df` landed at 17:27Z and is what the findings below are re-checked against. Current head after this pass: `4bbaba0`.

### Important 1 — already discharged by `05003df`

The finding was that no runner-named artifact derives completion evidence the gate accepts. `05003df` applied the second remedy the review itself proposed:

- `lib/test/run-parallel.sh` now writes its terminal block — recombined tally, `elapsed`, and the `aggregate` verdict on **both** the passing and the failing path — into `$RUN_ROOT/coordinator.log`, and names it on exit as `run-parallel: retained coordinator log: <abs>`. The failing path still puts `aggregate FAILED` on stderr for the console while the file gets it too.
- `lib/test/run-shard.sh` appends its `retained log:` marker **into** the file it names, before the `cat`, so the named path carries its own marker.
- `lib/test/run.sh` self-identifies (`run.sh: serial suite complete (skip-suite-modules=…, skip-python-pool=…)`), which is the derivation branch the review said was absent.

Verified end-to-end against current-head bytes: a coordinator-log fixture derives `command = "lib/test/run-parallel.sh"`, `exit_status = 0`, and `classify_whole_suite_command` returns `(True, 'whole-suite runner lib/test/run-parallel.sh')`. Not re-fixed.

### Suggestion 4 — already discharged by `05003df`

`_segment_tokens` now re-strips assignments after unwrapping a wrapper head (`env VAR=1 <runner>`). Confirmed: `env FOO=1 lib/test/run-parallel.sh` classifies as whole-suite, `env DEVFLOW_SKIP_PYTHON_POOL=1 lib/test/run.sh` is refused as reduced-population. An accept-side row for the bare `env` head is added here.

### Important 2 — accepted, fixed

`RunnerLogError` now extends `_CodedError`. Its four exact codes and the `runner_log` prefix are declared, validated at construction, and sealed afterwards. A new test asserts a typo'd code and an unknown prefix raise `ValueError`, and that both `.reason` and the backing `._detail` refuse post-construction assignment. Mutation-checked: reverting the base class to `Exception` turns 1 failure + 6 errors RED.

### Suggestion 1 — accepted, fixed

Confirmed live at `05003df`: a log carrying `700 passed, 4 failed` beside `aggregate CLEAN` derived `exit_status: 0`. Now refused as `runner_log_aggregate_contradicts_tally`. The marker outranks the tally only in the direction that turns a clean-looking tally red; the reverse is truncation or tampering. Mutation-checked.

### Suggestion 2 — accepted, fixed (not covered by the earlier deferral)

The prior cycle's deferral was about *validating the `--require-shards` value against the true partition* — a different residual, which still stands. This finding is about the derivation having no branch at all, and it is fixed: `cmd_combine` emits `shard-tally combine: required partition covered (<n> shard(s)): <ids>` **only** when `--require-shards` was given, membership reconciled, and the tally is clean — exactly the accepted recombination — so that line now derives `lib/test/shard-tally.py combine --require-shards "<ids>"`. The bare `shard-tally combine: <n> shard(s):` roster line is deliberately **not** matched (it prints for every combine, reconciled or not — matching it would be the #1289 false green), and the branch is tested after `aggregate`, so a coordinator log that tees both is still owned by its own verdict. Both directions mutation-checked.

### Suggestion 3 — accepted, fixed

Added: accept-side wrapper-head rows (`bash lib/test/run.sh`, `python3 lib/test/shard-tally.py combine … --require-shards …`), the `--require-shards=` equals form, the bare `env` head, an unparseable-segment reject row, a combine-subcommand-not-first reject row, a multi-tally coordinator fixture pinning last-tally-wins, and the shard branch's tally-decides-result path.

### Suggestion 5 — 1 of 3 already discharged; 2 of 3 refuted

- `TestFinishFromRunnerLog` docstring: the provenance paragraph (the "two of five implement runs analysed… run 32957163134" narrative) **was** present at `e833237` and `05003df` already removed it. Discharged.
- `verification-flight.py:380-394` at the reviewed commit `e833237` is the `chmod` breadcrumb, `_flight_path`, and `_atomic_replace` — no provenance narrative, and untouched by this PR.
- `test_python_scripts.py:34-39` at `e833237` is the tail of the pre-existing module docstring about the deferrals-matcher payload, also untouched by this PR (`git diff 73fe58c..HEAD -- lib/test/test_python_scripts.py` shows a single added hunk far below it).

The prevention-only comment standard binds comments a change adds or modifies, so neither of those two is in scope for this PR. No edit made; recorded here as the deferral trace.

### Verification

Per this repo's fix-loop discipline, no CI conclusion is cited for this pass's own progress.

- `lib/test/run-module.sh harness-python-guards` — 45 passed, 0 failed (the module that owns `scripts/verification-flight.py`).
- `python3 lib/test/test_verification_flight.py` — 150 passed.
- `lib/test/test_python_scripts.py` — 6126 passed, 29 failed, matching the known `#1655` environment-only local baseline that is green in CI.
- `ruff check` clean on both touched files. Local ruff is 0.15.21, skewed from CI's pin; the 44 findings it reports on `test_python_scripts.py` reproduce identically on the merge-base copy, so they are pre-existing and uncharacterised locally, not from this change.
- No `.sh` files touched, so no shellcheck delta.

